PR #7, opened 2015-07-26
Cosmetic: remove execute bits from non executabe files

athoik:
No need to have execute bit on Makefile.am, *.xml, *.jpg


littlesat:
Can't be merged... A log of filemode changes

littlesat:
Oeps... It seems intended

athoik:
Only permissions changed, there are no changes in files. Github doesn't accept such a push?

PR #8, opened 2015-07-27
Some OE-A + PE changes - Part 1

persianpros:
Thanks to OE-A members for OE-A changes


athoik:
I think it would be better to create a commit per feature, that would make easier the merge workflow

persianpros:
As you said "custom build" then no need for it in the main releases because developers could enable it for their own use ;)

Ping more reliable addresses will give you fast ping/network test which I test it myself, even some google domains like google.co.uk/.de/.nl/... are not accessible in all countries so why not have something globally

And about the "Close files after being used" commit we have to be sure not rely on most cases also it's just the beginning and more like this commit will come

Well I didn't know we have more than 1GB flash available, if there is then it should be 2048 instead of 1024 and I agree with you that we have to use a better technique but for now it will do the job

I thought sending bunch of commits would be better not one by one, maybe one of PLi members could point me to the right direction

Anyway it's the first part, PLi could merge what they want, the thing is we're using these commits with our patches and I'm sending merge requests for contribution only :)

milosoftware:
Quite convenient Github's pull/merge stuff. Seems I can't cherry-pick though, could you split them up?

persianpros:
Which commits do you want ? I will send them again in another merge request

PR #9, opened 2015-07-27
Servicemp3 bug in delayed subtitles.

christophecvr:
If You tried to add a delay (positif or negatif) to srt subtitles,
	it did not worked out and the result was no subtitles.
	It was a small error, just a forgotten division by 90.
	With this patch delayed srt subtitles are working.


christophecvr:
Thanks

PR #10, opened 2015-07-27
Fix vtuner frontend detection

persianpros:
Thanks Skaman


littlesat:
Why is this required?

PR #11, opened 2015-07-27
larger buffer for 1024MB systems like solo2/duo2

persianpros:
Thanks to OE-A


littlesat:
What is the advantage?

PR #12, opened 2015-07-27
Some fixes

persianpros:
Thanks to OE-A


milosoftware:
Please split up unrelated fixes. The DVB-C fix for example looks okay, but I don't like the "close file" patches.

PR #14, opened 2015-07-28
latest JESS patch from adenin

persianpros:
Also with some OE-A changes

I think you really need to review this before merging it because OE-A is messy always specially in e2 part as they don't merge with latest PLi changes

I hope this patch could help you save some time at least

### DIFF
--- a/data/unicable.xml
+++ b/data/unicable.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="iso-8859-1"?>
-<unicable version="20100522">
+<unicable version="20150726">
 <!--
 structur of tree
 unicable (version)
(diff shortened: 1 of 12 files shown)


RobvanderDoes:
< I think you really need to review this before merging it because OE-A is messy always specially in e2 part as they don't merge with latest PLi changes >
As the name already reveals: the OE-A has nothing to do with (whatever which version of) E1/2.

milosoftware:
What is a "JESS"?

persianpros:
http://forums.openpli.org/topic/37669-e2-patch-for-jess-jultec-enhanced-stacking-system-en50607-support/

littlesat:
I did overlook this patch... But this patch does not really look 'nice'.... :(

PR #20, opened 2015-08-10
By some dvd's resume does work.

christophecvr:
Restore the abilyty to resume by dvd.By some dvd's it does work.
 By other dvd's not.

 When there is no resume postion or it is 0 ,
 No need to save a resume.cuts file or keep one with 0.

	modified:   lib/python/Screens/InfoBarGenerics.py
	modified:   lib/service/servicedvd.cpp

### DIFF
--- a/lib/python/Screens/InfoBarGenerics.py
+++ b/lib/python/Screens/InfoBarGenerics.py
@@ -2856,8 +2856,7 @@ class InfoBarCueSheetSupport:
 		print "new service started! trying to download cuts!"
 		self.downloadCuesheet()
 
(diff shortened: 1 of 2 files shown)


mirakels:
First of all these are 2 unrelated changes (although they talk about hte same subject), so they should be split (the first one actually is a revoert of a previous commit iirc)

The second change is done a the wrong place. it should be something like

if (m_cue_pts == 0  || !resume_info.title) {
    #  menu or pointless cue. remove old cue file if exists.
   ...
}

# save new cuefile)

christophecvr:
I suggest You read comment I left by pm.

PR #23, opened 2015-08-12
Revert Do not show resume notification when playing a DVD image

christophecvr:
modified:   lib/python/Screens/InfoBarGenerics.py


littlesat:
Why is the -15 sec requiered.... when it is related to VU only I would consider to revert this part later on!

PR #34, opened 2015-09-06
Fix for USALS-selecting-satellites bug(part 2)

Dima73:
same fix for NimManager


eriksl:
What does it do and how does it fix?

Dima73:
Correctly checks the list.
Python 2.7.3 (default, Feb 15 2015, 11:52:13)
[GCC 4.9.1] on linux2
>>> user_list = "549"
>>> number_sat = 49
>>> str(number_sat) in user_list
True

>>> number_sat = 549
>>> str(number_sat) in user_list
True

>>> number_sat = 9
>>> str(number_sat) in user_list
True

>>> number_sat = 360
>>> str(number_sat) in user_list
False

PR #37, opened 2015-09-11
DVDplayer plugin not always went away when a dvd is removed.

christophecvr:
The menu item in mainmenu DVD-player. Not always went away,  up on removal
 off dvd from reader. Now it does.  Changed the hard targetting to /media/sr
 to the real detected mountpoint.

	modified:   lib/python/Plugins/Extensions/DVDPlayer/plugin.py

### DIFF
--- a/lib/python/Plugins/Extensions/DVDPlayer/plugin.py
+++ b/lib/python/Plugins/Extensions/DVDPlayer/plugin.py
@@ -28,10 +28,10 @@ def filescan_open(list, session, **kwargs):
 	from Screens import DVD
 	if len(list) == 1 and list[0].mimetype == "video/x-dvd":
 		cd = harddiskmanager.getCD()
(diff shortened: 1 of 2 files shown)


Taapat:
Maybe I am missing something but:
1. Why in filescan_open and menu assign "cd = harddiskmanager.getCD()",
and after all the time call harddiskmanager.getCD() instead of using the cd?

2. If the detected_DVD value may be None, False or True, why in #90 change to "if detected_DVD == True:" if "if detected_DVD:" is a lot faster?

christophecvr:
1.It is required to scan for VIDEO_TS or video_ts (case iso9660) We are not using hardtargetting anymore like it was before to sr. The hotplug will now mount the dvd to volume lable if avbl which is standard practice. In linux
2. Gues that if detected_DVD may work as well but with the expected possible value it does work very stable.

PR #47, opened 2015-09-21
Avoid GS when wlan driver has non-standard structure

j00zek:
With this change unknown wlan card is listed as unkown instead of GS.


eriksl:
Applied.

PR #56, opened 2015-10-02
add new option Wakeup timer

Dima73:
-choose for each day required time for wakeup box
-changed the logic  startup to standby for auto boot:
1)timers record/zap goto standby
2)plugins not standby, e.g. epgrefresh/xmlimport
3) wakeup timer and 'Startup to Standby' as 'No, except Wakeup timer'
goto standby


Dima73:
No, this option is quite another.
For example the receiver is used as a server Stream.
Without zap timers.
By the way, step 2 simple power timer (off) at a specified time.

PR #67, opened 2015-10-14
PositionerSetup: add missing orb_pos in createConfig

Taapat:


littlesat:
Please evaluate if it could better be orb_pos = None as orb_pos = 0 is in fact a satellite location...
For now merged to avoid at least the GSOD

PR #73, opened 2015-10-24
fixed incorrect use of pointer to temporary object.

stevenhoving:
cppcheck reported as:
Summary: Dangerous usage of c_str(). The value returned by c_str() is invalid after this call.
Message: Dangerous usage of c_str(). The c_str() return value is only valid until its string is deleted.


eriksl:
I don't think this will give problem in practice, but the coding is better anyway, so I applied it. Thx!

PR #75, opened 2015-10-24
refactored config manager function names to be shorted without loosin…

stevenhoving:
…g there explicit naming. While I was at it I teached the getString (the old getConfigValue)
function how to handle a default value. This simplified a use case in the
mp3 service.


eriksl:
Rejected. Contributers are used to these method names even though your names are probably better. I will accept a separate patch for the null avoidance.

PR #76, opened 2015-10-24
remove unneeded struct type specifier from sizeof.

stevenhoving:
This seems to be a leftover from back in the days when everything was
written in C.


eriksl:
I usually review all patches once a week. For the other patches you were "lucky" MiLo beat me to it. Exactly why he applied some of them and not all I don't know. I guess he picked the functional changes.

This patch applied. Not just because the "struct" keyword is superfluous here, but because it fixes a real problem where 8192 as size was copied-and-pasted literally, which is an error.

PR #78, opened 2015-10-25
timer conflict detection advanced checking

Dima73:
-show option"Enable timer conflict detection" only
SystemInfo["Fallback_remote_tuner"]  is True.Since this option is only for advanced users.


eriksl:
Littlesat beat me to it. There are other circumstances where conflict detection may need to be disabled, so I won't accept this approach.

Dima73:
OK.
No problems.

Question ...
I make a patch for each timer to disable conflicts check?

eriksl:
If you want to.

Dima73:
Standard timers and autotimers

eriksl:
I you think it is necessary. The problem I had is solved with the global setting.

PR #79, opened 2015-10-28
servicemp3: http headers via service reference path string

mx3L:


eriksl:
And why exactly is this better?

mx3L:
Currently we can provide extra http headers via "config.mediaplayer.extraHeaders" config entry.
When it's used on bouquet items it's quite restrictive, since we could only have one extra header set for every item in bouquet.
With this patch we can set extra headers for every bouquet individually

http://forums.openpli.org/topic/29501-gstreamer-10/?view=findpost&p=511096

eriksl:
Shouldn't every service have it's own set of extraheaders? Something like 1:0:1:0:0:0:0:0:0:0:http://url_of_stream?extra_headers["example_header"]="example_header_content"&extra_headers["example_header_2"]="example_header_content"
I agree this would be a bit tricky to parse (and remove before passing it to gstreamer), but I guess it would be the most flexible approach.

mx3L:
I'm not sure I follow.
By "every service" you mean also servicedvb and not only gstreamer service?

If it's only about gstreamer service, then with this patch we can pass as many headers as neccesary for every service, like this:

4097:0:1:0:0:0:0:0:0:0:http://url_of_stream#example_header=example_header_content&example_header_2=example_header_content

eriksl:
That is very similar to what I had in mind ;)

And then indeed you have extra headers per service, not per bouquet.

It's just that I hope the hash isn't used in the providers URL's themselves.

mx3L:
As @athoik pointed out in forum '#' char is delimiting fragment part of url, which is normally not passed to server and is used only locally by browsers, so it should safe:

https://blog.httpwatch.com/2011/03/01/6-things-you-should-know-about-fragment-urls/

athoik:
We can add the same behaviour in lib/base/httpstream.cpp and then also TS streams will have extra headers :+1:

eriksl:
Yes, clear, nice. Applied.

PR #80, opened 2015-10-30
servicemp3: Always initialize pointer to NULL before pass to gst_tag_list_get_string

athoik:
More info: http://forums.openpli.org/topic/29501-gstreamer-10/page-93

### DIFF
--- a/lib/service/servicemp3.cpp
+++ b/lib/service/servicemp3.cpp
@@ -1295,7 +1295,7 @@ std::string eServiceMP3::getInfoString(int w)
 	}
 	if ( !tag )
 		return "";
-	gchar *value;
+	gchar *value = NULL;
 	if (m_stream_tags && gst_tag_list_get_string(m_stream_tags, tag, &value))
 	{
 		std::string res = value;


littlesat:
So base on the comments from mx3L I suggest we should not merge it...

eriksl:
Does this solve a potential segfault? In that case it may be solved in another way?

athoik:
We are using exacly the same pattern everywhere in servicemp3.cpp.
eg.

```
    g_codec = NULL;
....
    if (gst_tag_list_get_string(tags, GST_TAG_AUDIO_CODEC, &g_codec))
    {
        audio.codec = std::string(g_codec);
        g_free(g_codec);
    }
...
```

The segfault was happening because **the following** (wrong) pattern was followed:

```
   char * value; // a pointer uninitiliazed (that might point in a location != NULL)
   gst_tag_list_get_string(tags, tag, &value); // we are not checking if get string failed
   if(value){ // value != NULL so true, value was not touched by get string
       g_free(value) // here we are trying to free an uninitialized value, boom!
   }
```

Also looking on gst_tag_list_get_string code, it will return false if the value is not NULL. 

```
gboolean                                                                \
gst_tag_list_get_ ## name (const GstTagList *list, const gchar *tag,    \
                           type *value)                                 \
{                                                                       \
  GValue v = { 0, };                                                    \
                                                                        \
  g_return_val_if_fail (GST_IS_TAG_LIST (list), FALSE);                 \
  g_return_val_if_fail (tag != NULL, FALSE);                            \
  g_return_val_if_fail (value != NULL, FALSE);                          \
                                                                        \
  if (!gst_tag_list_copy_value (&v, list, tag))                         \
      return FALSE;                                                     \
  *value = COPY_FUNC (g_value_get_ ## name (&v));                       \
  g_value_unset (&v);                                                   \
  return ret;                                                           \
}                                                                       \

```

So in order `gst_tag_list_get_string` to work as expected we need to pass `NULL` `value` and we are already doing this in servicemp3. 

That commit fixes the only missing / forgoter uninitialized value that used in `gst_tag_list_get_string`.

eriksl:
Clear.

Applied.

mx3L:
>> The segfault was happening because the following (wrong) pattern was followed:...

agreed, but in discussed case we are not doing anything with *value*, when `gst_tag_list_get_string`
returns FALSE.

>Also looking on gst_tag_list_get_string code, it will return false if the value is not NULL.

Actually it's opposite, it will return false in case value is NULL. We are checking address of `value` pointer and not address where `value` pointer points to.
```
if (m_stream_tags && gst_tag_list_get_string(m_stream_tags, tag, &value))
```
We are passing address of pointer to `gst_tag_list_get_string` and it cannot be NULL, if it's NULL, then
```
if (m_stream_tags && gst_tag_list_get_string(m_stream_tags, tag, NULL))
```
will `gst_tag_list_get_string` return logically immediately FALSE.

>So in order gst_tag_list_get_string to work as expected we need to pass NULL value and we are already doing this in servicemp3. 

No need to pass NULL as explained

athoik:
>Actually it's opposite, it will return false in case value is NULL. We are checking address of value pointer and not address where value pointer points to.

```c
#include <stdio.h>
#include <stdlib.h>

void test();

void testa(char **value)
{
    if(value != 0) printf("A pointer not initialized at %p\n", value);
    else printf("A pointer not initialized that points to NULL!\n");
}

int main()
{
   test();
}

void test(){
   char * a_pointer_not_initialized;
   testa(&a_pointer_not_initialized);
}
```

Ok you are right, initializing to NULL doesn't affect `gst_tag_list_get_string`. 

So it just follows existing pattern in servicemp3.

mx3L:
>So it just follows existing pattern in servicemp3.

Agree, it's good approach

PR #81, opened 2015-10-31
eHttpStream: set http headers via service reference path string

athoik:
Follow commit 17d8930111831bb80775b8184fa23f5634537fe0 and allow the same behaviour for TS streams (SERVICE 1:0:1)


eriksl:
Applied.

PR #83, opened 2015-11-01
[AudioSelection] add ChoiceBox if more audio menu plugins (key blue)

Dima73:


littlesat:
Nice!

eriksl:
And what does it do actually?

Dima73:
Previously, the blue button opens the first plugin from the list of audio plugins(PluginDescriptor.WHERE_AUDIOMENU).
Now opens a list of audio plugins.

PR #85, opened 2015-11-05
Add option enable/disable conflict detection for each timer

Dima73:


littlesat:
Please do not make changes to the current general timer conflict detection and remove all checks for the fallback tuner.... T

eriksl:
I think littlesat means: don't use the fallback tuner setting (on/off) for assumptions with timers. I agree, we talked about that. Lets discuss on the forum.

Dima73:
<Please do not make changes to the current general timer conflict detection>
if only small rename text "Enable timer conflict detection"-->"Enable global timer conflict detection"

<remove all checks for the fallback tuner>
If it is not turned on "Enable fallback remote receiver", then I do not understand the meaning of the others changed.
The basis of all - use of a fallback tuner

littlesat:
I want to remove the general setting.... And go for your individual one only and then without a fallback tuner is availalble check....

Dima73:
ок

PR #87, opened 2015-11-07
Components Language: add os.environ["LANGUAGE"]

Dima73:
-it solves a strange problem with the installation of the native language
for some plugins, e.g. PartnerBox


littlesat:
And what about the utf8????

Dima73:
I'm sorry, I did not understand.

Dima73:
https://github.com/OpenPLi/enigma2-plugins/commit/f9ba4297c3fb3a0d8c6859068159d602ab70a228

eriksl:
Dimitrij, as we are using UTF-8 on enigma2, there is no need to select encodings. Just make sure all other encodings are converted to UTF-8, that is the real solution.

PR #100, opened 2015-11-17
servicemp3: use seeking un-pause only when necessary

mx3L:
related patch:
https://github.com/OpenPLi/gst-plugin-dvbmediasink/pull/5


eriksl:
Applied.

PR #113, opened 2015-12-04
Change some title translations, to avoid add a new very similar and repeating

Taapat:
Whay add new if alredy exist similar translations?


eriksl:
Applied.

PR #122, opened 2015-12-08
ChoiceBox: add set window title

Dima73:
e.g. Extensions menu and etc


Dima73:
Maybe specially made without text, example windowTitle = " "

eriksl:
Applied.

PR #123, opened 2015-12-09
adding invalid onid/tsid on 26.0E &7.0E

kam10:


eriksl:
The description is in the patch.

Applied.

PR #124, opened 2015-12-10
Fix rgb profile and optimize added timer_autotimer.png

Taapat:
Use pngquant to fix rgb profile and optipng to optimize png


eriksl:
Applied.

PR #131, opened 2015-12-13
SatelliteEquipmentControl: update plugin

Dima73:
-fix BSOD set tiltle
-small update size skin
-add for key menu "Set default settings?" yes/no


eriksl:
Applied.

PR #133, opened 2015-12-14
DVB-S sat config: set default nothing connected

Dima73:
-this solves the problem empty list of satellites

gma2/python/Plugins/SystemPlugins/Satfinder/plugin.py",
line 101, in createSetup
if len(nimmanager.getTransponders(int(self.tuning_sat.value))) < 1: #
Only offer 'predefined transponder' if some transponders exist
ValueError: invalid literal for int() with base 10: ''


eriksl:
Applied.

PR #135, opened 2015-12-16
Subservices: update code

Dima73:
-redesigned zap to service 
-add windowTitle
-applied code from opendreambox.org


eriksl:
I'd like to have some more explanation about this.

Dima73:
-add arg adjust=False if run self.session.nav.playService
-update keys if open ChoiceBox(SubservicesQuickzap)

PR #138, opened 2015-12-19
PictureInPicture: change text No free tuner! --> No free tuner for Pi…

Dima73:
…P service!

e.g. run PiP from the channel list/it more clear


eriksl:
I don't think it's appropriate, if no tuner is free, that can apply to PiP, but also to recordings.

Dima73:
Why is not appropriate?
It explains exactly why we can not play the service PiP.

eriksl:
It can also show when you try to watch a channel but a recording is active. Or streaming.

Dima73:
This notice, that we can not turn service PiP.
This window PictureInPicture.not live TV.

littlesat:
Closed.... As it does not really got added value and all translaters need to translate it...

PR #140, opened 2015-12-20
RecordTimer: try to disable pip when start record failed (not free tu…

Dima73:
…ner)


Dima73:
This need if enabled Pipzap and set servicelist dopipzap is False
e.g.https://github.com/OpenPLi/enigma2/blob/dc5246882de8c890b72b101157e0ab81014156e6/lib/python/Screens/Standby.py#L83

PR #146, opened 2015-12-26
CommonInterfaceAssignment: add option 'Add all services provider'

Dima73:
-small update set config in setDescrambleRules

### DIFF
--- a/lib/python/Plugins/SystemPlugins/CommonInterfaceAssignment/plugin.py
+++ b/lib/python/Plugins/SystemPlugins/CommonInterfaceAssignment/plugin.py
@@ -183,7 +183,7 @@ class CIconfigMenu(Screen):
 
 	def cancel(self):
 		self.saveXML()
-		activate_all(self)
+		activate_all(self, editcallback=True)
 		self.close()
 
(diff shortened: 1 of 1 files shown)


eriksl:
What does it do?

I am not going to allow replaces for unicode control characters on service names. If some service has illegal characters in it, we should replace them before adding them to the lamedb instead.

Split functional changes and cosmetics (white spaces etc.)

Dima73:
<Split functional changes and cosmetics (white spaces etc.)>
Where, I do not see them?

< am not going to allow replaces for unicode control characters on service names. If some service has illegal characters in it, we should replace them before adding them to the lamedb instead.>
I took the example of https://github.com/OpenPLi/enigma2/search?utf8=%E2%9C%93&q=replace%28%27\xc2\x86%27%2C+%27%27%29.replace%28%27\xc2\x87%27%2C+%27%27%29

eriksl:
service_name = service_name.replace('\xc2\x86', '').replace('\xc2\x87', '')

c286 and c287 are unicode control characters and imho they should not appear in service names.

PR #151, opened 2015-12-31
Converter/ServiceName.py: add type service number

Dima73:
-e.g.  when open Channel Selection and show selected service number

### DIFF
--- a/lib/python/Components/Converter/ServiceName.py
+++ b/lib/python/Components/Converter/ServiceName.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 from Components.Converter.Converter import Converter
-from enigma import iServiceInformation, iPlayableService, iPlayableServicePtr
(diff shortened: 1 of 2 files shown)


lib/python/Components/Converter/ServiceName.py
```diff
@@ -47,6 +50,15 @@ def getText(self):
 			if nref:
 				ref = nref
 			return ref.toString()
+		elif self.type == self.NUMBER:
+			if not ref:
+				ref = eServiceReference(info.getInfoString(iServiceInformation.sServiceref))
+			num = ref and ref.getChannelNum() or None
```
athoik: How about this?

```num = ref and str(ref.getChannelNum()) or '---'```

Dima73:
if it's not favorite list
  the answer ref.getChannelNum()  would be 0
num = ref and str(ref.getChannelNum())
print '0'

athoik:
So it work's just like the original code, 0 is not None.

PR #153, opened 2016-01-03
Improvise Parental control

Dima73:
step 1:
-remove flag hidden(dxDontshow)  for all services only use "Protect
services" and "Hide parentel locked services"
-mytest.py --> first InitParentalControl / two eDVBDB.getInstance()

### DIFF
--- a/lib/dvb/db.cpp
+++ b/lib/dvb/db.cpp
@@ -356,6 +356,8 @@ void eDVBDB::reloadServicelist()
 
 void eDVBDB::parseServiceData(ePtr<eDVBService> s, std::string str)
 {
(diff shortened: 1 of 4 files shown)


Dima73:
Why do not you understand?
I'm just trying to improve the performance of  function Parental Control.
save config...
1)if disable Parental Control -->reset all services blacklist(unhide)
2)	'if config.ParentalControl.hideBlacklist.value and (not config.ParentalControl.servicepinactive.value or config.ParentalControl.storeservicepin.value == "never"):
		config.ParentalControl.hideBlacklist.value = False'
What's wrong with that? It's absolutely right.
If the function is not active, it is necessary to put it in a position of not.

PR #154, opened 2016-01-08
fixed position for LcdLiveTV test 

ims21:


littlesat:
Whe do we get more here... we only need the fixed position for lcdLiveTV

PR #164, opened 2016-01-25
MovieSelection/MovieContextMenu: add new option "Auto play blu-ray file"

Dima73:
-select folder BDMV --> press menu and set this opton
play *.m2ts ( large size)
-add spaces for better readability

### DIFF
--- a/lib/python/Screens/MovieSelection.py
+++ b/lib/python/Screens/MovieSelection.py
@@ -302,10 +302,13 @@ class MovieBrowserConfiguration(ConfigListScreen,Screen):
 	def changedEntry(self):
 		for x in self.onChangedEntry:
 			x()
+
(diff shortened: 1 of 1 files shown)


milosoftware:
Please always create separate commits for "cosmetic-only" changes.

milosoftware:
Try integrating this with the standard can_/do_ functions that the movielist provides. This allows the function to be bound to a function key, and reduces the amount of code. Or integrate it like the DVD playback and make this the default action on BR images.

Dima73:
Sorry, but what you do not like that option?

milosoftware:
Make a separate commit that contains only the whitespace changes.

The functionality is fine, it would just be better to fully integrate it, so that pressing "OK" on a BR folder will play the movie, just like what happens for DVD folders.

PR #165, opened 2016-01-26
add new flag dxIsParentalProtected

Dima73:
-  flag dxIsParentalProtected  for hidden blacklist services parental
control


mirakels:
After this pull was included I do not have ANY OSD any more. Everything is black.
OSD seems to work as the PIG is still shown when for exampel pressing MENU or EPG

PR #167, opened 2016-01-28
MovieSelection: optimize isBlurayFolderAndFile

Taapat:
Removes unnecessary checks, which have already been made before.
Not create a list of m2ts files to later use sort, len, e.t.c... Simply set the largest file to self.playfile.
Add the possibility of Auto play blu-ray file simply by selecting the folder. Why always need to go inside /BDMV/STREAM?
Not used twice try. If there is an error or is really important exactly what?


Taapat:
Sorry for the three commits, if necessary I can create a new pull request with one commit.

PR #171, opened 2016-01-29
Renderer/CiModuleControl: not show state -1(error)

Dima73:
-strange response to formuler1, if there is no CI module answer
-1(error)


eriksl:
Applied.

PR #178, opened 2016-02-04
Timer Edit: add new option show notify conflist

Dima73:
- run  'Timer Edit' and press  menu if timer list in timer witch
option conflict_detection disable show "Timer sanity error" screen only
simulate


littlesat:
Why making this extra complicated????

PR #191, opened 2016-02-09
ChannelSelection: remove parental control stuff if open bouquet

Dima73:
it is quite unnecessary, as all services of the bouquet is the parent
control.
Moreover, there are other ways to get in a bunch without checking
/change bouquet(Channe lSelection) <>/ChannelSelectionEPG


littlesat:
We can block bouquets that was by design....

PR #198, opened 2016-02-13
Parental Control: add notify for default value 0000

Dima73:
If pin-cod 0000 parental control not enabled.Users did not know about
it.


Taapat:
This is really necessary. Without such information users choose simple code 0000 and does not understand why it does not work.
Only I think that the text could be some kind of follows:
"You PIN code is 0000. This is the default PIN code and it disable parental control!"

Dima73:
<Or maybe a hint? >
yes

PR #204, opened 2016-02-24
DVD.py: fixed key OK (show/hide infobar)

Dima73:
1) if press key OK run 2 tvice toggle show
2)small correct code


littlesat:
I understand but there is much much more happening inthis commit suggestion... At least summarize it in these comments... :)

Dima73:
-for print add name sreen "[DVD]"
-add a sanity check in function calls (self.in_menu / self.dvdScreen)
-add trigger 'iPlayableService.evEnd: self.__serviceStopped'
-change the code position, for better readability of the code (the first _init)

PR #206, opened 2016-02-29
Possibility to set custom pid for parsing epg data

Cofybreak:
This is very useful if a transponder epg be filed in 2 or more different pids or use non stardart one.
For example, the platform Dolce sent epg data on Romanian language on PID 0x12 and Bulgarian and Serbian on PID 0x12b. If create reader for PID 0x12b, some events are Romanian language, others are Bulgarian because of PID 0x12 no information about the language of epg data.
Usage:
Create a file "/etc/enigma2/epgpids.cusom" and fill in the format

Orbital position | TSID | ONID | Custom pid to use

The values ​​can be decimal or hexadecimal (0x ...), and combined
For orbital position using integers where 1 means 0,1 ° (19,2°E -> 192)
For orbital position can be used as negative numbers (0,8°W -> -8 or 3592)

examples:
390 17 28 299 # 39°E 12688V, epgpid to use: 299
0x186 0x11 0x1c 0x12b # Same, but in hexadecimal format
390 17 28 0x12b # Or combined

### DIFF
--- a/lib/dvb/epgcache.cpp
+++ b/lib/dvb/epgcache.cpp
@@ -405,10 +405,10 @@ eEPGCache::eEPGCache()
 		while (!pid_file.eof())
 		{
 			getline(pid_file, line);
-			if (line[0] == '#' ||
-				line.empty() ||
-				sscanf(line.c_str(), "%i %i %i %i", &op, &tsid, &onid, &eitpid) != 4) continue;
-			if (op < 0) op += 3600;
+			if (line[0] == '#' || sscanf(line.c_str(), "%i %i %i %i", &op, &tsid, &onid, &eitpid) != 4)
+				continue;
+			if (op < 0)
+				op += 3600;
 			if (eitpid != 0)
 			{
 				sprintf (optsidonid, "%x%04x%04x", op, tsid, onid);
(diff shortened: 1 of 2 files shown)


Dima73:
Need make vector, an example here
https://github.com/OpenPLi/enigma2/commit/f0b31031a62049d9a645fb2498a83e79265799e1

Cofybreak:
Will try with mapping

littlesat:
Does it still has a saperate config file?

Cofybreak:
Yes, but now this file is reading only when enigma2 start. Then we create a table, from which take a right epg pid. Same as "encodig.conf".
I don`t think that editing  by hand "lamedb" is a good idea.
And adding an extra option to the **channel** is illogical.
To the transpordes - maybe, but not to channels.

littlesat:
We prefer the config is moved somehow to the transponder part in lanedb

Dima73:
Everything is done correctly.
Use the configuration file is very convenient.
Why do we need "lamedb" ?
From which you can delete all the data,when use "Clear before scan".

littlesat:
Because i think we should do it differently... But please discuss it on our forum... We need to find the best way tomput e config file somewhere to avoid it is parsed many times.

Dima73:
The settings are read once at the start of the Enigma.

littlesat:
See forum....

littlesat:
Pushed for now... But I still think we need a config file that includes the languages and all PIDs and that the language selection should select the correct language from this config file

PR #210, opened 2016-03-02
[MediaPlayer] fix possible BS with some plugins

Dima73:
idea
https://github.com/openatv/enigma2/commit/2aebcdeda138c06e4f53bd3ce9e0c5fd7eddca12


littlesat:
Not acceptable... The mediastreamer plugin should not interfere with mediaplayer config.... I prefer the mediastreamer plugin to be fixed...

PR #224, opened 2016-03-13
PiP: solve two problem for swap services

Dima73:
1)InfoBarGenerics: use getCurrentlyPlayingServiceOrGroup() for correct
number service for alternative services
2)ChannelSelection: when swap services in history list there are two
identical entries(!!!),delete old entry and add new

### DIFF
--- a/lib/python/Screens/InfoBarGenerics.py
+++ b/lib/python/Screens/InfoBarGenerics.py
@@ -2224,7 +2224,7 @@ class InfoBarPiP:
 					self.session.pip.servicePath = currentServicePath
 					self.session.pip.servicePath[1] = currentBouquet
 				if slist and slist.dopipzap:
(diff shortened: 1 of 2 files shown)


lib/python/Screens/InfoBarGenerics.py
```diff
@@ -2224,8 +2224,7 @@ def swapPiP(self):
 					self.session.pip.servicePath = currentServicePath
 					self.session.pip.servicePath[1] = currentBouquet
 				if slist and slist.dopipzap:
-					# This unfortunately won't work with subservices
-					slist.setCurrentSelection(self.session.pip.getCurrentService())
+						slist.setCurrentSelection(self.session.pip.getCurrentService())
```
Taapat: It looks like that there is a error. Two tabs instead of one.

Dima73:
thanks

PR #227, opened 2016-03-17
Enable SNR calculation for osmini

Cofybreak:


athoik:
Why would you like to do that?

eriksl:
This codes is unnecessary and will void the advantage of the dvb4linuxV5 interface of the driver.

PR #230, opened 2016-03-18
Update pt_BR.po

frankviana:


frankviana:
This is only for test... but translate ok. Later, I'll try to translate all the missing words.

PR #233, opened 2016-03-22
Fix and Edit on Indonesian

budiarno:
Just fixed and edited lines, also changed the pic for the flag.


Fauzinani:
Thanks om pedode

PR #239, opened 2016-03-29
Harddisk: close file after open

Dima73:


littlesat:
Isn't it better op open the file and use the results directly.... Then I think close is not required at all.... But it still seems you found some open left overs here.... Thanks!

Dima73:
This is a standard action, to close the file.
Which should be used everywhere.
The patch does not break anything.

Dima73:
When you’re done with a file, call f.close() to close it and free up any system resources taken up by the open file. After calling f.close(), attempts to use the file object will automatically fail.


see https://docs.python.org/2/tutorial/inputoutput.html
7.2

mx3L:
````
import os

def open_files():
    f = open("./test1", "w")
    f1 = open("./test2", "w")
    os.system("ls -l /proc/self/fd/")


if __name__ == "__main__":
    open_files()
    os.system("ls -l /proc/self/fd/")
```

> total 0
> lrwx------ 1 marko marko 64 Mar 30 09:10 0 -> /dev/pts/21
> lrwx------ 1 marko marko 64 Mar 30 09:10 1 -> /dev/pts/21
> lrwx------ 1 marko marko 64 Mar 30 09:10 2 -> /dev/pts/21
> l-wx------ 1 marko marko 64 Mar 30 09:10 3 -> /home/marko/test1
> l-wx------ 1 marko marko 64 Mar 30 09:10 4 -> /home/marko/test2
> lr-x------ 1 marko marko 64 Mar 30 09:10 5 -> /proc/28088/fd
> lrwx------ 1 marko marko 64 Mar 30 09:10 6 -> socket:[3271768]
> lrwx------ 1 marko marko 64 Mar 30 09:10 9 -> socket:[3271503]
> total 0
> lrwx------ 1 marko marko 64 Mar 30 09:10 0 -> /dev/pts/21
> lrwx------ 1 marko marko 64 Mar 30 09:10 1 -> /dev/pts/21
> lrwx------ 1 marko marko 64 Mar 30 09:10 2 -> /dev/pts/21
> lr-x------ 1 marko marko 64 Mar 30 09:10 3 -> /proc/28090/fd
> lrwx------ 1 marko marko 64 Mar 30 09:10 6 -> socket:[3271768]
> lrwx------ 1 marko marko 64 Mar 30 09:10 9 -> socket:[3271503]



@littlesat is right

Dima73:
free up any system resources

milosoftware:
We've been through this before, and there have been more suggestions like this. Python is not Java. Python really releases things when they go out of scope. Calling close() just before the file object leaves scope is a pointless waste of code and effort. It doesn't do any good, and it harms by making the code larger and therefore slower.

If you want to have precise control over object lifetime, use the "with" statement.

Dima73:
Ok.Close.

PR #247, opened 2016-04-09
GMEPG: add key RADIO for show/hide window if not PIG and use preview

Dima73:


littlesat:
Sorry I do not understand the need?

Dima73:
When press red button(Zap to) or TV button(preview selected channel) live TV not visible if GMEPG screen not PIG window.Radio button hide GMEPG screen and live tv visible.

PR #248, opened 2016-04-10
Fix CI info - thanks betacentauri

Dima73:
CI info was not shown when debug log was disabled.
Seperating debug from program logic.


littlesat:
Will be merged a better way...

PR #250, opened 2016-04-13
JESS - improve NimManager code as suggested by athoik

arn354:


littlesat:
Seems to be already merged

PR #254, opened 2016-04-15
ATSC: Add missing support

athoik:
This commit add support for ATSC (Terrestrial and Cable) using devices supported by receiver kernel.

Reading atsc.xml reading/writing lamedb, scanning, piority, recording priority, alternative priority (mapped with DVB-C) and display proper tuner information were implemented.


littlesat:
I,m just thinking it mit be better to merge this in our next-master branch...

athoik:
This will make review by ATSC users harder since there is no next-master images. Also my/our job will be used by others maybe starting from today and not from us.

littlesat:
We're planning to create next-master hopefully within a few weeks... And due to jess and so I think it is better to push this in next-master

athoik:
Ok to the next-master when ready!

PR #256, opened 2016-04-16
epg cache: fixed out of bound write

stevenhoving:
Every once in a while I run cppcheck on the code and try to fix a couple of errors and create a pr for it. Usually I do not try to break or change to much. This time I fixed a weird possible out of bound write that happens when a sprintf creates a string that is smaller than 4 characters. When I was at it I saw some rouge crc functions and wondered why they where there to begin with. So I decided to clean it up...


stevenhoving:
@littlesat so the possible out of bound write commit causes issues. Can you tell me a bit more about it.

eriksl:
Yes Littlesat please explain.

eriksl:
I do have other objections to these patches b.t.w.

- too large, too much changes, break them up
- why not simply use a class for crc32
- writing directly to buffer of std::string

About the last remark: it will probably work, but of course it's as dirty as hell. The coder is not allowed to make assumptions about the internal representation of the data inside the string. Why not ditch the whole s(n)printf concept + fixed length strings and use stream semantics?

stevenhoving:
Thanks @ericsl, yea, i should have done it in more sepeated commits. I could create a crc32 class... 
About the sprintf, yea I could change it to stringstream. But lately I use https://github.com/c42f/tinyformat and I can say... Its really nice...

PR #267, opened 2016-04-22
ServiceMP3 enigma2 freeze by changing audio track while paused

christophecvr:
When media is paused and user tried to change the used audio track
 To the next avbl track enigma2 freeze stb freezed.
 To solve this we will pass to play state,
 change the track and pause again.
 works well on dm8000 and vuduo2 no sync issues.

### DIFF
--- a/lib/service/servicemp3.cpp
+++ b/lib/service/servicemp3.cpp
@@ -884,6 +884,12 @@ RESULT eServiceMP3::getLength(pts_t &pts)
 RESULT eServiceMP3::seekToImpl(pts_t to)
 {
 		/* convert pts to nanoseconds */
+	if (m_paused)
+	{
+		gst_element_set_state(m_gst_playbin, GST_STATE_PLAYING);
+		m_paused = false;
+		m_seek_paused = true;
+	}
 #if GST_VERSION_MAJOR < 1
 	gint64 time_nanoseconds = to * 11111LL;
 	if (!gst_element_seek (m_gst_playbin, m_currentTrickRatio, GST_FORMAT_TIME, (GstSeekFlags)(GST_SEEK_FLAG_FLUSH | GST_SEEK_FLAG_KEY_UNIT),
@@ -898,12 +904,6 @@ RESULT eServiceMP3::seekToImpl(pts_t to)
 	{
 		eDebug("[eServiceMP3] seekTo failed");
 		return -1;
-	}
-
-	if (m_paused)
-	{
-		m_seek_paused = true;
-		gst_element_set_state(m_gst_playbin, GST_STATE_PLAYING);
 	}
 
 	return 0;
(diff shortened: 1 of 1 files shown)


littlesat:
For latest gstreamer or also for the 0.1?

mx3L:
@christophecvr, @littlesat 
Sorry for jumping into this, but what about this one which I proposed in PM, which solves same issue and even fixes lost caps event?
https://gist.github.com/anonymous/55db2ce50621062bc2a6

I think it's not even a workaround just replacing order of flush and switch track.

christophecvr:
@littlesat I did not test on 0.1 , but I'm almost shure it's ok for 1.0 and 0.1 .

This issue is not directly gstreamer related, but cause we 
- need to work async by stb no other option.
- the problem was present on 0.1 , it also never worked ok on 0.1.
- I only found this ok solution on 1.0

christophecvr:
@mx3L 

gues I missed this pm or it most have been long ago.
Need to test it but I'm in doubt it will work however will try it out.

What I well can say strictly speaking if we would work only with video/audio. a seek flush is even not required up on change off audio track. But if working with /video/audio/subtitle its' a must.
and this for intern and extern subtitles and is also the cause off the pos - 90000 ns which was not brought in by me but by someone who worked on it long before me.

mx3L:
@christophecvr 
Look in Sequencing servicemp3 page 4, @MastaG is using it quite a long time in his images. It works, it also solve issue when you have different audio tracks like AAC and AC3 and when you are switching between them, you will not lose caps event in sink, so audio decoder gets correct streamtype with correctly written PES packets for both audio streams.

I didn't create pull request yet, because I don't understand, why exactly it solves the problem.

mx3L:
Here is what I wrote about patch in PM, but didn't get answer from Openpli guys @athoik and @littlesat 


> Please note that above patch was made by making assumption when dealing with missing caps event in audio sink, that something in gstreamer is dropping caps event when we first select audio track and then we drop it by flush. So I tested this assumption by first doing flushing and then change audio track. It solved the problem and it also solved the second one with deadlock. Problem is that I don't have any hard evidence to support my assumption, only that it works :).
> 
> So I'm not sure if you're willing to push it. I think both ways should work i.e. flush then change track or change track and then flush, but from a test it looks like there is only one way of doing it "right", 
> but I don't know why. I tried to debug it but didn't find where is dropping of caps event happening.
> 
> Should I create a pull request? Or do we need to find out what is the problem in gstreamer?

christophecvr:
@mx3L 

I never have had that patch. But Yess it does solve the freeze problem and indead it's just sequencing error. After You're patch the try to flush an non existend buffer is not there anymore and then we change track. I well did only a brutal test at this time, and only on dm8000. but there is still a small new error to solve after this patch . resume does not work ok anymore. But think i just found why. That will need more serious tests and try out's tommorow.

christophecvr:
@littlesat 

mx3l solution is also an option but ,... at this time the patch is not complete more needs to be done as it causes other isues. I think what it is but will work on it tommorow. 

So this pull request maybe removed.

And I work on the sequencing correction done by mx3L + extra correction off me.

christophecvr:
just to let the comments actif I reopend the pull request

christophecvr:
Think now all comments are read mx3L with some extra removal off work-arounds does it

PR #269, opened 2016-04-23
ServiceMP3 enigma2 freeze bug when paused

christophecvr:
When media is paused and user tried to change the used audio track
 To the next avbl track enigma2 freeze stb freezed.
 mx3L found out that problem is solved if we start with flushseek,
 then change the track problem is solved.
 Then removed the unneeded pas to playing back to stop on audio
 track change when paused.

	modified:   lib/service/servicemp3.h

### DIFF
--- a/lib/service/servicemp3.cpp
+++ b/lib/service/servicemp3.cpp
@@ -412,7 +412,6 @@ eServiceMP3::eServiceMP3(eServiceReference ref):
 	m_is_live = false;
 	m_use_prefillbuffer = false;
 	m_paused = false;
-	m_seek_paused = false;
 	m_cuesheet_loaded = false; /* cuesheet CVR */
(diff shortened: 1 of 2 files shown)


mx3L:
@christophecvr 
Are you sure that we want to remove m_seek_paused? It was added for this reason
https://github.com/OpenPLi/enigma2/commit/738e628e619e05c9df8df5007c5d8bafe68dc48d

christophecvr:
Yes 100 % shure enigma2 state is well updated.
The audio track is changed. All runs 100 % in sync with the new selected audio track. Chrosscheck show also that after you unpaused the media the right track last selceted track runs and is correctly showed as the actig one.

littlesat:
Closed as discussed via PM

PR #272, opened 2016-04-23
ServiceMP3 enigma2 freeze bug when paused.

christophecvr:
First changed sequence when changing audio track.
 First flush will be done then the trach change.
 Then relocation off the pas to playing before seekflush is done.
 This avoids a/v sync issues.

### DIFF
--- a/lib/service/servicemp3.cpp
+++ b/lib/service/servicemp3.cpp
@@ -884,6 +884,13 @@ RESULT eServiceMP3::getLength(pts_t &pts)
 RESULT eServiceMP3::seekToImpl(pts_t to)
 {
 		/* convert pts to nanoseconds */
+	if (m_paused)
+	{
+		m_paused = false;
+		m_seek_paused = true;
+		gst_element_set_state(m_gst_playbin, GST_STATE_PLAYING);
+	}
+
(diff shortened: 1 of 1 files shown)


mx3L:
Nice, I think it's a good idea to make sure pipeline is unpaused before seeking.

Some points:
1. I think purpose of m_paused flag is to know in which state we currently are. You're setting this flag before actual state change happens.
2. What if seeking is not successful? If you're in paused state it will unpause pipeline, which is not what we want if seeking is not successful, so I think we should check also for this condition.

Last, I don't want to sound rude, but I think it's polite to mention also my contribution, since part of this patch was done by me.

christophecvr:
1) ok there it's not important it will go in play at all times.
2) even if seek is not succesfull it will go back into pause.
    that is covvered by lines : case GST_STATE_CHANGE_PAUSED_TO_PLAYING:

christophecvr:
just closing to resent with updated comment for last point

mx3L:
1) yes, there shouldn't be side-effects, but I still think the purpose of this variable is that it's showing the actual state of pipeline, so it should be set only after state change happens and not before. If it's not making sense then OK.
2) you're right

PR #273, opened 2016-04-23
Service Mp3 enigma2 Freezing

christophecvr:
Done togheter with mx3l

### DIFF
--- a/lib/service/servicemp3.cpp
+++ b/lib/service/servicemp3.cpp
@@ -906,7 +906,7 @@ RESULT eServiceMP3::seekToImpl(pts_t to)
 		eDebug("[eServiceMP3] seekTo failed");
 		if (m_seek_paused)
 		{
-			gst_element_set_state(m_gst_playbin, GST_STATE_PLAYING);
+			gst_element_set_state(m_gst_playbin, GST_STATE_PAUSED);
 			m_paused = true;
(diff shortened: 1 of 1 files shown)


christophecvr:
However it's not really required I added the not seek succesfull case to reset the bin gst-state and flags where we where.

mx3L:
Uf, @christophecvr did you read what I wrote to your points in your last closed pull request?
https://github.com/OpenPLi/enigma2/pull/272#issuecomment-213734683

christophecvr:
1) pause flag adapted
2) Could be done there if the seek failed no problem it's a extra protection

mx3L:
1) looks like we don't understand each other, gst_element_set_state can be asynchronous call https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/GstElement.html#gst-element-set-state, so if it returns GST_STATE_CHANGE_ASYNC and you will set m_paused to false, it maybe not true yet, therefore I suggest to only set m_paused flag after state change happens, i.e. when state change message is posted on the bus.
2) ok, but more code

christophecvr:
Actually You want to wait on the state change message (playing) received before continuing with flush.

christophecvr:
However it works quit ok still not 100 %
For that I'll close the pull request for the time

PR #281, opened 2016-05-02
[ServiceScan] fix auto scan (thanks openATV developers)

Dima73:


littlesat:
The eTimer gives me a feeling this fix is not a real fix... it is more a work-a-round.... Can you please describe more what happens and why we need a timer here?

Dima73:
I just picked up without going into the code.
We spent a few months, but the error was not found.

Dima73:
Please see this http://forums.openpli.org/topic/40266-general-issues/?view=findpost&p=531929

littlesat:
I understand now what happens here... but I still have the idea it is 'driver' related... so it is still somehow a dirty work-a-round....

PR #291, opened 2016-05-15
E2 state update during paused whitout gst play/paused state change xx

christophecvr:
Set commit :
 https://github.com/OpenPLi/enigma2/commit/351a7ccabb94c43ab42ddb45a21da8f6826ec678
 For gst-1.0 only

	modified:   lib/service/servicemp3.h

### DIFF
--- a/lib/service/servicemp3.cpp
+++ b/lib/service/servicemp3.cpp
@@ -412,6 +412,7 @@ eServiceMP3::eServiceMP3(eServiceReference ref):
 	m_is_live = false;
 	m_use_prefillbuffer = false;
 	m_paused = false;
+	m_seek_paused = false;
 	m_cuesheet_loaded = false; /* cuesheet CVR */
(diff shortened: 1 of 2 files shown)


littlesat:
I prefer to put this in next-master.... as next-master is running on gstreamer >1.0 we also not need (and cleaned op) that conditional compiling there....

christophecvr:
no that commit is merged into master

another commit was merged into next-master the one who threads the rapid fingers issue or some mkv media which has been degraded in quality to have smaller media file problem .

christophecvr:
I just would well merge this since some users reported even image loss on gst-0.10 before this commit however in gst-0.10 there where always such problems when going to pause and unpause by mkv's.
the commit makes a perfect media behaviour when performing actions if paused on gst-1.0 but looks that it makes the already existing problems on gst-0.10 worse.

PR #307, opened 2016-05-29
add Megasat Unicable II LNB Diavolo

p1otr:
I have one of the missing LNBs and even though I can setup it via "user defined" option, it's much easier to simply select correct one from list of predefined ones.


arn354:
As pli has JESS not merged - the xml will not fit.

p1otr:
I removed all other LNBs then.

First 8 channels work as EN50494. If I'm not mistaken, that's the case for all EN50607 LNBs. Previous file (with all scrs and format attribute) worked fine for me (enigma2 didn't show src9 and later even though they're defined in XML)

PR #316, opened 2016-06-13
Translated more lines

budiarno:
Added more lines on Indonesian.


WanWizard:
@budiarno 

This merge has been reverted, for some reason the file was in the wrong position. Could you create a new update, but then with the file in the correct directory?

PR #321, opened 2016-06-17
fix mb micro rcu picture at wizard

miraclebox-git:


eriksl:
See 53562a4d04ca2b43df5a480bf7c135661f45b077

PR #324, opened 2016-06-20
Add support to Chinese charactor sets, Set default encoding by myself.

nx111:
hello!

I patch the project to support chinese charactor sets,such as gb2312/gbk/gb18030/big5.And,I can set default charactor set in encoding.conf, It fallback encoding when in dvb-text no encoding table is given.Just add a line as:
`     *     UTF-8`
This will set default encoding as utf-8.

And these encoding can be set in encoding.conf,in addition to ISO8859-x and ISO6937:
`GB2312 GBK GB18030 CP936 BIG5 CP950 UTF8 UTF-8  UTF16BE UTF16LE`

Among these encoding,GB2312/GBK/GB18030/CP936 use the same GB18030 charactor set,BIG5/CP950 use BIG5.


mirakels:
Why not replaice 'line'  inline instead of introducing an extra 'lowerline' variable. You can also remove the first check for '#'.

Also: if (scanf(...) && encoding == A  
           || scanf(...) && encoding == A)

can be simplified as
       if ( (scanf(...) || scanf(..)) && encoding == A)

PR #326, opened 2016-06-21
support chinese charactor encoding,and UTF16BE and UTF16LE. 

nx111:
hello!

I patch the project to support chinese charactor sets,such as gb2312/gbk/gb18030/big5.And,I can set default charactor set in encoding.conf, It fallback encoding when in dvb-text no encoding table is given.Just add a line as:
`* UTF-8`
This will set default encoding as utf-8.

And these encoding can be set in encoding.conf,in addition to ISO8859-x and ISO6937:
GB2312 GBK GB18030 CP936 BIG5 CP950 UTF8 UTF-8 UTF16BE UTF16LE

Among these encoding,GB2312/GBK/GB18030/CP936 use the same GB18030 charactor set,BIG5/CP950 use BIG5.


mirakels:
The patch to configure.ac will not be accepted. So your pull request cannot be satisfied.

Did you actually test your patch with varying data in de encoding file? The code to reduce the input line is buggy and may cause strange result... What is this 'start' variable suppose to do...

Maybe this could do the job:

```
-                       if ( line[0] == '#' )
-                               continue;
+                       int i, j = 0;
+                       for (i = 0; line[i]; i++) {
+                               if (line[i] == '#')
+                                       break; // skip comment
+
+                               if (line[i] < ' ' || (j == 0 && (line[i] == ' ' or line[i] == '\t')) )
+                                       continue; // skip control characters and leading whitespace
+
+                               line[j++] = isupper(line[i]) ? tolower(line[i]) : line[i];
+                       }
+                       if (j == 0)
+                               continue; // skip empty line
+                       line[j] = 0;

```


Also constructs like

```
+                       else if ( (sscanf( line, "0x%x 0x%x %s", &tsid, &onid, s_table ) == 3
+                                       && ( strcmp(s_table,"big5") == 0 || strcmp(s_table,"cp950") == 0 ) )
+                               ||(sscanf( line, "%d %d %s", &tsid, &onid, s_table ) == 3 
+                                       && ( strcmp(s_table,"big5") == 0 || strcmp(s_table,"cp950") == 0 ) )

```    


Can be simplified to:

```
+                       else if (       (sscanf(line, "0x%x 0x%x %s", &tsid, &onid, s_table ) == 3  ||
+                                       (sscanf(line, "%d %d %s",     &tsid, &onid, s_table ) == 3))
+                                       && (strcmp(s_table, "big5") == 0 || strcmp(s_table, "cp950") == 0) 
```

nx111:
simplified code in eDVBTextEncodingHandler, revert change in configure.ac.

`				if(line[start] > 0 && line[start] <= ' ' ){
					start++;
					continue;	//skip blank char, 'start' to find the offset of the first printable char
				}
				
				if(line[i] >= 'A' && line[i] <= 'Z')
					line[i]=line[i] + 0x20;		//lower case the char
			}
			if( i == start )
				continue;	//no printable char ,skip this line
`
here to find the offset of the printable char  by 'start',if it equal to the end of  line, indicate the line is blank,skip it.

mirakels:
so if the line has space after the the first non-space character, start will be increased and will not point the the start of printable code anymore.
Why do you want to skip non-printable characters at the beginning of the line and not further down the line?

Why not use the isupper() and tolower() functions? 
The sample code I hinted for checking the input line still looks more clear to me...

nx111:
I fixed it. thanks.

littlesat:
Can we merge it?

nx111:
yes
2016年7月10日 5:54 AM，"littlesat" <notifications@github.com>写道：

mirakels:
Sorry for the delay,
but in principle it could be merged, but the patches contains lots of reverts and fixes on fixes. Also in the  encoding.conf there seem to be 'null' changes:
   -0x2c1 0x600 #Thor 0.8°W
   ..  
   +0x2c1 0x600 #Thor 0.8°W

Please fix those and create clean patch set (new merge request). And also split the patches in logical steps. (for example, adding functionality, enabling this functionality, adding po support etc) So do not combine unrelated things in a single commit.

nx111:
OK

PR #331, opened 2016-06-28
Add menu_sort - thx vti

arn354:
1.) Set config.usage.menu_sort_mode to "user"
2.) Stay in menu tree
3.) Press Blue => Edit mode on
3.1) Select menu entry to edit or hide
3.2.1) Press Green to activate "Move mode", move menu entry to desired position and press Green again to exit "Move mode".
3.2.2) Press Yellow to hide/unhide selected menu entry
3.3) Press Blue to switch Edit mode off

Might be a good idea to add a additionel %Edit Screen to show colorbuttons once Editmode is enabled

PR #332, opened 2016-06-28
List.py - enhance functionality thx VTI

arn354:
This commit is necessary for menu_sort.
This solves also some issues with plugins which are mainly created for VTI.

### DIFF
--- a/lib/python/Components/Sources/List.py
+++ b/lib/python/Components/Sources/List.py
@@ -105,3 +105,30 @@ to generate HTML."""
 		self.list = list
 		self.index = old_index
 		self.disable_callbacks = False
+
+	def pageUp(self):
+		if self.getIndex() == 0:
+			self.index = self.count() - 1
+		elif self.getIndex() - 10 < 0:
+			self.index = 0
+		else:
+			self.index -= 10
+		self.setIndex(self.index)
+
+	def pageDown(self):
+		if self.getIndex() == self.count() - 1:
+			self.index = 0
+		elif self.getIndex() + 10 >= self.count():
+			self.index = self.count() - 1
+		else:
+			self.index += 10
+		self.setIndex(self.index)
+
+	def up(self):
+		self.selectPrevious()
+		
+	def down(self):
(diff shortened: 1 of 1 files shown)


mx3L:
Hello,

1. How do you know how many items has one page of List? I don't think it's right to just hardcode 10 entries for one page.

2. Why don't just use self.index which is defined as property which will call setIndex, getIndex methods, instead of calling them directly? or at least for cleaner code, we shouldn't mixed them up.

3. I'm wondering, what is the reason that these plugins don't use standard List Source functions selectNext, selectPrevious, getIndex and instead create and use aliases for them like down, up and getCurrentIndex.

RobvanderDoes:
This is (or should have been) part of this commit: https://github.com/OpenPLi/enigma2/pull/331#issuecomment-229058173.
As the latter is now in PLi and causes crashes, I suggest to merge this one as-is.
That whole new 'sort'-functionality needs cleaning up and enhancements anyway, so your suggestion could be part of that.

arn354:
I simply took over those functions.
Why they exist in VTI and why they were used by plugins of VTI (e.g. webradiofs) - i can't answer.

a) reject (then revert the menu_sort commit too)
or
b) merge and improve

RobvanderDoes:
LOL: webradiofs is not a VTi-plugin, but made by a rather eccentric coder. 
Remember: PLi tries 'clean coding', so it would be a good idea to enhance the lot.
But for now: please merge. And go on from there.

RobvanderDoes:
See http://forums.openpli.org/topic/41917-new-option-sort-menu/?p=555347

mx3L:
I think adding pageUp and pageDown is good suggestion, but aliases shouldn't be there.

This could be used as fix instead of patching List source:
https://gist.github.com/anonymous/c64e717f66777d2b5922ac44d54e0a9e

mx3L:
I don't understand, why merge something which is not correct and not to use suggested patch which does the same thing, but without polluting List Source?

PR #333, opened 2016-06-28
[SleepTimerEdit] add new option 'Power off timer'

Dima73:
-Select the days of the week to turn off
-Next day starts at (time)
-Force power off (even when not in standby)
-exceptionsExternalTools for exception time plugins, .e.g.
from Components.PowerOffTimer import exceptionsExternalTools
exceptionsExternalTools.append(("EPGRefresh", epgrefresh.isRunning))

### DIFF
--- a/lib/python/Components/Makefile.am
+++ b/lib/python/Components/Makefile.am
@@ -20,5 +20,5 @@ install_PYTHON = \
 	Ipkg.py SelectionList.py Scanner.py SystemInfo.py PackageInfo.py \
 	Task.py Console.py ResourceManager.py TuneTest.py \
 	Keyboard.py Sensors.py FanControl.py HdmiCec.py RcModel.py \
-	Netlink.py InputHotplug.py \
+	Netlink.py InputHotplug.py PowerOffTimer.py \
 	opkg.py
--- a/lib/python/Components/PowerOffTimer.py
+++ b/lib/python/Components/PowerOffTimer.py
@@ -1,0 +1,157 @@
+from RecordTimer import RecordTimerEntry
+from Screens.MessageBox import MessageBox
+from Screens import Standby
+from Tools import Notifications
+from Components.config import config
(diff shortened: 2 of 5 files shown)


littlesat:
Why do we need a poweroff timer???

We have already turn ON per day and the poweroff is auto (on activity).... I suggest this does not help at all...

Dima73:
This advanced functionality,
e.g. used as a server at certain times.
Start time 8:00 / end time 23:00

I think this option will appeal to many users.
I checked everything works fine, no errors noticed.

RobvanderDoes:
So IMHO it would be better to enhance the built-in functionality.

littlesat:
I do not ink many users will use this... So when we have a plugin I suggest leave it there... Whar a weird idea having a server on limited times.....

Dima73:
You decide for everyone.
Very sorry.

RobvanderDoes:
It's not a weird idea; it's a very good idea to power off devices that are not needed. In most households an STB/server/NAS/Amp/TV/PC/Laptop is only being used for a couple of hours per day. Letting the all work will cost an enormous amount of power per year. And that's exactly the reason that more and more people turn off devices (sometimes even by a complete power cut). So any device that could be automated to be on at times they are most likely to be used, and off at other times, will enhance proper use by people.

Dima73:
Yes, to save power.
In the world and so the resource crisis.

littlesat:
To save power we have already the activity timer and time to block this activity timer... That is why I do not understand a hard power down is needed....

Dima73:
I agree.
You can of course use the inactivity timer for this purpose.
But this is a direct option and it will find its user.

Dima73:
First step wake up timer.
Next step power off timer.
In my opinion it is logical.

RobvanderDoes:
Yep.
That's why many images have it that way.

PR #336, opened 2016-07-07
MBMICRO: KEY_TV --> KEY_TV2, add function toggleTvRadio, requiers driver 07.07.2016. 

miraclebox-git:


eriksl:
Refused in this form. No "if(box LIKE ...)" constructions in enigma2. I'd suggest to just add a key KEY_TOGGLE_TV or similar, make that send by the driver to enigma2 and just add handling for that in enigma2.

eriksl:
Above comment concerns 2caf9f2. The following two patches won't commit cleanly without 2caf9f2. Please rearrange (preferribly on a new clean branch).

miraclebox-git:
I will make clean pull request.

New patch does not use any kind of "if box LIKE".

Just added toggleTvRadio to KEY_TV2 and this is clean and good solution, even for future manufactuers who liked to use just TV/RADIO on same button.

PR #337, opened 2016-07-07
Invalid ONID/TSID on 7.0E

kam10:
Workaround for TP  scan overwrite on 7.0E for 11304H and 11262H with same ONID/TSID (126/30300).


eriksl:
Cleaned up and adopted.

PR #343, opened 2016-07-10
OSDPositionSetup: remove similar Overscan wizard

Taapat:
In a previous commit, I missed that in many places is used Overscan Wizard instead of Overscan wizard.
So change to Overscan Wizard to use only one translation for this.


littlesat:
We device that we prefer to use Overscan wizard... the second with small letters.... This for translations... Sorry

Taapat:
Ok, then need chage to Overscan wizard in all other places.

PR #344, opened 2016-07-10
skin_default: fix title position in MenuSort

Taapat:
Move title below the buttons, rather than above the buttons.


littlesat:
Why... can you post a screenshot on the forum?

Taapat:
I can post screenshot later here.
Check buttons size and position, and title size and position.
And you will understand without screenshot that they are on each other.

PR #349, opened 2016-07-12
Renderer [Picon] add option to replace type 4097 --> 1  for find picon

Dima73:


Taapat:
Why need orig = fields[2]?
Why simply not check fields[0] == '4097' before fields[2] != '2'?
https://github.com/Taapat/enigma2-openpli-vuplus/blob/master/lib/python/Components/Renderer/Picon.py#L79

littlesat:
Why replacing it by one???? You can also find picons with 4097 which makes them even more unique especially when you "use" the non used values...

Dima73:
Standard picon*. ipk not picons gstreamer type.
I do not see any problem, we first perform a standard search.

Taapat has long made such a patch, it is true idea.

Dima73:
same openATV
https://github.com/openatv/enigma2/blob/master/lib/python/Components/Renderer/Picon.py#L86

littlesat:
I think this is not a good idea...It is better to search on name for streams or keep the number and search also for it...

Dima73:
Why is this a bad idea?
This has long been used in other image, as required by the user.
In another Picon for type 4097 is not found, stream url changes often.
99% of users do not know how to change the references of manually.
After all, it is very convenient.

P.S.What is the cause of failure, because you do not need it?

Taapat:
If you change 4097 stream reference to sat channel reference you can use in this stream channel epg from sat channel.
So it's only logical, that you use also picion from sat channel.

Dima73:
I wanted to say, we use the EPG from the satellite for such channels.
One small replacement and we get ready to Picon.
Please apply this patch.

PR #359, opened 2016-07-23
[ChoiceBox] localize default WindowTitle 'Select'

Dima73:


littlesat:
Just push a slightly different solution... Thanks for reporting...

PR #367, opened 2016-08-01
Unicode/UTF16 string's first byte may less than 0x20, it will confuse to encoding id. Here is the fixes.

nx111:


mirakels:
nx111 in the old pull request I asked: 
can you explain why you add the 0x15 character to the front of string in convertDVBUTF8() ?
It seems to give some problems for some users...
char res[2048]; int t = 0; res[t++] = UTF8_ENCODING;

nx111:
Add the 0x15 character to the front of string in convertDVBUTF8()  to indicate that the output string is utf8 string,it will not be converted again.

Maybe it is unnecessary.

mirakels:
mapEncoding fix cherry picked.
Also took the encoding = -1 part out of the second commit as it is not related to this no_table_id.

Is no_table_id really needed? we llived without it for years...  And DVB standard defines that a first char < 0x20 indicates an encoding definition.
Otherwise it may be simpler to not use <ISOxxx> vut *ISOxxx
then the check can just be:

`
if (*s_table == '<') {
     noid = 1;
     s_table++
}
`

and no fooling around with copies of the encoding string are needed.

nx111:
Recoded somewhere in encoding.cpp.

UNICODE or UTF16 string 's first byte may be less than 0x20,if the string has not encoding id,but it is coding by unicode or utf16,convertDVBUTF8 may return incorrected utf8 string. 

Provide no table id config in encoding.conf and transport it to convertDVBUTF8, it will work correctly.

mirakels:
Ok, still a few comments. I don't like the '<' as a prefix. Isn't '*' better as it is also used to specify the default encoding, and this noid also server as a sort of forced default?

Create a #define NOTABLEID  0x80  and used that to set and check the encoding/table.
Make no_table_id an int and set it to 0 or NOTABLEID. Then do e.g. 'return BIG5 | no_table_id'
so at the end you can just return -1, and you do not need the last if statement 
Oh, and please watch coding style: space between ) and }
And indent by tabs, so do the add an i(notable_id) with a tab replaced by 4 spaces. Just add extra tabs to the other lines.
if (table >= 0  && table & 0x80)  could maybe be if (table >= NOTABLEID) ?

nx111:
Updated.

mirakels:
Hi nx111,
looks much better in my opinion.
Few small things
- I think we can remove the last 'else' in mapEncoding(). It does add nothing.
  The edebug and return can be backindented then and an empty line above the eDebug
- in estring.cpp the table &= 0x7F; should be coded using the NOTABLEID.
  e.g. table &= ~NOTABLEID;

Furthermore I think we need to wait a bit with applying this till we are sure the current code is stable enough....

itvro:
十分期待此更新整合到固件中去，我这有线中文EPG与台标仍然乱码。
感谢nx111大神的努力付出。
Thanks nx111 and mirakels very much!

nx111:
update again.

to itvro:
     You must edit encoding.conf youself.

itvro:
yes, i had edit encoding.conf and upload to usr/share/enigma2/ path
the encoding.conf modify like this:
*   GB18030

#TSID ONID

0x65 0x1c8e GB18030

0x66 0x1c8e GB18030

0x67 0x1c8e GB18030

0x68 0x1c8e GB18030

0x69 0x1c8e GB18030


***********************************
by the way ,can you tell me you qq number?
之前用过satfan 基于您的patch编译的enigma2进行替换，再按上面格式修改encoding.conf的方式，可以实现支持中文EPG与频道名称。但是7月24号之后就不能用那种方式替换了。所以麻烦nx111大师指点了。
英文不好，用了大段中文，呵呵，很多人都在支持你。

nx111:
to itvro:
   Q:381481281

mirakels:
first two commits incorporated in upstream

nx111:
first two commits  not incorporated in upstream yet.

mirakels:
sorry, still need to push. wnat to test locally, but need to rebuild buildsystem :(

nx111:
updated:
   To improve readablity,Add option 'N' or 'NOID' to the encoding name head splited by ':' , this indicates that the string has no  encoding id.

mirakels:
in that case the whole for lop can be replaced by just:
```
if (strncmp(s_table, "N:", 2) == 0 || strncmp(s_table, "NOID:", 5) == 0 ) {
				no_table_id = NOTABLEID;
				s_table = strchr(s_table, ':') + 1;
}
```

mirakels:
hi nx111, can you explain what is the reasoning behind your below commit:

https://github.com/OpenPLi/enigma2/commit/b55595e7ef7c18d05ae25497ec2ca38f289f6501

nx111:
when event text no encoding id ,
by “---”,converDVBUTF8 can get correct table by defaultEncodingTable that set in encoding.conf.

otherwise “eng”,c inverDVBUTF8 got ISO8859 always.

nx111:
updated.
in estring.cpp,just add 442-446,458,459 line.

isslayne:
Hi,nx111, I see you add code that supports  Chinese in your forked repository in June 2(support chinese charactor sets,such as gb18030 and big5. set default chraractor set by myself.);you're awesome!!!
last year,I use the STB to watch chinese program 115.5 satellite,the Chinese name is wrong code. So I want to  know whether you merge it into official master branch?

PR #375, opened 2016-08-08
Add "ud1" to Portuguese audio language choices.

lazaronixon:


mirakels:
Hi lazaronixon, can you explain why this ud1 is needed?

lazaronixon:
https://s9.postimg.org/vndyq0b3j/image.jpg

mirakels:
but shouldn't we have 
("por dub DUB ingles", _("Portuguese")),

too?

did you test this?

lazaronixon:
But "ud1" is Portuguese too ... 
Nowadays I need to select language manually "ud1" even that default language is portuguese. instead it ingles is selected automatically.
I was the author that added dub DUB on previous commits.

mirakels:
Hi lazaronixon, thanks for further explaining.
Can you revise the commit log to include this information so others can see the reasoning behind the update.

In general we like to see more extended commit logs than just oneliners that just tell in words what already is in the code to see. (and yes many of us do oneliners  but if we try to be more thoughtful, things may become much easier in the future to understand what is going on...)

Thanks

PR #382, opened 2016-08-18
AVSwitch: Add Multichannel PCM option

athoik:


littlesat:
Please move 
SystemInfo["MultichannelPCM"] = os.path.exists("/proc/stb/audio/multichannel_pcm")

to SystemInfo.py.... and consider to rename it HasMultichannelPCM.... and change path.exists to fileCheck so you can use that later...

PR #384, opened 2016-08-19
Support for multituner devices

athoik:
New tuners support multiple delivery systems like DVB-S2/S/T2/T2/C.

Till now Enigma2 consider that tuner can be S or T/C but never both.
This commit allow Enigma2 to use T/C with S when tuner supports it.

In order to use T/C channels you need to change tuner type once to T/C
and scan channels. Then switch to S. All channels will be available
without changing type.


littlesat:
I suggest for this -> In order to use T/C channels you need to change tuner type once to T/C
and scan channels. ... it is work in progress (needs more changes in python code)...

E.g. we need to add DVB/T or C options in satconfig/nimmanager when we have a combi tuner...
When we want to scan we need an option to select T/C or S...

As far I can see far far cleaner then the other solutions I saw before... ;)

PR #438, opened 2016-09-16
Enhanced SocketMMI.py with optional slot parameter

athoik:
Which will prevent crashes from Plugins built for the new API as Gutemine's AutoPin Plugin

More info: http://forums.openpli.org/topic/43561-screwed-socketmmipy/page-2

Related with issue: https://github.com/OpenPLi/enigma2/issues/397

### DIFF
--- a/lib/python/Plugins/Extensions/SocketMMI/SocketMMI.py
+++ b/lib/python/Plugins/Extensions/SocketMMI/SocketMMI.py
@@ -10,14 +10,22 @@ class SocketMMIMessageHandler:
 	def setSession(self, session):
 		self.session = session
 
-	def connected(self):
-		return socketmmi.getState(0)
+	def connected(self, slot=0):
+		return socketmmi.getState(slot)
 
-	def getName(self):
-		return socketmmi.getName(0)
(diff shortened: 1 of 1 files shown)


littlesat:
Do we need this autopin plugin?

littlesat:
Do we really need this?

athoik:
Depends who is "we", obviously you and me do not need this since we do not have a DMM receiver with CI.

eriksl:
We're going to add workarounds for gutemine's plugins???

athoik:
Do we need then socketmmi at all? If a user might need this is a DMM user that uses gutemine plugins.

BTW what issue #397 is for?

littlesat:
You need it when you have more then 1 ci in your box... But it is not the real fix. The real fix is embeded in dmms closed source cpp code. This is just the work-a-round that it does not give a gsod. Therfore I think we should not consider to accept this work-a-round. I prefer when we commit something like this we also should re-develop what dmm did put in their closed source and on top of that arrange that our ooen source autopin stuff is adjusted to that...

PR #439, opened 2016-09-16
[Ci] ci add option CIDelay for solo4k model

Dima73:


Dima73:
Not need move.When not module,this option not init.

littlesat:
Why open.. Do something, close?... And why needed?

littlesat:
I will merge it.... But one thing... And I know it still exist on other places... I prefer that the SystemInfo key settings should be moved to SystemInfo.py.... When I see you noticed this comment I will merge it...

PR #442, opened 2016-09-17
[vusolo4k] set 0 lcd brighthess when shutdown or reboot

Dima73:
-shutdown or reboot solo4k --> lcd display white screen, and then half
of a green screen ->it is very ugly


littlesat:
Why is e check on solo4k required? Isn't the check on the proc not enough?

Dima73:
Maybe this driver problem.
But no one is not fixed for the year.
But let's solve this problem in this way.
There are three lines of code, and the issue is closed.

littlesat:
I think not solve this cosmetic issue in enigma2 with a check on a specific box....

PR #446, opened 2016-09-18
[Ci] add option CIDelay for solo4k model

Dima73:


littlesat:
Why do you need this?

Dima73:
I do something wrong?

littlesat:
You're are not doing something wrong... i do not understand why it is required....

PR #448, opened 2016-09-19
[ChannelSelection] title - add "User - bouquets (TV)" to lokalize

Dima73:


Taapat:
I think it is not right.
And how about user radio bouquets?

I think that first of all need to correct removeModeStr.
Remove anymore unused translations from (TV) and (Radio).
And then create the conditions only for Bouquets or User - bouquets.
Or maybe we can use simply:
`if 'bouquets' in str.upper():`

littlesat:
I suggest it should be tolower this eay... ;)

Taapat:
Yes of course, str.upper(), sorry :(.

Taapat:
Maybe I miss something, but if we use the if 'bouquets' in str.lower(), then removeModeStr is not necessary and we can remove this function.

PR #458, opened 2016-09-24
[Ci] remove translatability to config dvbCiDelay

Dima73:


littlesat:
Thus can be even more simplified when both strings are equal you can just remove one... ;)

PR #473, opened 2016-10-09
servicemp3: improve audio and subtitle auto selection by language

Taapat:
Use enigma2 config.autolanguage settings to select audio and subtitle track depending on the language settings when start playback.


littlesat:
Sorry I really do not get what this tries to improve.... For me it is a ton of extra code with no real added value. Please explain in more detail what you're trying to improve...

littlesat:
OK... I see... it is for streams.... and mkvs....

PR #474, opened 2016-10-09
[ServiceInfo(Screens/Converter] fix vu+ aspect ratio for service 4:3 AVC

Dima73:
-More, more years that this service is 16: 9


athoik:
This seems like a workaround for VU+ drivers not beign able to detect correctly the aspect ratio. They should use PAR. 

I prefer not to merge this one.

Dima73:
This problem for several years.
VU + not solutions from there.

littlesat:
Maybe because "other" images do accept these kind of work-a-rounds there is no need for VU to adapt these kind of crap... I suggest we should not accept..!

RobvanderDoes:
This is one of the bugs all VU's have from the day the DUO was born, and one of the many bugs VU doesn't care about.

PR #482, opened 2016-10-12
[CableScan] Add decimal point to frequencies / update skin

Dima73:


littlesat:
This oatch has an issue... The frequency in the settings is an integer and nit a float.... It does also include two changes in one....

Dima73:
As will be right?

littlesat:
Sorry I saw you changed fequency to frequence...  This name is nog as wished but it will work a round the setting issue, but current config is gone and autocablescan will stop working until you reconfigure. I was looking in between for a solution for that config behavior...

What about the change in the id. I do not understand the reason why?

Dima73:
To inform the user.
Many users do not know what can be scanned without ID.

arn354:
why delete def keyCancel(self) if it is still used in self["actions"] ?

Dima73:
Use ConfigListScreen.keyCancel -->https://github.com/OpenPLi/enigma2/blob/53f768c987411b62eedfd105bdb9baac4be770be/lib/python/Components/ConfigList.py#L300

littlesat:
Can you please split this one into the all networks scan patch and skin changes patch... For the frequency I'm still investigating a different solution so it does not need a rename and is backwards compatible,with the current settings...

PR #484, opened 2016-10-13
[decoder] d'ont use audio when video decoder > 0, e.g.  start PiP

Dima73:


Dima73:
I do not understand.
I took the example here  #ifndef DREAMBOX

athoik:
This code should run for all boxes except vusolo4k or for vusolo4k?

Now it wont run for vusolo4k, if you need to run for vusolo4k you need ifdef, not ifndef.

Dima73:
It can is possible easier?
- sprintf(filename, "/dev/dvb/adapter%d/audio%d", demux ? demux->adapter : 0, dev);
+sprintf(filename, "/dev/dvb/adapter%d/audio%d", demux ? demux->adapter : 0, 0);

littlesat:
On what feature could we register this.... or is this another if VU+ then fix???

Dima73:
It seems only Solo+4K is /dev/dvb/adapter%d/audio1
dev == video_decoder --> for PiP use video_decoder = 1 --> so is open audio1
When stop service PiP -->stop audio1 (audio1 == audio0)

littlesat:
Can we give it a feature name?

Dima73:
I do not understand.

littlesat:
The patch says now Tuner solo 4k.... When accepted we should give it a feature name... When not vu needs to,solve their drivers if there is a real issue.... Sorry I also stilll ro not understand the real need for this..

Taapat:
For example the solo2 also has /dev/dvb/adapter0/audio1 but the sound with the PIP is working properly.
Is not better not to create a new eDVBAudio decoder if the audio decoder number is higher that 0, rather than create it with -1 values?

https://github.com/Taapat/enigma2-openpli-fulan/commit/513b727bfa4ec7aaffa36d2b31d2174f713a0131
(This commit I create because on the fulan receivers using PIP sound is on both channels.)

Taapat:
Maybe I miss something, but I really do not understand why need open /dev/dvb/adapter%d/audio%d, and /dev/dvb/adapter%d/demux%d and after set them to -1.
I think that in this case it is better simply not use eDVBAudio as I do for Fulan.

Dima73:
Taapat please add this patch

Taapat:
Maybe first of all we need to discuss here how to be better.

For example I have a question whether it is necessary to use an audio decoder which number is not 0?
If any receiver offers pip via scart, then definitely. But is it really work?

Dima73:
It worked many years ago at et9000 --> pip via scart --> SystemInfo["HasExternalPIP"] = not HardwareInfo().get_device_model().startswith("et9") and fileCheck("/proc/stb/vmpeg/1/external")
Now it does not work anywhere.

littlesat:
Please discuss further on the forum...

Taapat:
What you thing about this: https://github.com/OpenPLi/enigma2/pull/487

PR #485, opened 2016-10-13
[CableScan] update skin

Dima73:


littlesat:
Can we remove keyCancel?

Dima73:
Use ConfigListScreen.keyCancel -->https://github.com/OpenPLi/enigma2/blob/53f768c987411b62eedfd105bdb9baac4be770be/lib/python/Components/ConfigList.py#L300

PR #487, opened 2016-10-15
decoder: use audio only on primary decoder if external pip mode not used

Taapat:
This should fix the audio on vuplus and possible on other receivers after swap pip.
On vuplus lacks sound while not switch to another channel.

### DIFF
--- a/lib/dvb/decoder.cpp
+++ b/lib/dvb/decoder.cpp
@@ -845,7 +845,7 @@ int eTSMPEGDecoder::setState()
 	}
 	if (m_changed & changeAudio)
 	{
-		if ((m_apid >= 0) && (m_apid < 0x1FFF) && !noaudio && (m_decoder == 0 || m_pip_mode == "external"))
+		if ((m_apid >= 0) && (m_apid < 0x1FFF) && !noaudio && (m_decoder == 0 || eConfigManager::getConfigValue("config.av.pip_mode") == "external"))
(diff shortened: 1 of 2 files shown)


littlesat:
Apologies... me again...
Do we really need m_pip_mode? Can't we simply do:
if ((m_apid >= 0) && (m_apid < 0x1FFF) && !noaudio && (m_decoder == 0 || eConfigManager::getConfigValue("config.av.pip_mode") == "external"))

Taapat:
I think that it is better to use the function getConfigValue only once, to set m_pip_mode on eTSMPEGDecoder start, instead using this function at each channel change.

Taapat:
Yes, the config can be changed while the decoder is already running, but are external pip changes  working without enigma restart?

And as I understand, with the current drivers external pip do not work in et, therefore a possibility that anyone even uses this config actually does not exist.
Really therefore we should unnecessary call getConfigValue at each channel change?

Dima73:
checked, resolved to issue solo 4k

littlesat:
I still think we do not need m_pip_mode.... It does not make any sense... It is wrong... 
And indeed it seens with the latest already old drivers external pip is not working at all...

Taapat:
I changed it as you offeed.
But the reason is that by looking deeper, I see that config.av.pip_mode is not set on eTSMPEGDecoder start.

PR #498, opened 2016-10-26
[Channel selection context menu] Neat ordering of entries.

RobvanderDoes:


littlesat:
Most numbers already assinged were choosen via a linux standard... I would consider to keep at least the already numbered options... For the rest this looks nice ands gives a good feeling...

RobvanderDoes:
I have no idea what the 'Linux standard' is in this respect, but feel free to take the commit aboard and enhance as desired.

littlesat:
The keys that had already numbers, except for key 1 (parental) were based on linux standard.....

RobvanderDoes:
So again:

> I have no idea what the 'Linux standard' is in this respect, but feel free to take the commit aboard and enhance as desired.

RobvanderDoes:
And I really can't imagine that Linux has standards for 'use as startup service', 'Mark as dedicated 3D' etc. etc.

littlesat:
Use as startup has no,standard... But move, copy, delete, rename has....

PR #499, opened 2016-10-26
[Converter/ServiceInfo] clear all old video value if new service start

Dima73:


littlesat:
Do we need to clear for vu+ or is this idea beyond vu+?

Dima73:
for all box 
+ 
same wrong info in screen ServiceInfo

Taapat:
If it necessary evStart then evVideoSizeChanged does not work correctly.
If it is really necessary for all receivers maybe need to correct video size changed event in enigma?

Dima73:
no vu+
this wrong info from last service

Dima73:
If start new channel, but there is no picture then remains data of the previous channel

WanWizard:
Agreed. If you zap from a channel you can watch to a channel you don't, all info in the infobar still refers to the previous channel.

PR #506, opened 2016-10-31
eDVBServicePMTHandler: find first MPEG audio when cache is not available

athoik:
When cache is not available (eg on streams) and auto audio selection
is not possible because stream doesn't have any language info keep
the first MPEG audio as prefered source.


littlesat:
Why default on mpeg... Why not what is choosen as preferered system.?

athoik:
See here: http://forums.openpli.org/topic/44818-stream-relay-audio-channel-wrong

The problem seems to started since the AC3 included now in stream and Enigma2 has "sort memory" in strams.

littlesat:
Sorry I do not understand why we should prefer mpeg here?

PR #510, opened 2016-11-04
[ServiceInfo] sanity check for current play service

Dima73:
<   164.103>   File "/usr/lib/enigma2/python/Screens/ServiceInfo.py",
line 92, in __init__
<   164.104>     if serviceref and
session.nav.getCurrentlyPlayingServiceReference() != serviceref:
<
164.104> ValueError: invalid null reference in method
'eServiceReference___ne__', argument 2 of type 'eServiceReference const
&'


littlesat:
I think this is not required.....

But we still need to solve a GSOD.... watch a stream and go to the service info screen.... ;)

In ChannelSelection this is blocked.....

PR #512, opened 2016-11-04
ServiceInfo: optimization commit https://github.com/OpenPLi/enigma2/commit/514d579fbfb8d1598db6ed5e1a95e2117f55ac65

Taapat:
Why check serviceref and play_service, set serviceref to None, check it again and after change?
Instead of this check all necessary at once.


littlesat:
Thanks!

PR #514, opened 2016-11-08
[PictureInPicture] add notify  if use transcoding and start pip for s…

Dima73:
…olo4k

-PiP not avaible when use transcoding
- this added in drivers

20160516(solo4k)
- Support SD/HD Transcoding in live TV mode(mutually exclusive with Pip)
** Restriction : No Pip/MiniTV, Slow LCD Drawing

### DIFF
--- a/lib/python/Components/Sources/StreamService.py
+++ b/lib/python/Components/Sources/StreamService.py
@@ -1,5 +1,6 @@
 from Source import Source
 from Components.Element import cached
+from Components.SystemInfo import SystemInfo
 from enigma import eServiceReference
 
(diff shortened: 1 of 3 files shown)


littlesat:
We do not like this..

HardwareInfo().get_device_model().startswith('vusolo4k')

So I prefer to make a description/feature of it in SystemInfo,py...  (e.g. CanTranscodeAndPIPTogether) and cover it that way...
At least I try to collect all box "depended" cases in E2 python in this single file... (so when we get a different box with the same limitation we simply can add the box there and the "feature" is captured)

PR #517, opened 2016-11-10
frontend: fix fallback to FE_READ_SNR

athoik:
We need to make sure that provided value is not ovveriden once set
when DTV_STAT_CNR provides both DECIBEL and RELATIVE values.


athoik:
A newer simpler version was provided here: https://github.com/OpenPLi/enigma2/pull/518

PR #518, opened 2016-11-10
frontend: fix fallback to FE_READ_SNR

athoik:
Once we get the desired value from DVB API return it, so when no value
is found we are going to fallback to old DVB API.


pieterg:
I interpreted the dvb v5 stats documentation differently, up to 4 carrier groups can be described ('layers').

For delsys' with just one carrier group, I would expect only one stat to be present, with either DECIBEL or RELATIVE scale.

But as you interpreted it, yes, of course it might be that delsys' with just one carrier group would provide both scales...

don't see any examples if that though, in any of the current linux frontend drivers. But let's prepare for that anyway.

PR #522, opened 2016-11-14
pmt: read cached subtitle

athoik:
When pmtparser fails to find valid PMT or flag do not use dvb is used and our cache contain subtitle data add them to program.

TODO: handle teletext subtitles as well.

Note, in cache we do not store the language for dvb subtitles, so autoselection wont work.


littlesat:
I do not understan the advantage...subs can be there and not... And when it is not there for a specific program it can be gone, so no sub....

PR #527, opened 2016-11-17
[SleepTimerEdit] for "Enable wakeup timer" add more options

Dima73:
-options: Yes/No/Yes, only from standby/Yes, only from deep standby

### DIFF
--- a/lib/python/Components/UsageConfig.py
+++ b/lib/python/Components/UsageConfig.py
@@ -140,8 +140,11 @@ def InitUsageConfig():
 		("yes", _("Yes")),
 		("except", _("No, except Wakeup timer")) ])
 
-	config.usage.wakeup_menu = ConfigNothing()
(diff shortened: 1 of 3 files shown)


littlesat:
I think this is too much detail... Asking for even more detail....

Dima73:
for currentEventTime...

a) use self.session.nav.getCurrentlyPlayingServiceReference instead self.session.nav.getCurrentlyPlayingServiceOrGroup solved problem of receipt event if service alternative
b) if not remaining not need return config.recording.margin_after.value * 60, e.g. margin_after = 5 min.

for "Enable wakeup timer" value (goto box --> live tv)...
a)No (default)
b)Yes (deep standby/standby)
c)Yes, only from standby
d)Yes, only from deep standby
Here, all options are on the user's choice.

What is the problem?

PR #533, opened 2016-11-20
[skin] set default skin as primary if not PLI-HD and Magic

Dima73:


Taapat:
Do you also look at the code?
It is already here: https://github.com/O...er/skin.py#L104
And it's used in the case where there is an error on the DEFAULT_SKIN loading, not only if it does not exist.

Taapat:
Here just need to remove SD hardware fallback, and not to introduce additional unnecessary check.

PR #547, opened 2016-12-07
[hdmi cec] update python part

Dima73:
step 1 - update plugin

-don't use import Components.HdmiCec two twice
-save config only is changed
-create new setup only changed 'Enabled' entry


littlesat:
I really do not understand the need for this sorry.....

PR #551, opened 2016-12-09
add new hdmi option for vu 4k models

Dima73:


littlesat:
Why does this patch also adapt config.av.hdmicolorspace... is the change in colorspace 'valid' for the other boxes....?

Dima73:
choice Edid(Auto) is only VU+4k box
https://github.com/openatv/enigma2/blob/ac32b1c5f7e09e3fe65b76fd6dcb2401ae0da9de/lib/python/Components/AVSwitch.py#L600

PR #552, opened 2016-12-11
[NimManager] add sanity check wneh parse unicable.xml

Dima73:


littlesat:
I will not push it as the unicable.xml SHOULD be there.... when it is not there we SHOULD have an error....

PR #559, opened 2016-12-16
New option Mute the screen.

rocco99222:
Want to mute the screen without the sound. Please not advice to use (AV)receiver or other external sound system.

I request the feature to mute the video of the TV channel without the sound. When mute screen, want to display black screen, OpenPli bootscreen, own choosing image or the radio background. (preffer OpenPli bootscreen or own image)
Thanks


littlesat:
I do not zee a change here...?

PR #560, opened 2016-12-16
Feature request: Select own mount path.

rocco99222:
At VTi can choose own mount path, at OpenPli NOT. Want add OpenPli option to change mount path.
Thanks,


littlesat:
I do not zee a patch here???

PR #563, opened 2016-12-17
[Menu] "Sort order for menu entries" add option "Default with numbers"

Dima73:
-add show number entry before name
-changed number zap to two digits (pause 1.5 sec)


littlesat:
Nope... When you want to do something with numbers it should be a complete separate feature. Not "glue" it with another feature.... In addition I would suggest when using numbers we should not add numbers in the stream. Then use the graphic numbers instead...

PR #578, opened 2016-12-28
Add new option "Show picons in display", e.g.  et8500/solo4k

Dima73:
-To use this setting, you need to add in the skin attribute
isFrontDisplayPicon="1"
Example:
widget source="session.CurrentService" render="Picon" isFrontDisplayPicon="1"
zPosition="0" position="0, 0" size="480,320" transparent="1"
alphatest="blend"


littlesat:
The basic idea is good to make this "programmable"... but I think this patch needs some improvement...

self.show is not required. This could be requested by self.instance.visible
In addition I do not like the showPicon="0" attribute idea. The attribute is checked on "1" which we never will use.  I would prefer to set an attribute isFrontDisplayPicon="1" instead to indicate the icon is intended for the front display which adds the extra condition check on the config as requirement to display this picon.

Dima73:
done

PR #579, opened 2016-12-29
Update ru.po

Dima73:


eriksl:
Could at least one of you actually check what you request or merge? This one had a serious error that breaks the build. I've fixed it now.

Dima73:
I have checked the ru.po file in Poedit.
Everything was OK...
Sorry.

PR #586, opened 2017-01-01
Mapping full Miraclebox RCU

hawkeye116477:


littlesat:
Please do it not this way... Arrange it via hotkey instead...

PR #591, opened 2017-01-05
[SatelliteEquipmentControl] default onfig.sec.delay_between_diseqc_re…

Dima73:
…peats same as NimManager


littlesat:
Imdid change this back to as it was 120 as for some boxes 60 did not work.... But domyou think this solves the issue for formuler?

Dima73:
I do not know.
Just I wanted to make the same default.

littlesat:
Strange that in this plugin defaults are hardly set....

PR #592, opened 2017-01-06
Add converter VfdDisplay

Taapat:
This adds converter that you can use in skin_text.xml for 7 segment vfd displays.

By default you can use:
<widget source=session.CurrentService render=Label position=0,0 size=1,1>
    <convert type=VfdDisplay></convert>
</widget>
This show service number after the channel switching, and after 5 seconds show the clock.

Possible attributes:
  Number - show channel number
  Clock - show clock
  12h or 12H - show the clock in 12 hour format
  Noblink - do not show a blinking : between the digits (probably not supported by display)
  Nozero - do not show a zero before the channel number and the clock (64 instead of 0064)
  Loop - constantly change the channel number and the clock
  digit (2) - delay time in seconds (default 5)

Examples;
  Show only channel number:
    <convert type=VfdDisplay>Number</convert>
  Show only clock:
    <convert type=VfdDisplay>Clock</convert>
  Show channel number and after 10 seconds clock:
    <convert type=VfdDisplay>10</convert>
  Loop channel number and clock:
    <convert type=VfdDisplay>Loop</convert>
  Loop channel number and clock, dealy 10 seconds, do not show a zero before:
    <convert type=VfdDisplay>Loop;10;Nozero</convert>

### DIFF
--- a/lib/python/Components/Converter/Makefile.am
+++ b/lib/python/Components/Converter/Makefile.am
@@ -8,4 +8,4 @@ install_PYTHON = \
 	ServiceOrbitalPosition.py CryptoInfo.py TextCase.py \
 	ValueBitTest.py TunerInfo.py ConfigEntryTest.py ClientsStreaming.py TemplatedMultiContent.py ProgressToText.py \
 	Combine.py SensorToText.py ValueToPixmap.py PliExtraInfo.py genre.py TransponderInfo.py \
-	RotorPosition.py
+	RotorPosition.py VfdDisplay.py
--- a/lib/python/Components/Converter/VfdDisplay.py
+++ b/lib/python/Components/Converter/VfdDisplay.py
@@ -14,13 +14,13 @@ class VfdDisplay(Poll, Converter, object):
 		self.showclock = 0
 		self.delay = 5000
 		self.loop = -1
-		self.type = type.split(';')
-		if 'Number' in self.type and 'Clock' not in self.type:  # Only channel number
+		self.type = type.lower().split(';')
+		if 'number' in self.type and 'clock' not in self.type:  # Only channel number
 			self.delay = 0
 			self.poll_enabled = False
 		else:
 			self.poll_enabled = True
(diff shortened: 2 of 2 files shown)


littlesat:
if '12h' in self.type or '12H' in self.type:
-> if '12h' in self.type.lower():

Taapat:
self.type is list, so therefore can not be used lower()

if ... else It works much faster than and, and therefore I chose them.
If you can justify why the and is better (although it working slowly) I can change it.

Taapat:
About the first point. I think something like this before, but I do not know why, but I suddenly found a solution only after your comment. Thank you for your initiative.

The second I disagree with you, but you're the boss.

Dima73:
if /else one line...
What for?
Who needs it?
It is being implemented in all of the code of Enigma.
It is very difficult to read and understand.

littlesat:
Does it work now or does it as it also required a display_txt.xml

littlesat:
It is not diffucult to read it is so standaard python

Taapat:
It also required a skin_text.xml.

This is an independent converter that can be used in skin_text.xml or in any other skin which
introduce the same functionality as in the plugin VfdDisplay.
Without this converter can not make changes in the skin.
Therefore, I want to add their first so after that I could point to these changes.

Speaking about zgemma h3. First need to remove the plugin VfdDisplay from the image so that not conflict with skin_text.xml.
I've created a pull request, but it looks like that zgemma do not pay attention to it: https://github.com/zgemma-star/meta-zgemma/pull/6

After the update in zgemma bsp I can change skin_text.xml for zgemma h3 to use this converter.

Dima73:
This converter for the future.
All right.
Please apply this patch.

littlesat:
OK for the 'players'....

zgemma-star:
with this solution the skin_text.xml should been build against the 7segment display option.
There should come a generic solution (Machine-feature) and so that not all boxes should been add manual with it's own skin_text.xml

Taapat:
I agree. It would be better to use 7segment display option.

PR #594, opened 2017-01-10
configure: add vuuno4k to the models that need the event driver workaround

Taapat:


littlesat:
I agree... but when the next time there appears a new VU box we again need to change the same thing... that was why I considered to go for a when the string starts with vu construction....

Taapat:
Are you sure that the issue does not exist anymore for the Ultimo4K?
With the number buttons also works letters in all modes?

eriksl:
Created a new commit that also includes the Ultimo4k (for the future).

Thanks for the report!

PR #600, opened 2017-01-13
Support more audio/video formats

persianpros:
Amlogic devices (and generally arm SoCs) could handle more audio/video formats.


Taapat:
Or these new formats do not need add also in servicemp3.cpp https://github.com/OpenPLi/enigma2/blob/master/lib/service/servicemp3.cpp#L77

PR #607, opened 2017-01-13
Improve skin text for 7 segment displays

Taapat:
Add skin_text_h3.xml for zgemma h3.
Add skin_text_7segment.xml and configure option --with-7segment.
This allows use this skin for all receivers with 7 segment display without creating a skin for each receiver.


Taapat:
Please use together with this: https://github.com/OpenPLi/openpli-oe-core/pull/158

PR #609, opened 2017-01-14
[ScrollLabel] add sanity check when self.instance is not None

Dima73:


littlesat:
As far Imcan see this sanity check is not required at all...

PR #617, opened 2017-01-21
Add ChannelSelection in 7segment skin text

Taapat:
This show selected channel number in display when you in ChannelSelection screen


Taapat:
It looks like it is better without this.

PR #618, opened 2017-01-22
Simplify Font freetype 1 & 2 compatibility

stevenhoving:
Once in a while I like to spend some time on my favorite set-top box software project.
This time, while browsing the code and doing some cppcheck runs I stumbled upon a
ifdef construction that was wraps around function definitions of the Font class. And as
it seemed to me, its only purpose was to building enigma with freetype 1 or freetype 2.
So, by moving the type difference to a typedef we achieve exactly the same and have
more readable code as a nice bonus (code readability is actually the only motivation of
this pr).


eriksl:
Yeah, definitely better, all of it!

PR #627, opened 2017-01-29
[PositionerSetup] stop play service when use tuner config mode 'loopt…

Dima73:
…hrough/satposdepends


littlesat:
A bit pitty that there was a cpp cleanup (the ;) also inside this patch...  ;)

PR #630, opened 2017-02-05
[ServiceInfo] Avoid unexpected values in "self.getTitle()"

Dima73:
-thanks Huevos

### DIFF
--- a/lib/python/Screens/ServiceInfo.py
+++ b/lib/python/Screens/ServiceInfo.py
@@ -182,10 +182,12 @@ class ServiceInfo(Screen):
 				if self["key_blue"].text == _("Tuner setting values"):
 					self["Title"].text = _("Service info - tuner setting values")
 					self["key_blue"].text = self["blue"].text = _("Tuner live values")
+					data = True
 				else:
 					self["Title"].text = _("Service info - tuner live values")
 					self["key_blue"].text = self["blue"].text = _("Tuner setting values")
-				frontendData = self.feinfo and self.feinfo.getAll(self.getTitle() == _("Service info - tuner setting values"))
+					data = False
+				frontendData = self.feinfo and self.feinfo.getAll(data)
 				self.fillList(self.getFEData(frontendData))
 			elif self.transponder_info:
 				self["Title"].text = _("Service info - tuner setting values")


littlesat:
Why is this required? Sorry I do not see it....

Dima73:
I see this error, though not always.

littlesat:
When, How....? Without having an issue this change is not required....???

I can imagine when you use large menu paths it might go wrong... so that's why I recommended to use the blue text.... no extra variable required...

Dima73:
Use the blue text as True/False.Why so difficult.
It's so much easier - data=True / data=False

Taapat:
Maybe it would be better: 
```
 	def ShowTransponderInformation(self):
 		if self.type == TYPE_SERVICE_INFO:
-			if self.feinfo and self.feinfo.getAll(True):
+			frontendData = self.feinfo and self.feinfo.getAll(True)
+			if frontendData:
 				if self["key_blue"].text == _("Tuner setting values"):
 					self["Title"].text = _("Service info - tuner setting values")
 					self["key_blue"].text = self["blue"].text = _("Tuner live values")
 				else:
 					self["Title"].text = _("Service info - tuner live values")
 					self["key_blue"].text = self["blue"].text = _("Tuner setting values")
-				frontendData = self.feinfo and self.feinfo.getAll(self.getTitle() == _("Service info - tuner setting values"))
+					frontendData = self.feinfo.getAll(False)
 				self.fillList(self.getFEData(frontendData))
 			elif self.transponder_info:
 				self["Title"].text = _("Service info - tuner setting values")
```

littlesat:
Why add extra variable when it is not required?
I'm afraid Taapat's solution is not what is intended here...

Dima73:
ok,close

Taapat:
If you're talking about my offer then I did not add the extra variable.
frontendData already is used.
I think that even worse is in addition use functions self.feinfo.getAll and self.getTitle as it is now.
I propose to use additional self.feinfo.getAll only if need data from the tuner, and use existing data from already used function self.feinfo.getAll.

Do you really think that a good solution is **additional**:
`frontendData = self.feinfo and self.feinfo.getAll(self.getTitle() == _("Service info - tuner setting values"))`

mirakels:
I agree with Taapat's code. It makes it much more readable and avoids unnecessary calls (for data you already requested).
Also it avoids duplicate strings eventually leading to bugs: supposes i the future one renames
     "Service info - tuner setting values"
to 
    "Service info - Tuner setting values"
it is easy to forget the same strings in the 
    self.getTitle() == ...
sniplet

Dima73:
Example:
localization bla-bla-bla.po
msgid "Tuner setting values"
msgstr ""

always return self.getTitle() = ""

I do not understand your disputes.
I have added only two lines !!!!

PR #635, opened 2017-02-07
Return back duplicate translation in the config

Taapat:
This can be useful after you change the language in the start wizard.
It does not provide a full translation for all.
It works only in one direction- from the English to other language.
But it is very likely that in the start wizard you need to translate only the config values after language change from English to your local.

### DIFF
--- a/lib/python/Components/config.py
+++ b/lib/python/Components/config.py
@@ -350,11 +350,15 @@ class ConfigSelection(ConfigElement):
 	def getText(self):
 		if self._descr is None:
 			self._descr = self.description[self.value]
+		if self._descr:
+			return _(self._descr)
 		return self._descr
 
 	def getMulti(self, selected):
 		if self._descr is None:
 			self._descr = self.description[self.value]
+		if self._descr:
+			return ("text", _(self._descr))
 		return ("text", self._descr)
 
 	# HTML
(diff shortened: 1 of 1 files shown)


littlesat:
It does also not work correctly in one way... there are some stings with %s etc... that do not work... the only thing what really works is an enigma2 restart...

Taapat:
I agree with you that the only thing what really works is an enigma2 restart.
I added this by Dimitrij request.
But as I wrote, possible that almost all config translation, which is necessary in the start wizard working.

littlesat:
Is it OK I close this merge request... and again sorry I did close per accident the other one related to the service info....

Dima73:
When use start wizard need config translation.
1)install new image
2)use start wizard
3)select language , e.g. Russian/China
4)configure tuners -->sat config  language is English(need Russian/China)
Ordinary users do not know English.

littlesat:
This is what we should do....

When use start wizard need config translation.
1)install new image
2)use start wizard
3)select language , e.g. Russian/Chin
-----> NOW RESTART ENIGMA 2 EN CONTINUE WIZARED. <-----
4)configure tuners -->sat config language is English(need Russian/China)
Ordinary users do not know English.

Dima73:
Again...
 sat config language is English
Ordinary users do not know English.

littlesat:
as far I know satconfig is in local language..... once you do a reboot post you set the language...

Dima73:
When use wizard message RESTART ENIGMA 2  not visible.

littlesat:
We could not make that visiable or add it...

Dima73:
Let's return everything back as use ALL the other image.
You see, many people do not speak English.
For whom we make the image?

littlesat:
Make it right... Do a restart e2 and then continie the wizard....

Dima73:
New language need before continie the wizard

littlesat:
We can do two things.... Merge this commit and restart enigma2 after then wizard is done... Or restart enigma2 adter the language is changed... But still via the red key uou can instantly change the language which never can translate back fromma different language the english....

Dima73:
Yes, that is correct.

Dima73:
## littlesat
Please apply this patch.

littlesat:
I will not merge that... I hope I have time tomorrow to make a real solution...

PR #636, opened 2017-02-07
ServiceInfo: small optimization in ShowTransponderInformation

Taapat:
Avoids unnecessary function calls.
Call self.feinfo.getAll(False) only if need data from the tuner.

### DIFF
--- a/lib/python/Screens/ServiceInfo.py
+++ b/lib/python/Screens/ServiceInfo.py
@@ -178,14 +178,15 @@ class ServiceInfo(Screen):
 
 	def ShowTransponderInformation(self):
 		if self.type == TYPE_SERVICE_INFO:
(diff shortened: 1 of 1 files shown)


mirakels:
I think you should credit Dima73 in the commit log for pointing in this direction in the first place. (#630)

Taapat:
You can add this on merge.

littlesat:
I still do not understand why this is required....  With this patch you cannot toggle anymore at all as far I can see...

Taapat:
This is necessary because using 
`self.feinfo and self.feinfo.getAll(self.getTitle() == _("Service info - tuner setting values"))` insted of `self.feinfo.getAll(False)` is unreasonable and strange.

You are wrong if you think that it affects the toggle.
For toggle it is used self["key_blue"].text and it is not changed here.

PR #637, opened 2017-02-07
[frontend] add snr calculation Astrometa/os mega

Dima73:
-thanks openATV

### DIFF
--- a/lib/dvb/frontend.cpp
+++ b/lib/dvb/frontend.cpp
@@ -1050,6 +1050,15 @@ void eDVBFrontend::calculateSignalQuality(int snr, int &signalquality, int &sign
 		default: break;
 		}
 	}
+	else if (!strcmp(m_description, "Broadcom BCM73XX") ||
+			 !strcmp(m_description, "FTS-260 (Montage RS6000)") ||
+			 !strcmp(m_description, "Panasonic MN88472") ||
+			 !strcmp(m_description, "Panasonic MN88473"))
+	{
+		ret = snr * 100 / 256;
+		if (!strcmp(m_description, "FTS-260 (Montage RS6000)"))
+			sat_max = 1490;
+	}
 	else if (!strcmp(m_description, "Si216x"))
 	{
 		eDVBFrontendParametersTerrestrial parm;


mirakels:
Maybe this is nicer:

```
if m_description in ("Broadcom BCM73XX",
                               "FTS-260 (Montage RS6000)",
                               "Panasonic MN88472",
                               "Panasonic MN88473"
                               ):

```

athoik:
Please do not merge this one is not ok.

We are using DVB API 5, why do we need the wrong calculations from DVB API 3?

Dima73:
1)why DVB API 3 wrong?
2)DVB API 5 calculations really wrong for os mega/sundtek
 can not return a value without checking them 
https://github.com/OpenPLi/enigma2/blob/f544402e58516905b91b65fcb319e719d0c87238/lib/dvb/frontend.cpp#L1184

And please do not say that driver developers are required to give the correct value, because they are making money.

athoik:
Please read here: https://linuxtv.org/downloads/v4l-dvb-apis/uapi/dvb/frontend-stat-properties.html

FE_SCALE_RELATIVE - parameter is a unsigned value, where 0 means 0% and 65535 means 100%.

Explain now what is wrong.

Dima73:
os mega Broadcom BCM73XX  always return 99%
os mega multituner dvb-t type  always return 99%

Dima73:
need add if(!signalquality && !signalqualitydb || signalquality != signalqualitydb)

athoik:
If the driver returns DVB API 5 statistics you will never get into (fallback) DVB API 3.

If the driver returns wrong DVB API 5 statistics, drivers should be fixed, not Enigma2.

So the core you are proposing will never work, because DVB API 5 has preference over DVB API 3.

Dima73:
> 
If the driver returns wrong DVB API 5 statistics, drivers should be fixed, not Enigma2.
> 
Serious?
I want and I can do the correct values himself, now and for all users.
Let us then apply DVB API 3 and values as they are, without workaround.
Then we write requests to manufacturers.
And users can not care, let them see the deceptive values.

This was done with Sundek,thank you very much for that.

PR #645, opened 2017-02-11
Add possibility to remove native key binding

persianpros:


littlesat:
Why should you need this?

persianpros:
:( github is such pain in ... because it automatically merges new commits in last merge request and I have to wait for it to be merged or closed.
Please tell me which ones do you accept so I can send them one by one from different forks.

littlesat:
All of this is currently a bit too much to evaluate.... It might have a big impact and it does not help to finalyse other stuff (like openpli 5)

persianpros:
As far as I can see they are all safe to merge because cleanups are not anything new and the rest won't be used at the moment and they're  just new functions.

littlesat:
Can you please redo these commit requests.?

And still please clarify why we should unbind key stuff?

PR #646, opened 2017-02-11
servicem2ts: try to open as udf also img and nrg images

Taapat:
I do not know whether they exist, but this commit it provides: https://github.com/OpenPLi/enigma2/commit/8bd6209812275e7b439e00884148213425afcdce


persianpros:
Excellent ! thanks a lot :)

PR #657, opened 2017-02-13
Add function to add/release input devices on demand

persianpros:
Also enable rc debug for testing remotes


mirakels:
Now we have the /dev/input/event scan code at two places. Not good!
If we need scans and releases in separate functions then also use those functions in the first place (e.g. in eInputDeviceInit()  and ~eInputDeviceInit()

Also the mapping from PLAY to PLAYPAUSE should preferably be done be keymap files and selecting the correct keymap with each remote control.
And definae;ly it should be different commit!

PR #658, opened 2017-02-13
configure.ac check for msginit/msgmerge/msguniq/xgettext too, We need…

persianpros:
… these commands for pot/po creation and updating translations

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -16,6 +16,10 @@ AC_CONFIG_MACRO_DIR([m4])
 AC_OPENMP()
 
 AC_PATH_PROG(MSGFMT, msgfmt, AC_MSG_ERROR(Could not find msgfmt))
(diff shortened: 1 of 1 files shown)


littlesat:
Please clarify the need...

persianpros:
It's already in the commit : check for msginit/msgmerge/msguniq/xgettext too, We need these commands for pot/po creation and updating translations.

Taapat:
Did you checked configure.ac, or simply copied these changes from somewhere?
All of this already is in configure.ac if you compile with the option --with-po: https://github.com/OpenPLi/enigma2/blob/master/configure.ac#L133

persianpros:
Hmm, I think the new changes would apply without the need of --with-po am I wrong ?
And which one is better ?

Taapat:
These tools are required only if you compile with the flag --with-po.
Why need check for them when not using them?

PR #659, opened 2017-02-13
More accuracy in detecting WLAN brands

persianpros:
We're supporting so many WLAN brands but we're not showing them to the users and I don't know why.


littlesat:
Isn't it better to create a dictionary table to quickly search the full brandname.... (and when it is not in the dictionary default to "Unknown"  or keep what was given).... instead of creating a huge if, then..

persianpros:
Up to you, It's what I use and I thought it could be good for PLi too.

littlesat:
I hoped you can make it better.... ;)

PR #661, opened 2017-02-13
Cleanup python files #1

persianpros:
I did add "Fix forgotten parentheses" to the last merge but you just closed it.

### DIFF
--- a/RecordTimer.py
+++ b/RecordTimer.py
@@ -45,7 +45,7 @@ def parseEvent(ev, description = True):
 	eit = ev.getEventId()
 	begin -= config.recording.margin_before.value * 60
 	end += config.recording.margin_after.value * 60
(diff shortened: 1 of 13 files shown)


RecordTimer.py
```diff
@@ -45,9 +45,12 @@ def parseEvent(ev, description = True):
 	eit = ev.getEventId()
 	begin -= config.recording.margin_before.value * 60
 	end += config.recording.margin_after.value * 60
-	return (begin, end, name, description, eit)
+	return begin, end, name, description, eit
```
Taapat: It definitely will cause GSOD for example here: https://github.com/OpenPLi/enigma2/blob/8a6c57c705422ed5d056be9d11d627d9d0b6ffac/lib/python/Screens/InfoBarGenerics.py#L2396

PR #663, opened 2017-02-13
[About] crashlog - add folder /tmp/enigma2_crash.log for search

Dima73:


littlesat:
As far I know not required as the crashlog is not written in /tmp.....?

Dima73:
https://github.com/OpenPLi/enigma2/search?utf8=%E2%9C%93&q=%2Ftmp%2Fenigma2_crash.log

littlesat:
And when do you think this might happen?

Dima73:
Maybe directory /home/root does not exist or read-only...

PR #664, opened 2017-02-14
Cleanup python files #1

persianpros:


littlesat:
Are you sure the source.close() is required in keymapparser?
In e2reactor.py I would consider to indent the two error..... lines.
is 
class AFTEREVENT:
 +	def __init__(self):
 +		pass
 +
really required (was simply pass not sufficiant)?

persianpros:
Yes we have to close the files otherwise for some STBs specially the ones with low flash it wil end up in I/O errors because too many files are open.
When we can close it right after we're done with it why not just do it ? This way we can be sure about it not hope for it to be closed.

littlesat:
This is not requiredd at all...

littlesat:
That is not requried at all...

PR #666, opened 2017-02-14
Cleanup python files #3

persianpros:


littlesat:
-			open(test_name + ".bogus_results", "wb").write(results)
 +			f = open(test_name + ".bogus_results", "wb")
 +			f.write(results)
 +			f.close()
is no improvement.

Is
 	class LexicalHandler:
 -		pass
 +		def __init__(self):
 +			pass
really required ?

persianpros:
Yes we have to close the files otherwise for some STBs specially the ones with low flash it wil end up in I/O errors because too many files are open.
When we can close it right after we're done with it why not just do it ? This way we can be sure about it not hope for it to be closed.

PR #668, opened 2017-02-14
Cleanup python files #2

persianpros:


littlesat:
class additionalWidget:
 -	pass
 +	def __init__(self):
 +		pass

is not realy required

-		return (SizeTuple(pos), SizeTuple(size))
 +		return SizeTuple(pos), SizeTuple(size)
2x is not good

persianpros:
I don't have write access here but you do, Simply edit that line then merge it :)

PR #672, opened 2017-02-15
Use standard "FuzzyDate.py" display in "MovieList.py"

persianpros:


lib/python/Tools/FuzzyDate.py
```diff
@@ -20,17 +19,15 @@ def FuzzyTime(t, inPast = False):
 		# same year
 		if inPast:
 			# I want the day in the movielist
-			date = "%s %d.%d" % (dayOfWeek[d[6]], d[2], d[1])
+			date = _("%s %02d.%02d.") % (dayOfWeek[d[6]], d[2], d[1])
```
Taapat: dayOfWeek return already translated value.
Translating the second time already translated values you can get unexpected results.

lib/python/Tools/FuzzyDate.py
```diff
@@ -20,17 +19,15 @@ def FuzzyTime(t, inPast = False):
...
-		date = "%d.%d.%d" % (d[2], d[1], d[0])
-
-	timeres = "%d:%02d" % (d[3], d[4])
-
-	return (date, timeres)
+		date = _("%02d.%02d.%d") % (d[2], d[1], d[0])
 
+	timeres = _("%02d:%02d") % (d[3], d[4])
 
+	return date, timeres
```
Taapat: This certainly will cause GSODhere: https://github.com/OpenPLi/enigma2/blob/master/lib/python/Tools/FuzzyDate.py#L39
and here: https://github.com/OpenPLi/enigma2/blob/master/lib/python/Tools/FuzzyDate.py#L41

Did you check what you offer?

PR #673, opened 2017-02-15
Cleanup GetEcmInfo.py

persianpros:


littlesat:
The only cleanup here is the dictonary... But I'm doubting it makes it more readable...

PR #674, opened 2017-02-15
Cleanup Profile.py

persianpros:


littlesat:
this does not cleanup anything... It can stay in one line...

PR #683, opened 2017-02-15
Cleanup ValueToPixmap.py

persianpros:


littlesat:
Sorry... i do not understand the real need... Beside of course the fileExist import can be removed

littlesat:
And regards the open files.... on the one liners the file is directly closed.....

persianpros:
It will search for country flags in default and current skin so skinners have the ability to introduce country flags

stevenhoving:
So skinners do not have the option to supply there own country flag images?
If that is the case, why don't you do it something like this?

```
# check skin image path
filePath = resolveFilename(SCOPE_CURRENT_SKIN, "countries/" + val + ".png")
if fileExists(filePath):
    return LoadPixmap(cached=True, filePath)
# check default skin image path
filePath = resolveFilename(SCOPE_CURRENT_SKIN, "countries/" + val[3:].lower() + ".png")
if fileExists(filePath):
    return LoadPixmap(cached=True, filePath)
# skin image missing country flag
filePath = resolveFilename(SCOPE_CURRENT_SKIN, "countries/missing.png"
return LoadPixmap(cached=True, filePath)
```

The advantage of this, is that it will first check the skin's country flag image existence, falls back to the default skin country flag image. And if that doesn't exist load the missing image flag.

littlesat:
Why should a skin need it's own flags...?

stevenhoving:
@littlesat good points, even the developers of kodi (formerly known as xbmc) recognized the issue of skin's being packed with loads of duplicate resources. They actually made it a separate skin dependency and as a result everybody started using them. But to answer your question (although you asked @persianpros the same question) I think a skinner would like to use its own flags, just because he can. Or maybe because he does not agree with the style or image resolution of the default country flag images (hence he should solve that).
In other words, I don't think there is a good reason for skin's to pack there own set of country images. And if the skinner wants a different style or higher image resolution he should create a pr to fix that. That way the other skins can profit from the improved country images as well.

littlesat:
Clear.... But the way it is made is not really suffician... It also does multiple checks... When a skin wants to use it's own flags then I suggest we could use the parameter stuff of a skin.... And not parse first for a flag the skin directory and the scan were the flags are now... Sure it asks the skin to add something in the parameter part... But we will get faster and smarter code in return...

littlesat:
Sorry I closed it...

PR #685, opened 2017-02-15
[SoftcamSetup] remove unnecessary import / simple restart service

Dima73:
-


littlesat:
Again a dubble commit... Why do we need try except when we know the plugin exists...?

Dima73:
This is for the safety of users.
Simple practice.
This third-party plugins.

PR #687, opened 2017-02-17
Cleanup FrontpanelLed.py

persianpros:


littlesat:
I do not think this is a real cleanup

PR #688, opened 2017-02-17
Cleanup Picon.py

persianpros:


littlesat:
Why?

persianpros:
It's a fix for searching mounted devices like hdd and usb

littlesat:
Sorry Imdo not understan the need.... As far I know mointed devices do work as it should be...

PR #690, opened 2017-02-17
Cleanup Progress.py

persianpros:


littlesat:
Nit the correct method... Look in InfobarGenerics how you domreally check maxint...

PR #691, opened 2017-02-17
Cleanup VideoSize.py

persianpros:


lib/python/Components/Renderer/VideoSize.py
```diff
@@ -20,7 +20,7 @@ def changed(self, what):
 			return
 		xresol = info.getInfo(iServiceInformation.sVideoWidth)
 		yresol = info.getInfo(iServiceInformation.sVideoHeight)
-		if (xresol>0):
+		if xresol>0:
```
Taapat: And why not?
`if xresol > 0:`
instead of
`if xresol>0:`

persianpros: Git members could do it easily ;)

PR #702, opened 2017-02-17
Cleanup GUISkin.py

persianpros:


littlesat:
Are younsure this is ok?

PR #706, opened 2017-02-17
Cleanup HdmiCec.py

persianpros:


littlesat:
The firet part is not ok... As it still allows multiple instances... The second part is ok

PR #711, opened 2017-02-17
Cleanup MediaPlayer.py

persianpros:


littlesat:
Second part not ok....

PR #712, opened 2017-02-17
Cleanup MovieList.py

persianpros:


littlesat:
Only the last part ok...

PR #725, opened 2017-02-19
Cleanup timer.py

persianpros:


littlesat:
As far I can se the last change is not ok...

PR #746, opened 2017-02-19
[StbHardware/Timezones] add func setRTCoffset

Dima73:
- some reciever use /proc/stb/fp/rtc_offset  for correct local time

### DIFF
--- a/lib/python/Tools/StbHardware.py
+++ b/lib/python/Tools/StbHardware.py
@@ -1,6 +1,6 @@
 from fcntl import ioctl
 from struct import pack, unpack
-from time import time, localtime
+from time import time, localtime, gmtime
 
 def getFPVersion():
 	ret = None
--- a/lib/python/Tools/StbHardware.py
+++ b/lib/python/Tools/StbHardware.py
@@ -1,5 +1,6 @@
 from fcntl import ioctl
 from struct import pack, unpack
+from time import time, localtime
 
 def getFPVersion():
 	ret = None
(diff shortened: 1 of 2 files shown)


lib/python/Tools/StbHardware.py
```diff
@@ -1,5 +1,6 @@
 from fcntl import ioctl
 from struct import pack, unpack
+from time import time, localtime
```
athoik: Did you miss gmtime here?

littlesat:
I'll wait for the merge until adhoik's comment is checked and we have green light or a fix...

littlesat:
Can I have a better description for what it exactly is doing....?

Dima73:
Sets the difference for your time zone.

athoik:
What about boxes without rtc_offset? How those boxes work?

Dima73:
I do not know 
mine so the time zone is adjusted automatically.

littlesat:
I think e2 compensates for timezone so no need to additionally do it for rtc...

Dima73:
Wnen use only rtc time answer rtc incorrect

littlesat:
Sorry I still don't underatand explain with examples....

Dima73:
Example...
 timezone + 2
date 01.01.1970 02.00
rtc time 01.01.1970 00.00
When use rtc time ase2 time need first add rtc_offset((localtime(time()).tm_hour-gmtime(time()).tm_hour)*3600)

littlesat:
Need to finetune this later....

PR #749, opened 2017-02-20
[A/V settings] add option 'Audio volume step size' limit (1-10)

Dima73:


littlesat:
Why should you do this???? i think this make no sense at all....!!!

persianpros:
I guess with this we can set how much we want to increase or decrease the volume right ?

Dima73:
Current step size is 5( 0-100).

persianpros:
It's really a good option in my opinion :)

Taykun345:
Seems like useful setting.

littlesat:
Current step size is 5( 0-100).
->
I you sure I cannot find it in the code (sorry)

littlesat:
Sorry I'm a bit lost here... in volume control I see +/- 1....

littlesat:
Actually there are 64 steps (0..63) Then they put it in a scale of 100 (0..100)... and then.... reduce this....
Looks like 3 steps of conversion... this is really "crasy"....

littlesat:
Now I found were the steps 5 are made....

Don't you thing we should go for 0..63 range instead of going from 0..100 first....?

littlesat:
But then something else kicked in what requires 0..100..... :(

littlesat:
Sounds like it is 'stolen' from here...
http://git.opendreambox.org/?p=enigma2.git;a=commitdiff;h=477209fc58d324d75a62f87f0502ddd298c04161
So we even can't use it...

Dima73:
Yes, I took the example of the DMM.
This is the easiest option, openATV option is very difficult.

littlesat:
We have to copy it with care.... And I'm still thinking about it because the real steps go by 5.... I think when we make this we have to make it real.... Even more simple.... Set the step in cpp and make the steps there and move the step option from up down... And remove the default... I hope you understand what I mean...

littlesat:
I just want to set a step size as config to that cpp stuff... ;)

littlesat:
FYI: https://github.com/OpenPLi/enigma2/commit/741785b948c96c20697e682dae10cde53874b7eb

It was made wierd so I need make some stuff to keep it backwards compatible....

PR #750, opened 2017-02-22
About: read CPU temperature from sysfs

athoik:
Some machines report cpu temperature using sysfs.

We can read the file /sys/class/thermal/thermal_zone0/temp,
which will output the temperature in millicelsius.


littlesat:
Thanks!

https://github.com/OpenPLi/enigma2/commit/b0f6d92a4fad12a11e1764089853351f4f272415

PR #751, opened 2017-02-22
About: read image install date using /etc/version

athoik:
This solves install date displayed as 1970 in some machines

### DIFF
--- a/lib/python/Components/About.py
+++ b/lib/python/Components/About.py
@@ -20,7 +20,7 @@ def getImageVersionString():
 
 def getFlashDateString():
 	try:
-		return time.strftime(_("%Y-%m-%d %H:%M"), time.localtime(os.stat("/boot").st_ctime))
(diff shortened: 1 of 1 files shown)


littlesat:
Are you sure this presents the install date..... for my ET10K it goes back to Jul 31  2015 which was never the install date...???

littlesat:
I think we better can go for /etc/installed... but then we also have the risk that the time is unknown when it is created :(.... So somehow we need to write a date with the image in E2 as soon we have the first time a date above 1970...

athoik:
Sure, once there is a date after the date that Enigma created it should create file and use the info from there, else use unknown.

littlesat:
Maybe we just need to search for a file that has a timestamp of the build...

littlesat:
I did some checks over several hardware and debugs for an hour and it seems this commit request is pretty reasonable... Sorry for all the delay!!!

PR #752, opened 2017-02-22
[TimerEdit] fixed issue 'disable timer' when state is waiting

Dima73:


Dima73:
more details https://forums.openpli.org/topic/48666-issues-when-deactivating-a-repeat-timer-just-before-the-recording-is-starting/?view=findpost&p=672201

PR #760, opened 2017-02-24
Streamserver: use 30 seconds socket timeout

athoik:
The streamserver locks a tuner for several minutes, if there is
an abnormal connection drop (eg power issue, cable removal etc).

Use TCP_USER_TIMEOUT with timeouf of 30 seconds to allow streamserver
to recover faster, releasing the tuner back to Enigma2.


littlesat:
As Erik is busy with this topic I wait for him to merge this...

athoik:
The newer version by @eriksl  obsoletes that PR. 

https://github.com/OpenPLi/enigma2/commit/4064434047b32ca9c1b3305b4afcbc0b3399b812

So we are ok to close it ;)

PR #767, opened 2017-02-26
Cleanup ChoiceBox.py

persianpros:


persianpros:
Is it ok to import whole enigma ? I mean what you have in line 8

PR #772, opened 2017-02-26
InfoBar.py: Remove useless import

persianpros:
Also cleanup


persianpros:
Is it ok to import whole enigma ? I mean what you have in line 12

PR #777, opened 2017-02-26
Add new option 'use dvbapi5 statistics' for show snr/db in gui

Dima73:
For DVB API 5 need same recalculation.
Samy simple option if DVB API 5 incorrect values, return to old DVB API.
In the DVB API made great recalculation value.

et8500/et7x00/sundtek/osmega --> incorrect values for DVB API 5


Taapat:
This is used only if dvbapi version is 5.10 or higher.
I think that everything that is related with this, it is necessary to include in check of dvbapi version.
Otherwise, we get a non working config and wraning about unused variable on receivers that use less dvbapi version.

Dima73:
How to do it?
Please help.

Taapat:
In c++ code add:
```
#if DVB_API_VERSION > 5 || DVB_API_VERSION == 5 && DVB_API_VERSION_MINOR >= 10
...
#endif
```
How to do it setup.xml in I do not know.

But the next question is how many receivers used dvbapi version less than 5.10?
I suspect that we can remove this checks in enigma2.

littlesat:
Is this only needed for et8500/et7x00/sundtek/osmega....
Basically they should fix it... 
But when it is box specific why do we need a config in the UI?

Dima73:
Manufacturers can not fix it.
Let's just go back to the old api.

PR #803, opened 2017-02-27
Add new option 'use dvbapi5 statistics' for show snr/db in gui

Dima73:
dvbapi5 statistics  show incorrect values from sundtek/osmega/et7000/et8500
return old api


littlesat:
Why do we need a config for this?

Dima73:
Manufacturers use false values dvbapi5 statistics snr/db.

littlesat:
They Should fix it... And to make a config of it is bad. The UI mist not be involved with this. Then make a feature of it instead.........

Dima73:
littlesat
I do not understand what should I do?

littlesat:
Basically wakt until xtrend solve the issue... Or add it as feature not a config....

Dima73:
What means "add it as feature not a config"?

littlesat:
Yep now it is a config... I think that is 'not done' when it is for specific boxes... then it should be via a SystemInfo.py (key) -or- a machine feature.

PR #809, opened 2017-03-01
Remove results files

persianpros:


littlesat:
Are you sure this can be removed?

persianpros:
Yes it's not like main folder ;)

PR #812, opened 2017-03-01
[frontend] add variable use_dvbapi5_statistics

Dima73:
-exclusion list for tuners with wrong snr/db value dvb api 5

### DIFF
--- a/lib/dvb/frontend.cpp
+++ b/lib/dvb/frontend.cpp
@@ -676,7 +676,7 @@ int eDVBFrontend::openFrontend()
 	if (!m_multitype)
 		m_type = feSatellite;
 
-	use_dvbapi5_statistics = (!strstr(m_description, "FTM-4862 (Availink AVL6862)") && !strstr(m_description, "Sundtek"));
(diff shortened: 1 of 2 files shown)


littlesat:
I do not understand this line..

if (use_dvbapi5_statistics && m_dvbversion >= DVB_VERSION(5, 10))

Shouldn't it be

if (use_dvbapi5_statistics || m_dvbversion >= DVB_VERSION(5, 10))

Or could we simply consider to add the tuners in that specific line...

if (strstr(m_description, "FTM-4862 (Availink AVL6862)") || strstr(m_description, "Sundtek" || m_dvbversion >= DVB_VERSION(5, 10))

??

Dima73:
use_dvbapi5_statistics assigned once when openFrontend

littlesat:
Do we need this bool...

use_dvbapi5_statistics

When we can put all in the if..... ?  (with some comments).

The declaration of this bool is not required at all......

littlesat:
I'm afraid you still don't understand... As far I can see the bool is not required at all... Everything can be put in the if by changing just one line (as far I can see)...

littlesat:
FYI: https://github.com/OpenPLi/enigma2/commit/be101d226a74a2bd6b7c99d6d7a17dfac1b458d6

PR #822, opened 2017-03-04
Refactor dvb text encoding handler

stevenhoving:
Note, this pr also includes the code from [pr 821](https://github.com/OpenPLi/enigma2/pull/821) simply because it depends on it.

Now this is a bit different, because I am working on a bigger overhaul of the character encoding system (see https://github.com/stevenhoving/dvb_utf8) I have spend this afternoon figuring out how to incorporate my newly developed code in the enigma code base. This also means that I have figure out how the dvb character encoding quirk overwrite system works.
Although looking at the code didn't really sparked any ideas, I did find a couple of things in the code that could be improved. Like moving the `defaultEncodingTable` global variable inside the `eDVBTextEncodingHandler` class. The way I see it, they belong to each other.

**But I do have 2 additional questions:**
- Why does this file have its own `inline char tolower(char c)` implementation? Why don't you just use the implementation provided by `#include <ctype.h>`?
- You guys are using GCC6 right? Last time I checked it changed the default C++ version to 14, so unless you guys changed it back to `-std=gnu++98`, I could replace those ugly iterator definitions with `auto`. Wait, before you start to `#resist` about `auto`... let me explain. I do not mean AAA (Almost Always Auto), I would only propose to allow `auto` to be used in those ugly iterator situations.
For example, code like this:
```C++
for (std::map<eServiceReferenceDVB, ePtr<eDVBService> >::iterator sit(m_services.begin());
		sit != m_services.end(); ++sit)
	{
		if (sit->first.getTransportStreamID() == Tsid &&
			sit->first.getOriginalNetworkID() == Onid &&
			sit->first.getServiceID() == Sid)
			return sit->first;
	}
```
could become
```C++
for (auto &sit : m_services)
{
    auto &ref = sit.first; // service reference
    if (   ref.getTransportStreamID() == Tsid
        && ref.getOriginalNetworkID() == Onid
        && ref.getServiceID() == Sid)
            return ref;
}
```

Which in the end, is exactly the same as the code above but a bit easier to read (I know the code is not from this file, but I choose it only to illustrate the advantage).

### DIFF
--- a/lib/base/encoding.cpp
+++ b/lib/base/encoding.cpp
@@ -48,8 +48,8 @@ eDVBTextEncodingHandler::eDVBTextEncodingHandler()
 		/* no personalized encoding.conf, fallback to the system default */
 		file = eEnv::resolve("${datadir}/enigma2/encoding.conf");
 	}
-	CFile f(file.c_str(), "rt");
 
+	CFile f(file, "rt");
 	if (f)
 	{
 		size_t bufsize = 256;
--- a/lib/base/encoding.cpp
+++ b/lib/base/encoding.cpp
@@ -1,26 +1,26 @@
 #include <cstdio>
 #include <cstdlib>
+
 #include <lib/base/cfile.h>
 #include <lib/base/encoding.h>
 #include <lib/base/eerror.h>
 #include <lib/base/eenv.h>
 
-eDVBTextEncodingHandler encodingHandler;  // the one and only instance
-int defaultEncodingTable = 1;   // the one and only instance
+static eDVBTextEncodingHandler encodingHandler;  // the one and only instance
 
 inline char tolower(char c)
 {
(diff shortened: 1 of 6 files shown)


littlesat:
It seems this one has conflicts....

stevenhoving:
I will rebase... In a moment...

littlesat:
Imhad to revert the other merge request from you as they did not compile.... Now you've to wait for next weekend because I cannot test/evaluate

stevenhoving:
@littlesat not a problem, in the mean time I am trying to get enigma2 to build with travis-ci. So it can build and test my merge requests. See https://travis-ci.org/stevenhoving/enigma2
Its just WIP for now.

stevenhoving:
@littlesat Don't merge this pr just yet, let me finish my 'continuous integration' experiment.

littlesat:
Is it then ok I close this merge request and wait for te final?

PR #824, opened 2017-03-05
improve ci/vtuner support on fbc boxes

Dima73:
thanks O-E guys


littlesat:
What exactly does it improve?.... I do not see the problem which this merge request should improve....

pieterg:
vtuner ioctl's and fbc are unrelated.

littlesat:
Jjji merged it because it is a fix... But it needs to be revieuwed as it can be improved....

pieterg:
the commit needs to be split, as it addresses two unrelated things.
the fbc / ci stuff can be applied.
Regarding the vtuner ioctl's, we do not use the conflicting VTUNER_SET_RESPONSE, so I don't think we need the alternative numbering.
Even if vu dropped the backward compatibility for the original numbering, this is not related to fbc in any way.

PR #838, opened 2017-03-11
dvbci: restore checking CI slots before writing to tsmux

athoik:
If there is more that 1 CI nodes we use the DM8000 way (CI0 / CI1 ...)
If there is 1 CI node or none we use the DM7025 way (CI and writing /proc/stb/tsmux/input2).

Also initialize 2 inputs (dm7025) or 4 inputs (dm8000) like before.
Although the proper solution would be to initialize all inputs (with FBC we have way more than 4 inputs) by checking the existance of /proc/stb/tsmux/inputX nodes.
Or just initialize them, code will just produce a warning in setInputSource, not visible by default debug level.


eriksl:
Applied.

PR #839, opened 2017-03-11
Create configure option to define crash log email

athoik:
Use in bitbake with the following way:

```
def get_crashaddr(d):
    if d.getVar('CRASHADDR', True):
        return '--with-crashlogemail="${CRASHADDR}"'
    else:
        return 'the OpenPLi forum'

EXTRA_OECONF += "${@get_crashaddr(d)}"
```

More info: https://github.com/OpenPLi/openpli-oe-core/commit/ab71836560d67bc6ee3025b1b364b89508dd960d


eriksl:
Applied, thx!

PR #846, opened 2017-03-15
[RemoteControlType] add some model

Dima73:
-vs1000/vs1500/hd51/hd11/hd500c/hd530c/formuler4turbo

Warning: vs1500(hd51?) really not work for current drivers.
Tested remote VU+solo4k/Formuler1.


Dima73:
1)According to my tests multi rc noe work for vs1500.
2)This is a simple sort by numbers. So it's more convenient to read.

PR #847, opened 2017-03-15
dvbci: Initialize only two first inputs for legacy compatibility

athoik:
It seems that machines with no CI or one CI still require to initilize up to two inputs.
Bring back legacy compatibility and investigate further.


littlesat:
Sorry I just make my own commit a few seconds ago... and also sorry that in that commit I forgot to mention you.... 1000 times xqq

PR #850, opened 2017-03-16
added queue for delay between sended keyEvents

ims21:


littlesat:
Please use the return option I discussed with you via pm... Now when the box is in standby it might not quitmainloop....

PR #860, opened 2017-03-20
Language: Fallback to en_EN when selected language does not exist

athoik:
More info: https://forums.openpli.org/topic/49334-todays-update-vu-solo2/page-4#entry687866


littlesat:
Thanks!

PR #861, opened 2017-03-20
add symbol_recording hd51/vs1500 for SystemInfo["LCDsymbol_circle_rec…

Dima73:
…ording"]

-Previously, it was controlled by the driver.
Therefore, streaming was considered as a record.
We pass control to the image.


littlesat:
Do we really need the VS1500/HD51 check?

Dima73:
Yes, this need only VS1500/HD51.
Same symbol is F1/3 --> not work and etx900 --> work fine as controlled by  driver

PR #875, opened 2017-04-06
[frontend] correct SNR value for AVL6211(hd51/vs1500)

Dima73:
-we will remove it when the drivers are fixed

### DIFF
--- a/lib/dvb/frontend.cpp
+++ b/lib/dvb/frontend.cpp
@@ -1243,7 +1243,7 @@ int eDVBFrontend::readFrontendData(int type)
 							{
 								return signalqualitydb;
 							}
-							if(!signalquality)
+							if(!signalquality || !strstr(m_description, "AVL6211"))
 							{
 								/* provide an estimated percentage when drivers lack this info */
 								calculateSignalPercentage(signalqualitydb, signalquality);


littlesat:
We should wait for the drivers to be fixed is my suggestion... When we merge this then there is no motivation for the manufacturer to fix it as we already worked-around it...

Dima73:
ok,wait.
But how much should we wait?

littlesat:
When the manufacturer does not fix it forever... 
Did OE-A/OpenATV worked around that?

Dima73:
I did not see(OE-A/OpenATV).

HD-Digital:
The AVL6211 gives now the "real" SNR value. This means it's not manipulate anymore that value is always between 80 till 100%. Can you please explain why this new value should been wrong.

Dima73:
example:
vs1500
dB 13.3-13.5
SNR 68%

Really(Formuler1/old driver et8500/dual tuner Vuplus DVB-S NIM(AVL6222) solo4k)
dB 13.2
SNR 88%

Dima73:
example:
vs1500
dB 12.0
SNR 60%

Really
dB 10.8
SNR 67%

Issue #422, opened 2016-09-06
Update libsigc++ to 2.0


athoik:
- [x] Update to libsigc++ to 2.0 using `#defines`
- [x] Change Signal0..5 to sigc::signal0..5
- [x] Change Slot0..5 to sigc::slot0..5
- [x] Change Connection to sigc::connection
- [x] Change Object to sigc::trackable
- [x] Wait for OpenPLi 5 release
    - [ ] Update serviceapp to newer libsigc++
- [ ] Remove libsigc++ 1.2 from meta-openpli and use meta-openembedded sigc++ 2.0

libsigc++ work in progress: https://github.com/athoik/enigma2/tree/sigc
serviceapp work in progress: https://github.com/athoik/serviceapp/tree/sigc

PR #882, opened 2017-04-10
Update libsigc++ to 2.0

athoik:
Porting was done mostly using instructions found here: https://mail.gnome.org/archives/libsigc-list/2005-July/msg00010.html

The Object typedef was changed to sigc::trackable to identify that sigc++ is used, instead of generic "Object".
The Connection typedef was changed to sigc::connection as well.

Most defines in libsig_comp.h deleted, only CONNECT and CONNECT_EXTRA left.

Closes: #422


athoik:
Update to libsigc++ 2.0 requires also the OE update https://github.com/OpenPLi/openpli-oe-core/pull/200

PR #886, opened 2017-04-14
[SleepTimerEdit] add option "Sleeptimer action" (standby/deep standby)

Dima73:


littlesat:
This commit request include two other commits.... and this is not the preferable method...

PR #910, opened 2017-05-07
Components folder python cleanup: Remove useless imports

persianpros:


WanWizard:
@littlesat can you verify?

PR #911, opened 2017-05-07
ScanSetup for terrestrial use terestrial provider setting instead ind…

ims21:
…ependent region

- solve wrong displaying chanel numbers for scaning


littlesat:
Why not use the property value!

ims21:
ok, merge it and then i will create patche, where will be replaced self.TerrestrialRegion.getText() with  self.TerrestrialRegion.description[self.TerrestrialRegion.value]

PR #913, opened 2017-05-08
enigma2.sh improvements

persianpros:
Thanks to "arn354", "atvcaptain", "andyblac".


WanWizard:
I'd like to know the why for all three, before merging. If needed, they all seem to be done in the wrong place. We don't do workaround like ATV, we like to fix things permanently...

littlesat:
I also do not see the real need....???

persianpros:
Well in my opinion "OOM score" is a nice thing to have but the rest are just workarounds for some situations.

littlesat:
Why is that OOM score a nice thing to have?

PR #914, opened 2017-05-08
Add c++11 check for older g++ versions

persianpros:
Thanks to skl


littlesat:
Please clarify what exextly this is doing/improving/having benefits......?

persianpros:
https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx_11.html

In some cases developers get errors because of the absence of these files.

littlesat:
But do we get these issues?

PR #915, opened 2017-05-08
Fix typo, kB is wrong and just like MB, GB we should use KB

persianpros:


ims21:
I do not know, if you know Metric prefix SI, but kilo is always "k" and not "K"

persianpros:
One "kilobyte" (KB) is 1024 bytes in JEDEC-standard, whereas the definition has shifted, in most contexts, to mean 1000 bytes (kB) in accordance with SI. To resolve this difference, binary prefixes (kiB) are used.

Now take a look at https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Components/Converter/MovieInfo.py#L57 we're dividing it to 1024 so it's KB ;)

ims21:
omg ...

persianpros:
Is it wrong ?

ims21:
all Win / Linux filemanagers using kB ... It's optically highlighting that the file is small ...

persianpros:
But in windows I have KB :|

littlesat:
I have no idea if I should merge this request.... because both are fine in my perspective....

Pr0metheus2:
I prefer also kB...

persianpros:
Still it's the right thing to do because it's 1024 not 1000

ims21:
All know, that kilo is in binary 2 to 10 and not 10 to 3 ... 
I prefer kB, not KB...  As I saw on filesystems (Linux, W10), both using B, kB, MB

persianpros:
But in windows 7 and server 2008 r2 I have KB not kB, If you need i can provide you some screenshots ;)

ims21:
ubuntu, server 2012, W7, W10 ... kB  ...  It seems, it will be nationaly dependent .

... make it translatable (use KB) and who will want, - will change it in translation - in .po files

ims21:
but then create actual .pot and po files for enigma2

littlesat:
When it is indeed language depended we should indeed make it translatable

ims21:
I think, it is ... because my all programs and operation systems using kB ... his KB. How is it in NL ?
If it will be translatable, then translator can choose, how he will want :-)

Pr0metheus2:
I have also KB in Win7.

blzr:
checked here on _some_ PC's - on all win (xp,7,8,10) running machines there is KB, regardless PL or EN locale; on ubuntu 16.04 I have KiB; kB nowhere to be found...

Pr0metheus2:
Then, maybe we should do what is right and not what we like... :)

persianpros:
As I said before the true form is KB when we have 1024 and KiB when we have 1000 so it's better to merge this request ;)

blzr:
actually KiB stands for kibybyte = 1024 bytes (same as a KB - 'kilo'byte),
kB is 1000 bytes...

persianpros:
Should I close this ?

ims21:
i think, littlesat will merge it.

littlesat:
When a language needs to have it in capital letters it can be arranged by translation. For now I think we should leave it as-is.... Please re-open when there are new good arguments to change it....

persianpros:
But it's the right thing to do, Check https://github.com/OpenPLi/enigma2/pull/915#issuecomment-300740600 there is no kB :|

blzr:
@littlesat, 
apparently you did not read above comments at all (?)
change has nothing to do with language, kB = 1000 bytes, KB (or KiB) = 1024 bytes, so for now we use wrong unit, simple as that...

eriksl:
Seconded.

And KB is wrong anyway. "k" as in unit prefix is always written lowercase (as in "km", not "Km"), which is a bit illogical, but that's the way it is.

eriksl:
Some of the other changes may be useful though.

PR #933, opened 2017-05-15
Dish.py: Cleanup skin + Skip unwanted timeout

persianpros:


littlesat:
There is more in this merge request then meantioned in the description and therefore Imcannot accept this request. E.g. Ot also removes the default skin stuff...

PR #941, opened 2017-05-17
VirtualKeyBoard.py: Fix possible BSOD + Some cleanups

persianpros:
And please check https://github.com/OpenPLi/enigma2/pull/915 again.


ims21:
cleaning in InfoBar is nonsence ... very unclear if is all in one line. but it is only my opinion

littlesat:
I afreed with ims and Plessers offer 4 different merge requests....

persianpros:
It's about not using \ in imports nothing more (InfoBar.py)

littlesat:
It's choosing between extreme long lines -or- having some "\".... a question of "taste"... so I think this is not mandatory to change.

Please offer the other 3 in a separate merge request.

persianpros:
What about https://github.com/OpenPLi/enigma2/pull/915 ?

PR #950, opened 2017-05-21
Rename receiver to STB as PLi's enigma2 is a framebuffer based fronte…

persianpros:
…nd for DVB functions on Linux settop boxes and a STB is not just a receiver.

Please update all po files after this merge just like https://github.com/OpenPLi/enigma2/commit/ab2ddec4a801001d5702bfa6508ca287ff76fd0c


betacentauri:
Is it really a good idea?
At least in Germany I would still translate it to receiver as set top box has no good German translation.

Taykun345:
Better to stick with receiver. We in Slovenia translate this to 
"sprejemnik", for STB we also do not have good translation.

persianpros:
Well in Persian is quite the opposite but it should not rely on translations as you can still use "receiver" if you want but "receiver" is a very simple word and STB is better.

ims21:
Here is "receiver" well... STB has here not equivalent - only "STB" again. 
And in addition, it was here infamous secret police.

If you have problem with this, translate is in persian .po.

persianpros:
As I said you can use receiver still but it should not rely on translations, It's a word but receiver is just very simple :)

littlesat:
I suggest we keep it receiver

PR #951, opened 2017-05-21
ServiceInfo.py: Add "IS_STEREO"

persianpros:


littlesat:
Why is this required... please explain more?

eriksl:
I suggest showing the amount of channels instead, 1 .. 8.

persianpros:
@eriksl That's a good idea so please do it :)

littlesat:
i still do not get the real added value? Where is it used...do you have some examples?

littlesat:
Why not.... It does not harm....

PR #953, opened 2017-05-21
eServerSocket: Nicer solution to "fix remote client's address being g…

Huevos:
…arbage"

Use of AF_* values rather than PF_* ones (they happen to have the same values, but AF_* ones are the syntactically correct ones).

Use of the IN6_IS_ADDR_V4MAPPED macro, rather than hard-wiring mystical numbers with no explanation.

Remove an unnecessary assignment on a declaration (as it is immediately followed by an assignment anyway).

Related commit: https://github.com/OpenPLi/enigma2/commit/41211499b4ccf036859e45654bca428dbb882e9b

Credit: Birdman

### DIFF
--- a/lib/network/serversocket.cpp
+++ b/lib/network/serversocket.cpp
@@ -35,30 +35,27 @@ void eServerSocket::notifier(int)
 
 	switch(client_addr.sock.sa_family)
 	{
-		case(PF_LOCAL):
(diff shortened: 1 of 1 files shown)


eriksl:
Actually PF_* is the preferred syntaxis and AF_* is legacy compatibility.

PR #955, opened 2017-05-23
Nim manager patches

Huevos:
Sort out various issues causing "nimListCompressed" to output rubbish.

### DIFF
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -540,7 +540,7 @@ class NIM(object):
 		return connectable[self.getType()]
 
 	def getSlotID(self, slot=None):
-		return chr(ord('A') + (slot or self.slot))
+		return chr(ord('A') + (slot if slot is not None else self.slot))
 
 	def getSlotName(self):
 		# get a friendly description for a slot name.
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -609,8 +609,8 @@ class NIM(object):
 		return "%s: %s" % (self.slot_name, self.getFullDescription())
 
 	def getFriendlyFullDescriptionCompressed(self):
-		if self.isFBCRoot():
-			return "%s-%s: %s" % (self.slot_name, self.getSlotID(self.slot + 7), self.getFullDescription())
+		if self.isFBCTuner():
+			return "%s %s-%s: %s" % (_("Tuner"), self.getSlotID(self.slot - (self.slot % 8)), self.getSlotID((self.slot - (self.slot % 8)) + 7), self.getFullDescription())
 		#compress by combining dual tuners by checking if the next tuner has a rf switch
 		elif os.access("/proc/stb/frontend/%d/rf_switch" % (self.frontend_id + 1), os.F_OK):
 			return "%s-%s: %s" % (self.slot_name, self.getSlotID(self.slot + 1), self.getFullDescription())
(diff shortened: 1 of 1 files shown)


littlesat:
It is not lazy coding that self or self.slot.... and we want only to show the root FBC tuner.... Displayed as 
First Tuner -dash- Last tuner - Space Description....
I'm afraid you do not understand something....

Huevos:
No it doesn't do what it is supposed to do.
If you run that on Solo4K the current code prints:

Tuner: A-H Vu-Tuner blablabla (DVB_S2)
Tuner: B-I Vu-Tuner blablabla (DVB_S2)
Tuner: I-J VU Tuner blablabla (DVB-S2)

Is that is what OpenPLi intended?

Huevos:
Also, my code can be run on any tuner in the FBC complex and it returns the correct start and end tuner letters.

arn354:
I can confirm this - Tuner A (0) and B (1) are internal root fbc-tuners. both of their possible leaf-tuners are C to H (2 to 7). 
the current code does not list it correctly.

littlesat:
Wierd I cannot confirm it.... Could it be that root tuners are differently defined on oe-a images...?

littlesat:
Hopefully this might help

https://github.com/OpenPLi/enigma2/commit/e8e8fc5a38d85620b7f858298e6f9b55faa82c8f

Huevos:
> The middle patch I cannot follow....

How that works is if "self.slot" is any number between 0 and 7 it will return:
Tuner: A-H Vu-Tuner blablabla (DVB_S2)

If  "self.slot" is any number between 8 and 15 it will return:
Tuner: I-P Vu-Tuner blablabla (DVB_S2)

So as long as "self.slot" is part of an FBC complex it returns the correct parent name for the whole complex.

Huevos:
> Or does SOLO does not have the rf switch on the second FBC root tuner?
No, Solo4K does not have rf_switch on any of the FBC tuners.

But it is internally linkable. i.e. connect sat cable to socket "A" and tuners A-H are all connected, including tuner B.

littlesat:
Sounds like another driver weirdness.... solved....

eriksl:
Littlesat, the internal FBC tuner in the Solo4k is EXACTLY the same as the pluggable FBC tuners.

What you probably didn't take into account is that the Ultimo4k can have TWO FBC pluggable tuners, both of them DVB-S, so you will have 2 x 2 physical inputs ("root" tuners), the first two are connectable to ABCDEFGH the second two inputs are connectable to IJKLMNOP.

eriksl:
The FBC tuners do not have an rf_switch entry, none of them, because they don't use it. The FBC linking code uses another mechanism.

littlesat:
The ultimo4K reports an rf swtich here on both FBC tuners.... the DVB-C in the first slot and teh DVB-S2 in the second slot..... That was why I did not see the "issue".... but it seems to be solved now as far I can see.... Can I close this thread?

Huevos:
If you don't want OpenPLi to have access to the second commit you can close.

For me the second commit is useful.

e.g. if nim.slot = 4 you can call nim.getFriendlyFullDescriptionCompressed()
Output will be: Tuner: A-H Vu-Tuner blablabla (DVB_S2)

With the current code calling nim.getFriendlyFullDescriptionCompressed() on tuners other than first root tuner gives strange results like: Tuner: D-K Vu-Tuner blablabla (DVB_S2)

IMO, you should be able to call this function individually and get a correct result. With the current code that is impossible.

littlesat:
Now I understand....

littlesat:
Why did you add _("Tuner") to that middle patch...? I'm afraid this is not intended....

Huevos:
No. Won't work quite right.

Close this request and I will create another one. 

I fully understand what you mean about "Tuner".

PR #956, opened 2017-05-23
eServerSocket: Nicer solution to...

Huevos:
Nicer solution to "fix remote client's address being garbage"

Use AF_* values rather than PF_* ones (they happen to have the same values, but AF_* ones are the syntactically correct ones).

Use the IN6_IS_ADDR_V4MAPPED macro, rather than hard-wiring mystical numbers with no explanation.

Removing an unnecessary assignment on a declaration (as it is immediately followed by an assignment anyway).

Related commit: https://github.com/OpenPLi/enigma2/commit/41211499b4ccf036859e45654bca428dbb882e9b

Credit: Birdman

### DIFF
--- a/lib/network/serversocket.cpp
+++ b/lib/network/serversocket.cpp
@@ -35,30 +35,27 @@ void eServerSocket::notifier(int)
 
 	switch(client_addr.sock.sa_family)
 	{
-		case(PF_LOCAL):
+		case(AF_LOCAL):
 		{
 			strRemoteHost = "(local)";
 			break;
 		}
 
-		case(PF_INET):
+		case(AF_INET):
 		{
-			strRemoteHost = inet_ntop(PF_INET, &client_addr.sock_in.sin_addr, straddr, sizeof(straddr));
+			strRemoteHost = inet_ntop(AF_INET, &client_addr.sock_in.sin_addr, straddr, sizeof(straddr));
 			break;
 		}
 
-		case(PF_INET6):
+		case(AF_INET6):
 		{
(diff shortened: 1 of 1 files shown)


Huevos:
More info on AF/PF:
https://stackoverflow.com/questions/6729366/what-is-the-difference-between-af-inet-and-pf-inet-in-socket-programming

eriksl:
Yeah more or less. At some time in the past (long ago) it was conceived that PF_* should be the successor of AF_* because AF_* would not appropriately cover the semantics. In the meantime everybody kept using AF_* so they were forced to keep them the same numerical value and now it's guaranteed so (and has been for a long time), because nobody bothered to start using PF_*. Anyway, I can live with both, if you rather see AF_*, be my guest.

eriksl:
_Removing an unnecessary assignment on a declaration (as it is immediately followed by an assignment anyway)._

Actually I can't find anything in your patch that does this. It removes a variable that is no longer required, fair enough. But besides that it only adds redundant braces around a single line.

In fact the whole patch doesn't seem to bring any functional changes, it's all cosmetic. Fine by me, but then call it that way and don't make unnecessary changes to the layout (add/remove braces, if/then style etc.). The use of IN6_IS_ADDR_V4MAPPED is nice, but will probably not generate any other code.

So, agree to change PF_* to AF_* and introduction of IN6_IS_ADDR_V4MAPPED and call it "cosmetic patch".

Huevos:
Ok, so you want me to create another pull request?

Huevos:
Ok, new version here.
https://github.com/OpenPLi/enigma2/pull/957
I hope this is acceptable.

PR #957, opened 2017-05-23
eServerSocket: Cosmetic patch

Huevos:
Use AF_* values rather than PF_* ones

Use IN6_IS_ADDR_V4MAPPED macro

Credit: Birdman


eriksl:
Applied. It wasn't necessary to create a new PR, I would have selectively cherry-picked.

PR #959, opened 2017-05-26
NimManager: Allow getFriendlyFullDescriptionCompressed() to...

Huevos:
Allow getFriendlyFullDescriptionCompressed() to be called on any slot within the FBC complex and return the correct name for the whole complex.

i.e. calling this function when nim.slot = 4 will return: "Tuner A-H: Vu-Tuner blablabla (DVB_S2)".

Also, this commit along with the previous two from Littlesat allow these functions to be called individually with arguments or as part of a chain, giving the coder maximum flexibility for their future use in other applications, plugins,etc, not just the "About" screen.


littlesat:
Thanks...mwas already implemented slightly differently.... Sorry

PR #970, opened 2017-06-09
Add option "Toggle 12 or 8 character display mode" - VFD osmega

Dima73:


littlesat:
Again I thing this could be set by language... So when it is required for a specific language, then enable it.... No additional user setting needed....

littlesat:
Why did you close this..... instead if continue the discussion.... :(

littlesat:
See... https://github.com/openatv/enigma2/commit/f16b7baf3792ed15d78e33e7938574889da3d100

maybe we should go to call the SystemInfo "hasXcoreVFD"... and also include these receivers.... but we should also find a smart way to find the correct proc... Looks like the box name is included

littlesat:
SystemInfo["hasXcoreVFD"] = fileCheck("/sys/module/brcmstb_%s/parameters/pt6302_cgram" % HardwareInfo().get_device_model())

littlesat:
Merged.... but slightly different and also compatible with the 4K devices with the same display.

PR #988, opened 2017-06-18
[HDMI-CEC] plugin - set default enabled as off

Dima73:
-this plugin is not for normal users

### DIFF
--- a/lib/python/Components/HdmiCec.py
+++ b/lib/python/Components/HdmiCec.py
@@ -46,7 +46,7 @@ cmdList = {
 	}
 
 config.hdmicec = ConfigSubsection()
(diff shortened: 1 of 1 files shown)


Taykun345:
I agree. I for example want this off, as I use Logitech remote.

littlesat:
I suggest we should left it ON.... the way it is now is fine....

ims21:
"I agree. I for example want this off, as I use Logitech remote" 

 ... good argument ...   omg

Taykun345:
Yeah, not a good argument :D

WanWizard:
We agreed the default should be "off". ;-)

PR #993, opened 2017-06-20
Label.py: Detect render source errors

persianpros:


littlesat:
if(source):

->

if source:

PR #994, opened 2017-06-20
FrontendStatus.py: Turn negative values to 100%

persianpros:
106% is displayed as -6%
Thanks to arn354


littlesat:
When does this occur... Why do we need this???

PR #995, opened 2017-06-20
Fix to work dvbci on vtuner

athoik:
based on: http://code.vuplus.com/gitweb/?p=vuplus_dvbapp;a=commit;h=50e7a4d8e274532c6b0519d48f91a13757c7eaeb


pieterg:
Why not use sectionfilters instead? They use a lot less resources

athoik:
I am not sure how to use section filters (I think that dvbtime uses section filters).

Although maybe VU+ choose to use PES readers because they are guarantee to work on USB tuners? Or all USB tuners can support section filters all well?

pieterg:
Sure, sectionfilters are implemented by the dvb-core.
And in fact the filters are started on the vtuner dvb adapter, not on the usb dvb adapter

PR #1004, opened 2017-06-26
[CableScan/FastScan] add sanity check if Session is None

Dima73:
-When install plugin without restart GUI

### DIFF
--- a/lib/python/Plugins/SystemPlugins/CableScan/plugin.py
+++ b/lib/python/Plugins/SystemPlugins/CableScan/plugin.py
@@ -248,7 +248,7 @@ def getNimList():
 	return [x for x in nimmanager.getNimListOfType("DVB-C") if config.Nims[x].configMode.value != "nothing"]
 
 def CableScanMain(session, **kwargs):
-		Session.open(CableScanScreen, getNimList())
+	session.open(CableScanScreen, getNimList())
 
 def restartScanAutoStartTimer(reply=False):
 	if reply:
@@ -259,7 +259,7 @@ def restartScanAutoStartTimer(reply=False):
 
 def CableScanAuto():
 	nimlist = getNimList()
-	if nimlist:
+	if nimlist and Session:
 		if Session.nav.RecordTimer.isRecording():
 			restartScanAutoStartTimer()
 		else:
@@ -278,8 +278,9 @@ def standbyCountChanged(value):
 
 def startSession(session, **kwargs):
 	global Session
-	Session = session
-	config.misc.standbyCounter.addNotifier(standbyCountChanged, initial_call=False)
+	if Session is None:
+		Session = session
+		config.misc.standbyCounter.addNotifier(standbyCountChanged, initial_call=False)
 
 def CableScanStart(menuid, **kwargs):
 	if menuid == "scan" and getNimList():
--- a/lib/python/Plugins/SystemPlugins/FastScan/plugin.py
+++ b/lib/python/Plugins/SystemPlugins/FastScan/plugin.py
@@ -367,8 +367,9 @@ def standbyCountChanged(value):
 
 def startSession(session, **kwargs):
 	global Session
-	Session = session
-	config.misc.standbyCounter.addNotifier(standbyCountChanged, initial_call=False)
+	if Session is None:
+		Session = session
+		config.misc.standbyCounter.addNotifier(standbyCountChanged, initial_call=False)
 
 def FastScanStart(menuid, **kwargs):
 	if menuid == "scan":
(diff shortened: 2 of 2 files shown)


littlesat:
As far I can see not required....

Dima73:
Why such small cosmetic corrections cause such big problems?
I showed the crash-log, is it not enough?
The second day we spend on this precious time.
And there is no result so far.

littlesat:
Nope... As I could not create the crashlog...... And do not find a mechanism to get it.... Simply when I cannot verify/trigger the issue I do not want to make the code more complicated as needed..... And why do you need restart????

littlesat:
It seems this is required when you want to install/remove the plugin itself...We should try to use AUTOSTART ootion instead as with this we might also solve a gsod when put the box in standby after removing the plugin.

littlesat:
I was thinking about something like this... [untested]
--- a/lib/python/Plugins/SystemPlugins/CableScan/plugin.py
+++ b/lib/python/Plugins/SystemPlugins/CableScan/plugin.py
@@ -248,7 +248,7 @@ def getNimList():
        return [x for x in nimmanager.getNimListOfType("DVB-C") if config.Nims[x].configMode.value != "nothing"]

 def CableScanMain(session, **kwargs):
-               Session.open(CableScanScreen, getNimList())
+               session.open(CableScanScreen, getNimList())

 def restartScanAutoStartTimer(reply=False):
        if reply:
@@ -276,10 +276,14 @@ def standbyCountChanged(value):
                inStandby.onClose.append(leaveStandby)
                CableScanAutoStartTimer.startLongTimer(150)

-def startSession(session, **kwargs):
+def autostart(reason, **kwargs):
        global Session
-       Session = session
-       config.misc.standbyCounter.addNotifier(standbyCountChanged, initial_call=False)
+       if reason == 0 and "session" in kwargs and not Session:
+               Session = kwargs["session"]
+               config.misc.standbyCounter.addNotifier(standbyCountChanged, initial_call=False)
+       elif reason == 1:
+               Session = None
+               config.misc.standbyCounter.removeNotifier(standbyCountChanged)

 def CableScanStart(menuid, **kwargs):
        if menuid == "scan" and getNimList():
@@ -288,8 +292,8 @@ def CableScanStart(menuid, **kwargs):
                return []

 def Plugins(**kwargs):
-       if nimmanager.hasNimType("DVB-C"):
+       return nimmanager.hasNimType("DVB-C"):
                return [PluginDescriptor(name=_("Cable Scan"), description="Scan cable provider channels", where = PluginDescriptor.WHERE_MENU, fnc=CableScanStart),
-                       PluginDescriptor(where=[PluginDescriptor.WHERE_SESSIONSTART], fnc=startSession)]
+                       PluginDescriptor(where=[PluginDescriptor.WHERE_SESSIONSTART, PluginDescriptor.WHERE_AUTOSTART], fnc=autostart)]
        else:
                return []
diff --git a/lib/python/Plugins/SystemPlugins/FastScan/plugin.py b/lib/python/Plugins/SystemPlugins/FastScan/plugin.py
index b8a8cc533..27687da8e 100644
--- a/lib/python/Plugins/SystemPlugins/FastScan/plugin.py
+++ b/lib/python/Plugins/SystemPlugins/FastScan/plugin.py
@@ -365,10 +365,14 @@ def standbyCountChanged(value):
                inStandby.onClose.append(leaveStandby)
                FastScanAutoStartTimer.startLongTimer(90)

-def startSession(session, **kwargs):
+def autostart(reason, **kwargs):
        global Session
-       Session = session
-       config.misc.standbyCounter.addNotifier(standbyCountChanged, initial_call=False)
+       if reason == 0 and "session" in kwargs and not Session:
+               Session = kwargs["session"]
+               config.misc.standbyCounter.addNotifier(standbyCountChanged, initial_call=False)
+       elif reason == 1:
+               Session = None
+               config.misc.standbyCounter.removeNotifier(standbyCountChanged)

 def FastScanStart(menuid, **kwargs):
        if menuid == "scan":
@@ -379,6 +383,6 @@ def FastScanStart(menuid, **kwargs):
 def Plugins(**kwargs):
        if (nimmanager.hasNimType("DVB-S")):
                return [PluginDescriptor(name=_("Fast Scan"), description="Scan Dutch/Belgian sat provider", where = PluginDescriptor.WHERE_MENU, fnc=FastScanStart),
-                       PluginDescriptor(where=[PluginDescriptor.WHERE_SESSIONSTART], fnc=startSession)]
+                       PluginDescriptor(where=[PluginDescriptor.WHERE_SESSIONSTART, PluginDescriptor.WHERE_AUTOSTART], fnc=autostart)]
        else:
                return []

WanWizard:
You probably need to replace your local copy, as the repo branches have been rebuild. See the message on the forum.

Dima73:
I'll make a new patch.
It's just a mess.

PR #1006, opened 2017-06-27
FastScan: fix install/uninstall plugin without restart GUI

Dima73:


littlesat:
I just pushed something around the same time... Please verify if it is sufficiant.

PR #1008, opened 2017-06-28
[SoftcamSetup] add option enable/disable autostart

Dima73:


littlesat:
Please clarify why you should need this?

Dima73:
Over the past years, I have hundreds of times explained to users how to turn on the autostart emulator.

Now it is implemented in the plugin.

littlesat:
But when you install a softcam it will be automatically started... So no need to have it in the code... Sorry I do not see why we need to add such manu lines for this feature where I do not understand the real reason for...

And b.t.w.... This is not done os.system("....")

PR #1017, opened 2017-07-04
Network.py: Some changes (Broadcom and ping)

persianpros:


WanWizard:
why is tap0 (and tun0 for that matter) blacklisted?

persianpros:
Blacklist changes are for STB's with broadcom wlan chips like ultimo4k and it's useful for them, If you think it's not necessary I can remove them.

WanWizard:
tun0 wasn't in this PR, but was already blacklisted. Blacklisting them means you can't check the state of your VPN tunnels via the GUI anymore.

PR #1018, opened 2017-07-04
Some cleanups

persianpros:


lib/python/Components/Timezones.py
```diff
@@ -1,6 +1,6 @@
 import xml.etree.cElementTree
 
-import os
+from os import environ, unlink, symlink, path
```
athoik: What was wrong with import os? Please check better the commits from @milosoftware that talk about namespace pollution (or something like that).

PR #1021, opened 2017-07-04
Network.py: Avoid parsing errors (IPv6 related)

persianpros:
The output format of the full iproute2 "ip" differs from busybox "ip", making iNetwork fail here when iproute2 gets installed.
Avoid this by explicitely invoking "busybox ip".
Thanks to https://github.com/Schimmelreiter


littlesat:
So we do not merge this.... (this patch includes also a console thing when not install... so there is more included than the description indicates....

littlesat:
Mmmm it is just hiding errors regarding IPv6..... so in fact a not-done patch....

PR #1022, opened 2017-07-04
Network.py: Change ping order

persianpros:
google.com is accessible all over the world but .nl not, Also google is much faster for the tests.

### DIFF
--- a/lib/python/Components/Network.py
+++ b/lib/python/Components/Network.py
@@ -419,7 +419,7 @@ class Network:
 	def checkNetworkState(self,statecallback):
 		self.NetworkState = 0
 		self.pingConsole = Console()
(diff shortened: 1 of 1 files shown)


littlesat:
Not really required....

WanWizard:
Just a question from a Python n00b: does that for loop try all three, always, or does it quit after the first correct response?

If it always tests all three, this PR is indeed not required, if it bails out after success, it will definitely be faster in some parts of the world.

persianpros:
I did test this myself and it was faster than before.
I don't know why littlesat closed this simple request :(

WanWizard:
I got that from the intial PR comment, I was just wondering WHY it was faster, because looking at the code, as I undestand it, it checks all three anyway...

PR #1027, opened 2017-07-05
Language.py: Fix possible GSOD

persianpros:


littlesat:
As we cannot find how to trigger this possible GSOD... I did not merge this one....

PR #1037, opened 2017-07-11
Plugin SoftwareManager: fix packet parsing in PacketManager

Taapat:
opkg list now print description in several rows,
therefore, no longer working division by ' - '.
If no ' - ' in the text, assume that this is the description
and therefore add this text to the last packet description.

Before fix:
screenshot

After fix:
screenshot_1


littlesat:
I think this one should not be in rc.... We also consider to remove the software manager.... When merged then for develop.

Taapat:
I know that you want to remove it.
But this can not be the reason that offer a broken plugin with bug in image at this moment.

WanWizard:
I agree it's a bug fix.

PR #1041, opened 2017-07-15
[nimListCompressed] not visible only FBC link tuners

Dima73:
Discussion here
https://forums.openpli.org/topic/16424-some-questions-bugs/?view=findpost&p=739111

### DIFF
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -833,7 +833,7 @@ class NimManager:
 		return [slot.friendly_full_description for slot in self.nim_slots]
 
 	def nimListCompressed(self):
-		return [slot.friendly_full_description_compressed for slot in self.nim_slots if not(slot.isNotFirstFBCTuner() or slot.internally_connectable >= 0)]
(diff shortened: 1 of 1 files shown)


littlesat:
Please do not merge. We first need to investigate Why this one box makes the mistake with that dvb-t tuner. We should not revert something when it is driver related... or it can be fixed

littlesat:
Looks like a driverbug... or at least the internally connectable chec should be solved

littlesat:
In addition when we solve it like this because one box made a mistake we have to remove a few other lines...

Dima73:
Why can not we see the tuner internally connectable?
Who and why did this decide for everyone?

littlesat:
A driver issue Actually did... Please verify first Why it indicated conectable

littlesat:
Why closing it and do not try to get it right.... :(

Can you give me the following info of that box...
"/proc/bus/nim_sockets"
and
ls -R /proc/stb/frontend

Just to verify/investigate what is going-on on that box/tuner.... probably we could find a real solution....

PR #1042, opened 2017-07-15
Movieplayer - fix calls via number keys

ims21:


littlesat:
What about this: 
https://github.com/littlesat/enigma2/commit/4f21478a42d272cb883405c14adf92e1f860c549

PR #1050, opened 2017-07-19
[NimManager] Use mast name in settings, not xml list index.

Huevos:
If xml list index is used it gets messed up every time a new mast gets added further up the .xml file. Which means people unknowingly are trying to scan the wrong mast.

### DIFF
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -1387,7 +1387,7 @@ def InitNimManager(nimmgr, update_slots = []):
 		try:
 			nim.cable
 		except:
-			list = [(str(n), x[0]) for n, x in enumerate(nimmgr.cablesList)]
+			list = [(x[0], x[0]) for x in nimmgr.cablesList]
 			nim.cable = ConfigSubsection()
 			nim.cable.scan_networkid = ConfigInteger(default = 0, limits = (0, 99999))
 			possible_scan_types = [("bands", _("Frequency bands")), ("steps", _("Frequency steps"))]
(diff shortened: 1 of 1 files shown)


Huevos:
The other benefit of this commit is the settings file can be understood by a human being.

littlesat:
Can you Please further clarify with some examples as I really do not understand what you are trying to solve here...

littlesat:
What so you mean with mast?

Huevos:
Mast = transmitter.

Terrestrial.xml contains more than 100 transmitters. If new transmitters get added everything below what is added changes index. So now if my transmitter was at position 50 in the xml file it is now at position 52 so config selects wrong transmitter.

With this code the transmitter name is stored in the settings file, not a meaningless xml list index.

We have used this code in OpenViX in the public image for 4 months without any user complaints, so it is well tested.

AbuBaniaz:
At the moment, they are stored by position number within the xml file. Settings file contains `config.Nims.0.cable.scan_provider=72`. If the order changes or another provider is added, the position number obviously changes and users are unwittingly using a different provider.

This is what the pull request will achieve.
```
config.Nims.8.cable.scan_provider=Virgin Media UK (ex CW)
config.Nims.9.cable.scan_provider=Virgin Media UK (ex CW)
config.Nims.10.terrestrial=Crystal Palace, London (Europe DVB-T/T2)
```

littlesat:
So it is stored by name instead of number?

Huevos:
Yes

PR #1057, opened 2017-07-26
Tuner Config: DVB_C/T, add two tier country/region provider selection.

Huevos:
Select your country and then your local transmitter.

Output is logically sorted making the order in terrestrial.xml and cables.xml irrelevant.


Huevos:
cable

terrestrial

PR #1058, opened 2017-07-27
Add option for timer zap - "Use as PiP if possible"

Dima73:
thanks ims


ims21:
Dimitrij - I still thinking, that PiP timer should be:
 - no repeated 
 - not wakeup box from standby/deepstandy ... for this is ZAP timer 
- if is in PiP something, then replace it with PiP timer service

Create it with only: 
Name
Description
Typ (Pipzap)
Date
Start time
Conflict detection and channel

PR #1061, opened 2017-07-31
[RecordTimer] Add new option 'Event Name First" for recording filenames.

Huevos:
Adds a fourth type of filename composition. standard, long, short, then "Event Name First"
so recorded filenames do not have the date as the filename prefix.

Credit: AndyBlac
https://github.com/OpenViX/enigma2/commit/5ec78dfb9dec22dbdca8734662f7623c1037aa80

### DIFF
--- a/RecordTimer.py
+++ b/RecordTimer.py
@@ -215,7 +215,9 @@ class RecordTimerEntry(timer.TimerEntry, object):
 		name = name or self.name
 		filename = begin_date + " - " + service_name
 		if name:
-			if config.recording.filename_composition.value == "short":
+			if config.recording.filename_composition.value == "event":
+				filename = name + ' - ' + begin_date + "_" + service_name
+			elif config.recording.filename_composition.value == "short":
 				filename = strftime("%Y%m%d", localtime(self.begin)) + " - " + name
 			elif config.recording.filename_composition.value == "long":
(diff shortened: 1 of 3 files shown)


littlesat:
Why should we need this? What is the big advantage?

WanWizard:
I can see a use-case, if you record the same event from different channels (I do that quite a lot for series), and you are not using the GUI to access the files (I move them to series folders on my NAS). With a name starting with event, it is easier to sort and select.

AbuBaniaz:
The fourth format will be as pictured.
filename composition

You will only see the change if you **choose** to use the option and inspect the recording folder. The issue has been mentioned a few times on PLi forum.
To me, people record "events", so this should be the prominent/first thing you read on the filename. The date and service on which it was recorded are important, but not more important.

littlesat:
It is another setting and even with the current format the name from the program is in it. Somyou can still "order" and "filter" on it. I think it does not have a real added value. I think even short and long could be removed.
But from the other side it does not really harm as long there are no plugins that are pasing this stuff

WanWizard:
It is not about the event name being "in it", it is about the event name being "at the start". Without it, you need programme experience or special tools to be able to do something smart with them.

So, thanks for merging. ;-)

PR #1065, opened 2017-08-06
GMEPG - increased upper limit for event font size to 32px

ims21:


littlesat:
Why that big? I do not see any reason why????

littlesat:
It is a relative enlargement to what the skin is given.... So I really do not see any reason to do it like this...

ims21:
Do you think, that I did not try it ?  Interesting, that you made  permanently in FullNight skin this size to 26px (and user cannot configure it in menu GMEPG more... )

You can set different number of rows => you  can to want bigger text ...

ims21:
ok, "px" is wrong, but 32 is ok.

PR #1071, opened 2017-08-13
Adapt text in FlashImage.py

betacentauri:
Flash was successful so use close instead of abort.


littlesat:
Oeps... I already did it...

PR #1074, opened 2017-08-16
set bullet as mark for current mode in multiboot list

ims21:


littlesat:
Please do not show bullet... Just give the comment (current image) when mode and image slot are equal...

PR #1076, opened 2017-08-19
NextEpgInfo - added some formating parameters

ims21:
noLabel ... if is set to 1 then is not show 'Next' text
labelColor ... for 'Next' color
timeColor ... time color color

If color parameters are missing, then is used for this parameters color defined in foregroundColor


littlesat:
Do to fix color hex convert.... ;)

littlesat:
merge request has some not relevant added import,,,,

PR #1090, opened 2017-09-03
About.py: use nine characters for driver date

persianpros:
Since dreambox has added an 'a' to the driver date and it's possible others would follow, the driver date is shown as '0140-60-4a' instead of '2014-06-04a' in about screen of enigma2.
Tell "About.py" to use nine characters instead of eight to get rid of that bizarre driver date.
Fix by Hains

### DIFF
--- a/lib/python/Components/About.py
+++ b/lib/python/Components/About.py
@@ -83,7 +83,7 @@ def getCPUInfoString():
 			except:
 				pass
 		if temperature:
-			return "%s %s MHz (%s) %s°C" % (processor, cpu_speed, ngettext("%d core", "%d cores", cpu_count) % cpu_count, temperature)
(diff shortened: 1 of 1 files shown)


WanWizard:
Is it not better to strip the 'a'? What does it signify, why is there a letter in a date? Any idea?

Hains:
http://git.opendreambox.org/?p=opendreambox.git;a=commit;h=7fefec4f45e2b4878b92d8eabaac4b5fa6b46a43

Instead of 20150618 dream choose for 20140604a.

But this patch breaks the driver date for all other boxes.

WanWizard:
Brillant. 

So we need a patch, but one that would convert it to something like "04-06-2014 rev. a"?

PR #1097, opened 2017-09-09
fixed InfoBar import

ims21:


littlesat:
I already reverted the patch thhat causes the GSOD.... And I also could not verify the issue reported on OpenATV...

PR #1104, opened 2017-09-15
[xcore vfd] Enable 12 digit display by default

AbuBaniaz:


WanWizard:
This might lead to another week long discussion about changing defaults... @littlesat ?

littlesat:
This is the wrong way to do it indeed. It has to be done via SystemInfo.py.... e.g. Via systeminfo["defaultvfddisplaydefault"] where you can make it box depended

littlesat:
We should not change this default.... For our international users it needs to be set to 8...

PR #1107, opened 2017-09-17
Estonian translation Update

UrmasP:
Added new strings from 2 September enigma2.pot, Thanks to Pr2!
Translation by Henkka.


WanWizard:
There is a merge conflict, this can't be accepted.

PR #1118, opened 2017-09-24
[satfinder] fix "if" should be "elif" and...

Huevos:
remove unnecessary returns. These functions all return "None" so there is no point sending this back up the chain and the calling function doesn't request a retun value. In my opinion this change makes the code easier to read and understand.

### DIFF
--- a/lib/python/Plugins/SystemPlugins/Satfinder/plugin.py
+++ b/lib/python/Plugins/SystemPlugins/Satfinder/plugin.py
@@ -402,15 +402,13 @@ class Satfinder(ScanSetup, ServiceScan):
 
 	def retune(self, configElement):
 		if self.DVB_type.value == "DVB-S":
-			return self.retuneSat(configElement)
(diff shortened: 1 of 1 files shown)


athoik:
I would like to have the description on the patch, so if somebody examines histrory from git, to know why it was changed.

Huevos:
Hi Athoik.

The description is on the patch. https://github.com/OpenPLi/enigma2/commit/3596ebf5998a411e1b69f42d76b94353869be07e

My personally opinion is the code is confusing. It makes it appear as if something is being returned. I think if the purpose of the return is just to short circuit the function it should be written like this:

```
if something:
	doSomething()
	return
elif somethingElse:
	doSomethingElse()
	return
```

That way it is clear what is being returned is "None".

But... if later we want to add a new bit of code below we can't without removing all the returns. So in my opinion just let the if/else block finish of its own accord.

Example, 
```
if something:
	doSomething()
elif somethingElse:
	doSomethingElse()
new_code_here()
```

PR #1120, opened 2017-09-30
[EventName] if Extended egual Short use only Short description

Dima73:


littlesat:
as I remember this was done by purpose. No short fallback to extended

PR #1124, opened 2017-10-04
[Renderer/Listbox] solve negative value  attrib if update list

Dima73:
When use widget ...render="Listbox"  itemHeight=...
and setList twice, itemHeight reset to default.
When changed not update self.content,if it already exists.
Instead, use it def contentChanged, if you need to update self.content


littlesat:
Sorry I do not understand the issue you tried to solve here. From the forum the uploaded picture is small... I also have no idea how to trigger it.

PR #1129, opened 2017-10-23
ScanSetup.py: Multistream bug fix...

Huevos:
self.scan_sat.is_id is using "0 - 255" as the "limits" to ConfigInteger and "-1" as the "default" value. So the default is out of bounds. This means that once changed it is impossible for the user to return to the default value. This leads to various problems, from not being able to acquire tuner lock when scanning, to writing junk to lamedb. Also showing "-1" as a default value is meaningless to the user. How are they supposed to know this means NO_STREAM_ID FILTER? Zero cannot be used here as the default because that is a valid input stream id.

NO_STREAM_ID_FILTER is a CPP default. It would be good if this variable were available to Python in a similar way to "eDVBFrontendParametersSatellite.PLS_Root" rather than use a hard coded "-1" everywhere that some coders might not understand.

Please see the discussion on using a ConfigSelection field to deal with such a case: https://forums.openpli.org/topic/50339-configlist-item-advice/

### DIFF
--- a/lib/python/Screens/ScanSetup.py
+++ b/lib/python/Screens/ScanSetup.py
@@ -845,6 +845,7 @@ class ScanSetup(ConfigListScreen, Screen, CableTransponderSearchSupport, Terrest
 			self.createSetup()
 
 	def createConfig(self, frontendData):
+		NO_STREAM_ID_FILTER = -1
 		defaultSat = {
 			"orbpos": 192,
 			"system": eDVBFrontendParametersSatellite.System_DVB_S,
@@ -855,7 +856,7 @@ class ScanSetup(ConfigListScreen, Screen, CableTransponderSearchSupport, Terrest
 			"fec": eDVBFrontendParametersSatellite.FEC_Auto,
 			"fec_s2": eDVBFrontendParametersSatellite.FEC_9_10,
 			"modulation": eDVBFrontendParametersSatellite.Modulation_QPSK,
-			"is_id": -1,
+			"is_id": NO_STREAM_ID_FILTER,
 			"pls_mode": eDVBFrontendParametersSatellite.PLS_Root,
 			"pls_code": 1 }
 		defaultCab = {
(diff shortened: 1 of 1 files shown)


athoik:
Is it fast to have a `ConfigSelection` with 256 entries?

Huevos:
Depends what skin you are using. For me I can get through the whole list in 15 steps.

Highlight the ConfigSelection,
Press the movielist button,
press left or right to go through the list.

Not sure of a better way to do it. Zero is a valid value so sooner or later it will be used. Lots of transport streams currently use zero for TSID/ONID..

But anyway  I think 95% of the time people will just use the predefined transponder list anyway.

littlesat:
Mmm a difficult one... an integer you can enter and a choice list is slow... Sounds better to add a new config to set filter on/off. And do not show the filter value when it is off

Huevos:
Probably right. Give me some time and I'll see if I can come up with an elegant way to implement it.

Huevos:
Not as easy as it sounds because "self.scan_sat.is_id" has to run from "-1" to "255". Adding another field doesn't get around the fact that "-1" is absolutely necessary. And this value is used for predefined transponder comparison too so not having "-1" will break that too.

Personally I'd rather have a UI that works properly but that requires a few extra button presses, than one that is a lot more complicated and is just hiding potential bugs.

Also it is not a case of being "off". This variable runs from "-1" to "255" in the CPP code as well.

The real problem is that there is not (to my knowledge) a way of entering negative integers in ConfigInteger.

So the proper fix would be to add a ConfigSignedInteger function.

littlesat:
Sounds you like to need to change more... arrange the value to be 0-255 and add a bool config instead of adding a -1.
Basically the -1 is trick. To do it properly you need to get rid of it...
Good structured solution are not always easy...

littlesat:
Btw a signed integer function is also a work-a-round and indeed cpp code needs to be corrigated...

athoik:
Actually NO_STREAM_ID_FILTER means no multistream at all!

So all the options should be enabled or disabled with a boolean flag as littlesat mentions. 

Enable multistream options when is_id > NO_STREAM_ID_FILTER

Huevos:
Ok, so we are talking about switching to 2 variables where we are currently using 1, which means a lot of code is going to need rewriting in Python. Maybe the same in CPP. Is this fix no good as an interim solution?

littlesat:
Yep indeed.... the original design was not that smart....
But no stream id means no multistream.... furher investigation does it harm as it is now...? Or can't we simple arrange no multistream... change default back to -1... ?

B.t.w. a good discussion :)

Huevos:
I've got a solution. Seems to work, but need to do satfinder too before posting.

Huevos:
Here is my patch (currently in my OpenViX fork).
https://github.com/Huevos/enigma2/commit/4b7e233489dade63dc8d6cdb50660d17e9a0bf98

Huevos:
I noticed something else while testing...
On Lyngsat 30W 11491V is listed without input stream but lists PLS values. What I found is I can enter any value I like for is_id (except -1). Does this mean it is only possible to have PLS when is_id does not match NO_ID_STREAM_FILTER? If so the above patch will need to be modified slightly.

Huevos:
The above gives this: https://github.com/Huevos/enigma2/commit/c15e12adc2cd1a73dc1780ccf182ce8ec2f2bc20





This seems to work well in testing. It opens with the correct defaults and the values stick when switching between "Multistream" and "Ordinary". In "Ordinary" mode is_id, PLS_mode and PLS_code all return to default values. i.e. "-1, 0, 1". No junk in lamedb.

And when tuner is not multistream it also uses the above mentioned multistream defaults.

athoik:
I think this approach is clear, looks good to me. 👍

Huevos:
Ok. I'll close this and start a new PR.

Huevos:
Replaced by this: https://github.com/OpenPLi/enigma2/pull/1133

PR #1131, opened 2017-10-23
 Transponder.py: Rename a variable

Huevos:
"type" should be considered a reserved word as it is a built in Python function.


athoik:
Thanks! Well spotted ;)

PR #1133, opened 2017-10-29
ScanSetup and Satfinder patches

Huevos:


littlesat:
I'm just looking at the code.... And I only can conclude you should not combine the steam_id with NO_STREAM_ID_FILTER.... you need just to add a bool.... stream_id_enabled or so.... en when enable set a value 0-255.. And then indeed you also need some changes in cpp code (db, fastscan, frontend, frontendparms)

The basic principle here is that you should not mix IDs by adding an extra value for NO_STREAM_ID_FILTER in one config/variable....

$ grep NO_STREAM_ID_FILTER *.cpp
db.cpp:                         is_id = NO_STREAM_ID_FILTER,
db.cpp:                         if (static_cast<unsigned int>(sat.is_id) != NO_STREAM_ID_FILTER ||
db.cpp:                         is_id = NO_STREAM_ID_FILTER;
fastscan.cpp:                   fesat.is_id = NO_STREAM_ID_FILTER;
frontend.cpp:   is_id = NO_STREAM_ID_FILTER;
frontend.cpp:           bool multistream = (static_cast<unsigned int>(parm.is_id) != NO_STREAM_ID_FILTER || (parm.pls_code & 0x3FFFF) != 1 ||
frontendparms.cpp:      if (stream_id == NO_STREAM_ID_FILTER) return transponderParameters.is_id;
frontendparms.cpp:      if (stream_id == NO_STREAM_ID_FILTER) return transponderParameters.pls_mode;
frontendparms.cpp:      if (stream_id == NO_STREAM_ID_FILTER) return transponderParameters.pls_code;

littlesat:
[quote]
Here is an example. Open Satfinder. User defined transponder. Tune a multistream transponder. This entails entering an IS ID value between 0 - 255. Now try to tune an ordinary transponder. It is impossible because the GUI does not allow you to return the IS ID to "-1". So once you change the IS ID the only way to tune an "ordinary" transponder is close the plugin and reopen it.
[/quote]
Now I understand... That is why we need the extra bool and hide the filter when the bool is False...
But I think to do it in a structured way we should not recombine this bool in python and swap the -1..255 stuff to cpp.

And sorry for being such 'bitch'...

Huevos:
There is still a problem with that in the GUI. You lose your IS and PLS values when switching in and out of multistream mode.

Huevos:
Littlesat the cpp works fine.

you are focusing on a small part of the problem and missing the inportant thing. IS ID, pls mode and pls code are 3 variables that work together.  In cpp that is happening. In the python code it is not. This means junk is being written to lamedb. My patch fixes this.

littlesat:
I'm not confusing at all... when you have filter values 0-255 and a config for non filter YOU SHOULD not work-a-round here with a -1... you need to add a bool instead with a good descriptive name.

But now you come-up with lamedb... and this also means you need to add the 'flag' in lamedb which is not easily possible at all :-( (I think about setting editors etc). It sounds like the design from the beginning here is somehow crappy... (sorry to say).

Your patch in python looks clean... But still here we should consider to change something 'later'...

Thanks for listening...

Huevos:
The fields are already there in lamedb, it is just that the current python code was feeding in incorrect values from the GUI which, under certain circumstances causes channels to be "grey" in the channel list.

Here is an example:

- Tuner A, internal, satellite, not multistream capable. Simple single 28.2E.
- Tuner B, USB, TBS5925, multistream capable. Simple single 5.0W.
- Currently-playing-service is 5.0W multistream.
- Open ScanSetup. 
- Scan user defined DVB-S2 transponder on 28.2E.
- Go to "last scanned" bouquet. 
- All channels are "grey". 
- Patch fixes this.

As for how a filter should be best handled in the CPP code, that is your area of expertise, not mine.

Thanks for merging.

PR #1138, opened 2017-11-03
[ClockToText] hours with one digit look much better with preceding "0"

Dima73:


athoik (approved):
I prefer this form too.

littlesat:
Should be configurable by the skin and not always

littlesat:
It is a question if taste.... we have it for years without the proceeding zero!

Dima73:
Many times I was asked to fix this.

littlesat:
Is is not a fix...

littlesat:
Nothing was wrong... it is just meaning or tasteZzz

littlesat:
If done make it configurable by skin attribute or config

littlesat:
I told you before!

NietGiftig:
> Nothing was wrong... it is just meaning or tasteZzz

That is correct,  Dima73 has indeed a better taste.

PR #1150, opened 2017-11-22
Better 64-Bit support

persianpros:


pieterg:
Please explain, how is this 'better'?
If you want to force the use of 32bit datatypes on 64bit, use (u)int32_t, instead of unsigned int.
unsigned int might still be larger than 32bit. (the standard dictates it should be 'at least' 32bits, not 'exactly' 32bits)

PR #1155, opened 2017-11-26
update cs.po

ims21:


littlesat:
Shoudn’t We consider to merge this in rc?

PR #1156, opened 2017-11-26
ioprio.cpp: define the correct values for ARM and AARCH64

persianpros:


athoik:
Your patch is correct, although I would prefer to use a header that provides the syscall information to userspace without have to "find them" every time we like to support new architectures. 

Please have a look here: https://gist.github.com/athoik/6ff1e1b5f11b5207cdb38ba21b2f87d7

persianpros:
Your idea is better but when do you want do it? Maybe you can merge this now and do your thing later :)
Maybe helpful: https://bugzilla.redhat.com/show_bug.cgi?id=CVE-2016-7911

athoik:
I do not understand how the `CVE-2016-7911` affects enigma2.

I guess you are free to create PR against affected manufacturer BSP!

The PR for using syscall header is available here #1163. 

No need to commit this one any more.

PR #1163, opened 2017-11-28
ioprio: include syscall.h for SYS_ioprio_get/set

athoik:
The sys/syscall.h header contains the required SYS_ioprio_get and SYS_ioprio_set values.
So we don't need to define them manually.

The same approach is used also now in ionice: https://github.com/karelzak/util-linux/blob/master/schedutils/ionice.c

Also remove unused includes getopt.h, sys/ptrace.h and asm/unistd.h


persianpros:
+1 👍

PR #1164, opened 2017-11-28
sk.po

Pr0metheus2:
Small corrections


Pr0metheus2:
Damn, its jumping into develop...

PR #1169, opened 2017-11-29
Vu Zero 4K support

persianpros:


technl:
shouldn't the new name (vu2) also been used in rc_models.cfg?

in that file the old name (vuuno4kse) is still used instead of vu2....

persianpros:
Check all commits please, It's there ;)

technl:
You're right, I need new glasses :)

persianpros:
Not a big deal :) Also https://github.com/OpenPLi/openpli-oe-core/pull/361 is related ...

PR #1175, opened 2017-12-18
scan.cpp: Namespace subnet by DVB type

Huevos:


lib/dvb/scan.cpp
```diff
@@ -125,8 +125,16 @@ int eDVBScan::isValidONIDTSID(int orbital_position, eOriginalNetworkID onid, eTr
...
-		// on valid ONIDs, ignore frequency ("sub network") part
-	if (eConfigManager::getConfigBoolValue("config.usage.subnetwork", true)
+	// on valid ONIDs, ignore frequency ("sub network") part
+	if ((((hash >> 16) & 0xFFFF) == 0xFFFF) // cable
+		&& eConfigManager::getConfigBoolValue("config.usage.subnetwork_cable", true))
+		hash &= ~0xFFFF;
+	else if ((((hash >> 16) & 0xFFFF) == 0xEEEE) // terrestrial
+		&& eConfigManager::getConfigBoolValue("config.usage.subnetwork_terrestrial", true))
+		hash &= ~0xFFFF;
+	else if ((((hash >> 16) & 0xFFFF) != 0xFFFF) // not cable
```
athoik, line 135: No need to check for cable here, it will match in the first `if` above (same applies for terestrial).

Also there is a check for t/c in isValidOnidTsId, I guess it can be removed, because it will be dead code.

Huevos:
Athoik I don't understand what you mean.

First "if" clause checks for cable namespace and that cable_subnet is enabled. i.e checking for multiple things. If cable namespace is true but cable subnet is false that "if" clause will fail, which is the correct behaviour and it will pass to next if clause. But we don't want cable matching in subsequent clauses so the test for not cable is necessary.

Huevos:
> Can you please explain why we need this?

Currently you can turn namespace subnet on and off. That setting is global. This commit allows this setting to be done individually based on DVB type.

athoik:
@Huevos, see what I mean:

```
@@ -53,12 +53,6 @@ eDVBScan::~eDVBScan()

 int eDVBScan::isValidONIDTSID(int orbital_position, eOriginalNetworkID onid, eTransportStreamID tsid)
 {
-       /*
-        * Assume cable and terrestrial ONIDs/TSIDs are always valid,
-        * don't check them against the satellite blacklist.
-        */
-       if (orbital_position == 0xFFFF || orbital_position == 0xEEEE) return 1;
-
        int ret;
        switch (onid.get())
        {
@@ -125,9 +119,15 @@ int eDVBScan::isValidONIDTSID(int orbital_position, eOriginalNetworkID onid, eTr

 eDVBNamespace eDVBScan::buildNamespace(eOriginalNetworkID onid, eTransportStreamID tsid, unsigned long hash)
 {
-               // on valid ONIDs, ignore frequency ("sub network") part
-       if (eConfigManager::getConfigBoolValue("config.usage.subnetwork", true)
-               && isValidONIDTSID((hash >> 16) & 0xFFFF, onid, tsid))
+       int orb_pos = (hash >> 16) & 0xFFFF;
+       if (orb_pos == 0xFFFF) // cable
+               if (eConfigManager::getConfigBoolValue("config.usage.subnetwork_cable", true))
+                       hash &= ~0xFFFF;
+       else if (orb_pos == 0xEEEE) // terrestrial
+               if (eConfigManager::getConfigBoolValue("config.usage.subnetwork_terrestrial", true))
+                       hash &= ~0xFFFF;
+       else if (eConfigManager::getConfigBoolValue("config.usage.subnetwork", true)
+               && isValidONIDTSID(orb_pos, onid, tsid)) // on valid ONIDs, ignore frequency ("sub network") part
                hash &= ~0xFFFF;
        return eDVBNamespace(hash);
 }
```

Huevos:
Ok, thanks, I'll sort it out and submit again.

Huevos:
Closing... new request sent.

PR #1192, opened 2018-01-02
[dvbtime] More stable time-keeping

Schimmelreiter:
Slew small time diffs (Up to 15sec) rather than stepping them.

Tiny diffs (1-5sec) are normal even between good transponders, so permanent stepping would only make the system time jump around.
The clock jumping around by few seconds adds absolutely no value when it comes to accuracy but can ruin conditional access.

Slewing the system clock takes ages though (About 5sec correction in 3h), so it can not be used for large diffs.

Slewing small diffs and stepping only large ones reproduces the behaviour of ntpd/ntpdate.
This commit should remove the reason why oscam added the so-called CLOCKFIX (Which simply reverts all time changes in E2 by setting the last known time again), so preferably build oscam WITHOUT clockfix now.

### DIFF
--- a/lib/dvb/dvbtime.cpp
+++ b/lib/dvb/dvbtime.cpp
@@ -393,13 +393,32 @@ void eDVBLocalTimeHandler::updateTime( time_t tp_time, eDVBChannel *chan, int up
 					now.tm_sec);
 				time_difference = rtc_time - linuxTime;
 				eDebug("[eDVBLocalTimerHandler] RTC to Receiver time difference is %ld seconds", linuxTime - rtc_time );
-				if ( time_difference )
-				{
-					eDebug("[eDVBLocalTimerHandler] set Linux Time to RTC Time");
-					timeval tnow;
-					gettimeofday(&tnow,0);
-					tnow.tv_sec=rtc_time;
-					settimeofday(&tnow,0);
+				if ( time_difference ) {
+					if ( (time_difference >= -15) && (time_difference <= 15) ) {
+						// Slew small diffs ...
+						// Even good transponders can differ by 0-5 sec, if we would step these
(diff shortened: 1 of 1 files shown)


athoik:
Nice idea. 

Maybe add %m in order to display the system error as well?

```
eDebug("[eDVBLocalTimerHandler] slewing Linux Time by %03d seconds FAILED (%m)", time_difference);
```

Schimmelreiter:
Considered it, but was not able to test ist, because it never failed :)
But add whatever you like ...

eriksl:
The variable 't', it puzzles me. Apparently it's "somewhere" because it's referred to in other parts of the function as well, but I can't find the definition. You could say that's because I am lousy at coding or bad at coding, but I'd rather call it "code difficult to maintain" and therefore unwanted. Please refer to me where it's defined and I'll give it a more suitable name and some better maintainable code.

Schimmelreiter:
Dunno where it comes from, was there before in OpenATV.

Schimmelreiter:
Rather:
```s/tnow.tv_sec=t/tnow.tv_sec=rtc_time/```

eriksl:
I was going to say, if it's removed, it won't work ;-)

Schimmelreiter:
Damn code drift ...

Schimmelreiter:
Actually it was intended. ;)
I believe that any merge deserves a "Fix merge" commit so that it isn't alone. ;)

eriksl:
That's why I cherry pick them and then amend them. But don't worry, it's accompanied by a "coding style" commit from me. No fuss, it's not only your patch.

eriksl:
Succesfully cherry-picked, amended and applied. Merge is not required.

PR #1212, opened 2018-01-16
pmt: fix HbbTV url when AIT descriptor order is not the expected one

athoik:
When SIMPLE_APPLICATION_LOCATION_DESCRIPTOR is found before the TRANSPORT_PROTOCOL_DESCRIPTOR
then HbbTV url was set only to TRANSPORT_PROTOCOL_DESCRIPTOR loosing changes from SIMPLE_APPLICATION_LOCATION_DESCRIPTOR.

Now when TRANSPORT_PROTOCOL_DESCRIPTOR is found we are inserting url in front. So at the end of the loop we have
the right HbbTV url including the SIMPLE_APPLICATION_LOCATION_DESCRIPTOR.

Also push to m_aitInfoList when going out of loop, for the same reason. We cannot expect that order will be always
TRANSPORT_PROTOCOL_DESCRIPTOR followed by SIMPLE_APPLICATION_LOCATION_DESCRIPTOR.

This commit fixes the HbbTV url creation on ERT DIGEA HbbTV.

Credits: VU+ http://code.vuplus.com/gitweb/?p=vuplus_dvbapp;a=blobdiff;f=lib/dvb/pmt.cpp;h=ded43a03a64861d3b384f1160f0c53a439501cfd;hp=fbd9b8f815285b0f8f8ebc6e3dd16812007433a9;hb=097eb7b26bdc1fc761a745b71bfab29e076856c0;hpb=c1bbfb59b91856990845e3092380dc2a0b1da257

### DIFF
--- a/lib/dvb/pmt.cpp
+++ b/lib/dvb/pmt.cpp
@@ -282,9 +282,9 @@ void eDVBServicePMTHandler::AITready(int error)
 							{
 								if ((*i)->getApplicationControlCode() == 0x01) /* AUTOSTART */
 								{
-									m_HBBTVUrl = (*interactionit)->getUrlBase()->getUrl();
+									m_HBBTVUrl.insert(0, (*interactionit)->getUrlBase()->getUrl());
 								}
-								aitinfo.url = (*interactionit)->getUrlBase()->getUrl();
+								aitinfo.url.insert(0, (*interactionit)->getUrlBase()->getUrl());
 								break;
 							}
 							break;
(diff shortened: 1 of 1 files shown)


athoik:
Once commited to release-4 please cherry-pick to release-6.X.

WanWizard:
> Once commited to release-4 please cherry-pick to release-6.X.

Done. This also merges it back into develop, is that ok?

athoik:
Once we merge release-6.1 to develop it will become available. 

If we are not going to merge 6.1 back to develop, then we should cherry-pick to develop as well.

WanWizard:
Was just checking if you wanted that to happen. ;-) Merging now.

PR #1215, opened 2018-01-19
Skin.py: don't make "objecttypes" conditional

Huevos:
Don't make this otherwise very useful new feature conditional as doing so will severely limit its usefulness.

### DIFF
--- a/skin.py
+++ b/skin.py
@@ -980,12 +980,12 @@ def readSkin(screen, skin, names, desktop):
 	def process_screen(widget, context):
 		for w in widget.getchildren():
 			conditional = w.attrib.get('conditional')
-			if conditional:
-				if not [i for i in conditional.split(",") if i in screen.keys()]:
-					continue
-				objecttypes = w.attrib.get('objecttypes')
-				if objecttypes and not [i for i in objecttypes.split(",") if i == screen[conditional].__class__.__name__]:
-					continue
+			if conditional and not [i for i in conditional.split(",") if i in screen.keys()]:
+				continue
+			key = w.attrib.get('name') or w.attrib.get('source')
+			objecttypes = w.attrib.get('objecttypes')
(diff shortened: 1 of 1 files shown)


littlesat:
Sorry I need to investigate this one... Yesterday I did not get it to work for e.g. FlashImage and Hotkey.

littlesat:
I verified it... it cannot work... Sorry

Huevos:
Please look at this. Works fine and does not interfere with the conditional code.

https://github.com/Huevos/enigma2/blob/df6750aa5f2dd54cbb4da0fe3415ecda272f260f/skin.py#L1057-L1061

Huevos:
By the way your code is clearly wrong because a conditional can hold multiple conditions. And an object name will never contain a comma.

littlesat:
But is has side effect.... it is somehow not working... time you can try a develop image ;)

littlesat:
And you are right if I understand you... when the conditional has more then one thing te check in an or it cannot work... but first I need to fix and find the why of that site effect...

littlesat:
One side effect is e.g. ePixmap does not have a name or source... but you need to check an object being a specific class... probably we should consider having the first parameter of the value being the key to check for...

littlesat:
Instead of defining a new key parameter...

littlesat:
Or even do ‘key_red=Botton,Label’ and seperate also in =...

nautilus7:
Regarding hotkey... On this commit https://github.com/OpenPLi/enigma2/commit/5e89e8a5cd7ead533b8b6f9bab800ae14e9d0dcc, in line 344 you rename method "yellowButton" to "yellow", but in line 395, the def still remains " yellowButton.

Huevos:
Discussion continued on OpenPLi forum.
https://forums.openpli.org/topic/56924-button-and-label-pli-opinion/page-12#entry834755

PR #1216, opened 2018-01-20
While in pause for non DVB type display the name and length

pr2git:
While in pause for non DVB type display the name and length of the current playing service.

This is an addition to:
https://github.com/OpenPLi/enigma2/commit/e89a7bb4e97ab6283be4433e728c57ffe3760ac8


athoik:
Better to show something than nothing!

PR #1236, opened 2018-01-30
Convert "TimerEntry" class buttons to new style

nautilus7:
I tested it and it works ok. I renamed "OK" to "Save", as it better describes the creation or edit action on a timer.

### DIFF
--- a/data/skin_default.xml
+++ b/data/skin_default.xml
@@ -1215,12 +1215,12 @@ self.instance.move(ePoint(orgpos.x() + (orgwidth - newwidth)/2, orgpos.y() + (or
 	</screen>
 	<!-- Timer entry -->
 	<screen name="TimerEntry" position="center,center" size="560,440" title="Timer entry">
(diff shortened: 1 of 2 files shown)


littlesat:
It also needs some changes to the skin..
Cancelok templates can be temoved

RobvanderDoes:
But for backwards compatibility reasons you can leave the old name in as well as the new name.

littlesat:
Sorry I did not merge this one as I do not understand the need....

nautilus7:
Why? The need is similar to this: https://github.com/OpenPLi/enigma2/commit/2a7abe5d9df27b3b23815771dbc4977e06f0001d#diff-85d20ea518b17e47e87a1fb6ac80d56cL34

You changed "ok" and"cancel" button names to "key_red" and "key_green". 
I also changed Label() to staticText(), to be consistent with the rest of the buttons in that file.

PR #1254, opened 2018-02-08
Fix text messages in two screens

nautilus7:
Regarding Cable Scan, the changes are minimal, just a little bit better wording used in some expressions.

Regarding Factory Reset message, apart from giving a more detailed message, I omitted the unnecessary  line breaks (\n) to better display the message in skins. See example picture.

factoryreset


Dima73:
What for?
How much can I fix the localization?
I understand when they add a new one.
But this is a very old text.

nautilus7:
I am not sure I understand what you mean...

Do you mean that I (or anyone else) should not change any text message, because then the translators will have to redo the translations? I am sorry, but this in not a good way to view things. Should we stick with the old messages that they are written in a bad way (bad English, bad formatting, etc.) in order not having to redo translations?

RobvanderDoes:
Of course text should be enhanced wherever and whenever possible. Translating is simply a never ending hobby!

WanWizard:
Perhaps, to be complete, add that recordings will not be deleted?

P.S. Thanks, cleaning up the sometimes toe-curling English used is long overdue!

nautilus7:
Hi, thanks.

Recordings are kept if they are stored into usb or hdd only, right? Maybe we should add an additional sentence like "Files stored on external media like HDD and USB drives will stay intact."

RobvanderDoes:
"External" doesn't sound right; an HDD can be internal. So maybe:
"Files stored on media like HDD and USB drives will remain intact."

littlesat:
I like the factory reset patch... the Cablescan patch is not really required I suppose. Please offer these patches seperately

nautilus7:
@RobvanderDoes yes, you are right!

@littlesat, the Cable Scan fixes are more necessary than the factory reset fix in my opinion. Changes like:

Scanning failed --> Scan failed
Scanning completed --> Scan completed

I think this is the correct English wording. Others can confirm (or not).

WanWizard:
Confirmed. "The scan is scanning, when finished, the scan is complete".

nautilus7:
I did the change/addition to the "Factory reset" message. The bold sentence is the additional one.

> Factory reset will restore your receiver to its default configuration. All user data including system settings, tuner configuration, bouquets, services and plugins will be DELETED. **Recordings and other files stored on HDD and USB media will remain intact.** After completion, the system will restart automatically!
> 
>  Do you really want to proceed?


I messed things a little bit with git and it shows an additional merge commit now. Is it ok, or should I make new pull request?

Are we ok to keep changes to both CableScan and FactoryReset modules?

WanWizard:
As long as the merge is clean, it isn't a problem. We usually do a "squash and merge", which sqeezes everything into a single commit.

nautilus7:
Nice, thanks for the explanation.

On my local branch, I did "amend last commit" to update the "FactoryReset" commit with the new message, but as I found out, I should have done a pull first to get any new commits, and then a push. I did them vice versa.

PR #1258, opened 2018-02-10
qviart and zgemma h9 remote support

persianpros:


persianpros:
I don't understand what's wrong with rc_models.cfg?

athoik:
The build failed on 9dff1ae, because that commit is wrong, it is fixed on 82781dd.

You should "squash" commits 9dff1ae and 82781dd, if you like CI to build properly.

Check: https://travis-ci.org/OpenPLi/enigma2/builds/339796043

PR #1264, opened 2018-02-10
[FlashImage] create separate skin for compatibility with various skin…

Dima73:
…s, e.g. Default skin


littlesat:
What skins does it break and why?

Dima73:
https://github.com/OpenPLi/enigma2/blob/develop/data/skin_default.xml#L245

PR #1269, opened 2018-02-14
Typo fix

persianpros:
occured -> occurred
I'm updating my Persian translation after months and it seems we have some typos ...

### DIFF
--- a/lib/python/Screens/About.py
+++ b/lib/python/Screens/About.py
@@ -408,7 +408,7 @@ class Troubleshoot(Screen):
 
 	def appClosed(self, retval):
 		if retval:
-			self["AboutScrollLabel"].setText(_("Some error occured - Please try later"))
+			self["AboutScrollLabel"].setText(_("Some error occurred - Please try later"))
 
 	def dataAvail(self, data):
 		self["AboutScrollLabel"].appendText(data)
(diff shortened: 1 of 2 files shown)


nautilus7:
Better change the strings from this:

> Some error occurred - Please try later

to this:

> An error occurred. Please try again later.

persianpros:
I didn't want to change the string just fix the typo ;)

WanWizard:
I agree with @nautilus7 that if you update it, better make sure it's correct. Not a lot of effort.
There is already enough "toe-curling" English in Enigma, lets do something about it ... ;)

persianpros:
Ok so wait for my next merge request as I have to check other files too :)

WanWizard:
All the time in the world... :ok_hand:

nautilus7:
@WanWizard, I though you agreed that it was better we change the strings to what I suggested above.

persianpros:
Don't worry I'll change some strings this week as I have to check more files ;)

nautilus7:
OK good to know, BUT since it is already committed, translators might start translating these new strings, and then they will have to do them again.

@WanWizard said

> if you update it, better make sure it's correct

and then he committed the not-so-correct version...  I'm confused.

WanWizard:
Sorry, my mistake, wasn't paying (enough) attention. I'll commit the updates in a minute.

PR #1273, opened 2018-02-15
Update plugin.py - added FastScan provider - freeSAT [UPC Direct] Slovakia

Pr0metheus2:
Please revise line 131 for number 32 - not sure if it's correct, or where to get it from


Pr0metheus2:
No, index (32) is not correct. I will close this request unless I know correct value.

PR #1277, opened 2018-02-17
MovieSelection: Use established variable names so we don't need to...

Huevos:
...create a new screen in the skin.

We can't be just inventing new variable names every time someone writes a new class.

This needs fixing immediately while the class is still brand new.

### DIFF
--- a/data/skin_default.xml
+++ b/data/skin_default.xml
@@ -701,9 +701,9 @@ self.instance.move(ePoint(orgpos.x() + (orgwidth - newwidth)/2, orgpos.y() + (or
 		<widget name="key_yellow" position="280,0" size="140,40" valign="center" halign="center" zPosition="4"  foregroundColor="white" font="Regular;20" transparent="1" shadowColor="background" shadowOffset="-2,-2" />
 		<widget name="key_blue" position="420,0" size="140,40" valign="center" halign="center" zPosition="4"  foregroundColor="white" font="Regular;20" transparent="1" shadowColor="background" shadowOffset="-2,-2" />
 		<ePixmap pixmap="skin_default/buttons/key_menu.png" position="560,8" size="40,30" zPosition="2" alphatest="on"/>
-		<widget name="files" position="5,50" zPosition="2" size="590,308" itemHeight="28" font="Regular;20" foregroundColor="white" scrollbarMode="showOnDemand" />
+		<widget name="config" position="5,50" zPosition="2" size="590,308" itemHeight="28" font="Regular;20" foregroundColor="white" scrollbarMode="showOnDemand" />
 		<ePixmap pixmap="skin_default/div-h.png" position="5,363" zPosition="2" size="590,2" />
-		<widget name="text" position="5,368" zPosition="2" size="590,50" valign="center" halign="left" font="Regular;20" foregroundColor="white" />
+		<widget name="description" position="5,368" zPosition="2" size="590,50" valign="center" halign="left" font="Regular;20" foregroundColor="white" />
 	</screen>
 	<!-- Mute -->
 	<screen name="Mute" position="50,50" zPosition="10" size="34,45" backgroundColor="transparent" title="Mute" flags="wfNoBorder">
--- a/lib/python/Screens/MovieSelection.py
+++ b/lib/python/Screens/MovieSelection.py
@@ -2159,8 +2159,8 @@ class MovieSelectionFileManagerList(Screen):
 						index += 1
 		self.list = data
 
-		self["files"] = self.list
-		self["text"] = Label()
+		self["config"] = self.list
+		self["description"] = Label()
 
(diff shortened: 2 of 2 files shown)


nautilus7:
While I agree with using "established" variable names, can you please explain how a new skin screen won't be needed? Is there a trick to use a universal skin screen for many classes?

I've seen a way that defines multiple values in the "skinname" property, so an alternative screen can be used if the primary screen is not present in the skin. Is that what you mean?

ims21:
I can agree with "description" instead "text". But rename list with files to "config" is nonsence, imho!

Huevos:
At nautilous7.

config and list are used almost universally throughout the image.

You can make one standard templete. It can be shared between lots of screens including this one.

If you don't use standard names this is impossible.

In the skin I am working on I have one config template. It is used by satconfig.py, setup.py and many other screens including this one. Al the elements appear and disappear as required by the screen.

Huevos:
At IMS, I am not suggesting there is anything wrong with your code. It is just that if we stick to the same accepted names skins can be a whole lot simpler.

Huevos:
Btw, nice addition to movie selection.

RobvanderDoes:
I have two suggestions to this useful filemanager.
Add to the optional operations (so in addition to the existing 'Move' & 'Copy'):
1- 'Delete' 
2- 'Save as playlist'

nautilus7:
@Huevos, I see what you mean regarding skinning, but as a programmer, using descriptive names for my variables is more important! Ease of skinning comes second to my list.

In addition this method of skinning is a side effect of using these common variable names. It's not something intended. It just happened.

Of course, in cases like using "description" instead of "text" where the meaning is not altered is ok, but using "config" instead of "files" I think is wrong as @ims21 said.

Anyway, this pull request has a wrong and misleading title. By reading it in the first place I thought that there was some method of skinning (that I wasn't aware of) using the same screen (name) for many classes! But this is not the case. You still need a new screen! You just want all classes to have the same variable names, so you can use the same TEMPLATE.

Oh, and you pull request is wrong! See the comments: https://github.com/OpenPLi/enigma2/commit/3fc171a41661efd0b6b3451efe8aab1b2c60958d#comments

Huevos:
@nautilus7, please explain. If every other class in enigma2 uses "config" or "list" when dealing with a "ConflgList", why should this class be any different?

Huevos:
And by the way you don't even need a new screen in the skin.

nautilus7:
@Huevos, It doesn't change anything. The  coder must choose variable names that he feels they describe his code function better. Anyway, this is the least important thing... **It's just cosmetic.** There's no point in talking this further.

Regarding skinning, I asked you to explain how I can use an existing skin for this new screen... You told me that you can use your already existing template! Well, this won't require a new screen in YOUR skin. Everybody else's skin will need a new screen!

Huevos:
I think you need to read up on model-view-controller. A designer should not need to know the names of the variables used in the code.

https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller

nautilus7:
Maybe in a perfect world, he won't need to know the variable names, not in enigma2...

I will read your link though, thanks.

Huevos:
Yes, I agree, enigma2 is not a perfect world, but we can we can work towards one.

Everyone can see the benefits of calling the red button "key_red" and now we have an automatic button bar. My suggestion is to spread the same philosophy as far as possible to make things easier for everyone. Having some discipline when creating variable means code reuse, and not forever reinventing the wheel.

And BTW, I am not saying using whatever variable names you like is "wrong", but that it prevents code reuse.

ims21:
But priority is try changes on box before create PR, imho ...

Huevos:
Until such a point when PLi provide a developer image that is very difficult for developers without a build server.

nautilus7:
You can just replace the python file in your box and see if it works.

Huevos:
Of course you can't. This branch is not built. Are you suggesting testing a Python file from one branch in another branch?

Anyway this has been discussed on the forum and is a known problem.

ims21:
What do you write about nonsense ? Filemanager is in 'develop' branch only ! You cannot test this branch but you are changing code in this branch ? 
Omg...

Huevos:
IMS, this issue has been discussed on the forum. Yes we test our code but we can't test in PLi DEV branch. And this is not my fault. PLi have created this situation. All developers have been told to commit changes to the develop branch, but no image is being built. So what do we do? All walk away and turn our backs on PLi. I've committed lots of code to PLi over the years, but now PLi has made it almost impossible for external developers to contribute. Littlesat keeps asking for externally produced patches to be pushed back to PLi. But how can we if there is no image provided?

ims21:
but there on forum are last 5 night build develop images for download ...

Huevos:
For selected PLi members only.

WanWizard:
Please lets not discuss this in lots of different places, give me some time please... :sweat_smile:

PR #1279, opened 2018-02-18
Update some messages, Still I have no idea about https://github.com/O…

persianpros:
…penPLi/enigma2/blob/develop/lib/python/Screens/PictureInPicture.py#L192


nautilus7 (changes requested):
I am sorry, but after "," (comma) you can't use an upper case letter.

In addition, these strings are OK. Changing them is not necessary in the first place.

PR #1288, opened 2018-02-20
Greek translation update 20/2/2018

MCelliotG:
Greek translation update 20/2/2018

### DIFF
--- a/po/el.po
+++ b/po/el.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: OpenPLI Enigma2 Official Greek Translation\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2018-02-08 11:28+0100\n"
+"POT-Creation-Date: 2018-02-19 21:31+0100\n"
 "PO-Revision-Date: \n"
 "Last-Translator: MCelliot_g <mcelliot_g@hotmail.com>\n"
 "Language-Team: MCelliot_g <mcelliot_g@hotmail.com>\n"
@@ -609,7 +609,7 @@ msgstr ""
 "Ένας ολοκληρωμένος χρονοδιακόπτης θέλει να τερματίσει\n"
 "το δέκτη σας. Να γίνει τώρα;"
 
-msgid "A graphical EPG for all services of an specific bouquet"
+msgid "A graphical EPG for all services of a specific bouquet"
 msgstr "Γραφικό EPG για όλες τις υπηρεσίες σε συγκεκριμένο μπουκέτο"
 
 #, python-format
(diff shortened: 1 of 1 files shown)


po/el.po
```diff
@@ -2813,6 +2816,24 @@ msgstr "FEC:"
...
+"Do you really want to proceed?"
+msgstr ""
+"Η εργοστασιακή επαναφορά θα επαναφέρει το δέκτη σας στις αρχικές του "
+"ρυθμίσεις. Όλα τα δεδομένα χρήστη, συμπεριλαμβανομένων των ρυθμίσεων "
+"συστήματος, των ρυθμίσεων tuner, τα μπουκέτα και τα πρόσθετα ΘΑ ΔΙΑΓΡΑΦΟΥΝ. "
+"Οι εγγραφές και άλλα αρχεία που είναι αποθηκευμένα στους δίσκους HDD και USB "
+"θα παραμείνουν ως έχουν. Μετά την ολοκλήρωση, το σύστημα θα επανεκκινηθεί "
+"αυτόματα!\n"
+"\n"
+"Θέλετε σίγουρα να συνεχίσετε; "
```
nautilus7, line 2835: There is an extra space at the end.

po/el.po
```diff
@@ -5692,12 +5722,24 @@ msgstr "Ανίχνευση μπάντας %s"
 msgid "Scan additional SR"
 msgstr "Ανίχνευση πρόσθετου SR"
 
+#, python-format
+msgid "Scan completed, %d channel found."
+msgid_plural "Scan completed, %d channels found."
+msgstr[0] "Η ανίχνευση ολοκληρώθηκε, βρέθηκε %d κανάλι"
+msgstr[1] "Η ανίχνευση ολοκληρώθηκε, βρέθηκαν %d κανάλια"
```
nautilus7, line 5729: Consider adding a full stop at the end in these 2 strings.

nautilus7 (changes requested):
I made 2 minor suggestions on this updated version. Otherwise everything looks OK!

po/el.po
MCelliotG: Thank you, I've seen these already, but I had commited my pull, so they will be visible in the next pull request. These are new strings that changed recently, the auto suggest ignored some. Some strings are missing though from the new pot, like Press OK to start the scan. I don't know what happened and they are gone!

nautilus7: Don't you use git? If you did you could easily change you pull request by amending last commit. Anyway, they are not big issues. They can be fixed in the next revision of the translation.

That specific string you mention that is gone, is indeed gone! I changed it in the code and made it identical to another one, so 1 less string to translate. See commit https://github.com/OpenPLi/enigma2/commit/ed09d8025584a898fb42870d9f1f21bf4c380ddf

In addition, some other strings are completely removed, because they were obsolete.

nautilus7:
This PR can now be closed. An updated version is already merged.

MCelliotG:
Definitely right, thank you to everyone!

PR #1306, opened 2018-02-25
[skin_default] update StreamingClientsInfo for new format key

Dima73:


nautilus7:
You are not supposed to do it like this. The purpose of these conditional buttons is to create a template for your buttons and put it inside the basic skin template. Then you can use that basic template for all your screens, without having to know how many and which buttons a specific screen has.

This skin does not use templates at all, so there is not point in using conditional buttons per screen.

PR #1307, opened 2018-02-25
No need to restart GUI to active new language select

fairbird:


littlesat:
You’re wrong it is needed! See why we put it in. Parra are not (re)translated if you don’t... we did it on purpose.

PR #1318, opened 2018-03-01
Removes unnecessary functions in TextBox

Taapat:
Why for the close use separate functions ok and cancel if we can use close directly in ActionMap.


nautilus7 (approved):
Indeed!

PR #1323, opened 2018-03-05
moviemanager added into configurable buttons list too

ims21:


ims21:
It cannot be added by WHERE_MOVIELIST, because plugin must be called from MovieSelection due list.

PR #1339, opened 2018-03-10
fixed regression for bcm7401 based receivers as the kernel won't report mips on them

lprot:


littlesat:
There is more than one change in this merge request

lprot:
Previous two are already merged.

eriksl:
It's irrelevant, we don't have machines using bcm7401 in our portfolio.

lprot:
dm800, dm800se don't know about others, but before arm worked ok there

littlesat:
We do not support these dmm boxes anymore so still not relevant...

lprot:
well, the more akward way will be to patch the kernel on them :)

eriksl:
> dm800, dm800se don't know about others, but before arm worked ok there
Wrong, dm800se = bcm7405. So only dm800 left and we stopped supporting that one in OpenPLi 2.1...

lprot:
Hmm, right
```
root@dm800se:/# uname -a
Linux dm800se 3.2-dm800se #1 SMP Sat Sep 28 00:31:25 UTC 2013 mips GNU/Linux
```

PR #1342, opened 2018-03-11

IanSav:
This change eliminates the forced crash if a setup menu id is not found in the setup.xml file.  This sort of error should be identified during development and not reach production code.  Forcing a crash is unnecessary and anti-social.  The crash could have nasty side effects like stopping running recordings.  Instead of the crash an error condition is logged and the user is provided with a dummy menu item indicating that this menu item is unavailable.  The error contains the ID tag so that the issue can be reported to developers for correction.

The title returned is now translated to the local language.

To test the change edit setup.xml and change the "key=" attribute for an item in a Setup based menu.  Before the code change loading the menu containing this menu item will force a crash of the GUI.  After the code change an error will be logged and the menu will be displayed but the menu item that can't be found will be shown as an error message.  (It is not fatal to select this menu item.  If selected a blank Setup screen will be shown.)

### DIFF
--- a/lib/python/Screens/Setup.py
+++ b/lib/python/Screens/Setup.py
@@ -150,6 +150,8 @@ def getSetupTitle(id):
 	for x in xmldata.findall("setup"):
 		if x.get("key") == id:
 			title = x.get("title", "Settings").encode("UTF-8")
+			if title == "":
+				title = "** Setup error: '%s' title is blank!" % id
 	if title == "":
 		print "[Setup] Error: Setup ID '%s' not found in setup file!" % id
 		title = "** Setup error: '%s' section not found! **" % id
--- a/lib/python/Screens/Setup.py
+++ b/lib/python/Screens/Setup.py
@@ -27,12 +27,6 @@ def getConfigMenuItem(configElement):
 			return _(item.attrib["text"]), eval(configElement)
 	return "", None
 
-class SetupError(Exception):
-	def __init__(self, message):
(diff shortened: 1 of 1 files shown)


littlesat:
But is does not work as when it is not available title becomes “Settings” and not empty... as far I can see??

Huevos:
The exception is only raised if setup.xml is broken. Why would you want to hide an exception that shows a broken setup.xml file? Quickest way for a broken file to get fixed is a BSoD. This exception has been put in the code for precisely that reason. Why anyone would want to override that behaviour is beyond me.

IanSav:
@Littlesat: I believe that you saying that if the title attribute is present but set to "" then the error message is triggered even though the setup block is present.  Would you prefer a different error path to a specific message or would you like the current wording modified?  My preference would be to have a separate more appropriate message.

I will submit an update in a moment.

IanSav:
@Huevos: Forcing a system crash for a minor error is overkill and user hostile.  This could be likened to driving along a freeway and the car slams on the brakes and shuts down because the radio didn't turn on properly when the power button was activated.  Yes, it is an issue but the reaction is inappropriate.

The code supplied provides both logged and user feedback regarding the problem.  Throwing a BSoD during development is a developer's choice.  This is production code and system stability should be the greater priority.

IanSav:
I accidentally closed the this pull request.   Pull Request #1344  is this original pull request together with the change suggested / requested by Littlesat.

PR #1344, opened 2018-03-12

IanSav:
This change eliminates the forced crash if a setup menu id is not found in the setup.xml file. This sort of error should be identified during development and not reach production code. Forcing a crash is unnecessary and anti-social. The crash could have nasty side effects like stopping running recordings. Instead of the crash an error condition is logged and the user is provided with a dummy menu item indicating that this menu item is unavailable. The error contains the ID tag so that the issue can be reported to developers for correction.

Blank "title=" attributes now have a default value of "Settings" to stop blank titles being fed into the translation system.

To test the change edit setup.xml and change the "key=" attribute for an item in a Setup based menu. Before the code change loading the menu containing this menu item will force a crash of the GUI. After the code change an error will be logged and the menu will be displayed but the menu item that can't be found will be shown as an error message. (It is not fatal to select this menu item. If selected a blank Setup screen will be shown.)

### DIFF
--- a/lib/python/Screens/Setup.py
+++ b/lib/python/Screens/Setup.py
@@ -27,12 +27,6 @@ def getConfigMenuItem(configElement):
 			return _(item.attrib["text"]), eval(configElement)
 	return "", None
 
-class SetupError(Exception):
-	def __init__(self, message):
-		self.msg = message
-
-	def __str__(self):
-		return self.msg
 
 class SetupSummary(Screen):
 
@@ -151,7 +145,12 @@ class Setup(ConfigListScreen, Screen):
 
 def getSetupTitle(id):
 	xmldata = setupdom.getroot()
+	id = str(id)
+	title = ""
 	for x in xmldata.findall("setup"):
 		if x.get("key") == id:
-			return x.get("title", "").encode("UTF-8")
-	raise SetupError("unknown setup id '%s'!" % repr(id))
+			title = x.get("title", "").encode("UTF-8")
+	if title == "":
+		print "[Setup] Error: Setup ID '%s' not found in setup file!" % id
+		title = "** Setup error: '%s' section not found! **" % id
+	return _(title)
--- a/lib/python/Screens/Setup.py
+++ b/lib/python/Screens/Setup.py
@@ -151,6 +151,7 @@ def getSetupTitle(id):
 		if x.get("key") == id:
 			title = x.get("title", "").encode("UTF-8")
 			if title == "":
+				print "[Setup] Error: Setup ID '%s' title is missing or blank!" % id
 				title = "** Setup error: '%s' title is missing or blank!" % id
 	if title == "":
 		print "[Setup] Error: Setup ID '%s' not found in setup file!" % id
--- a/lib/python/Screens/Setup.py
+++ b/lib/python/Screens/Setup.py
@@ -151,7 +151,7 @@ def getSetupTitle(id):
 		if x.get("key") == id:
 			title = x.get("title", "").encode("UTF-8")
 			if title == "":
-				title = "** Setup warning: '%s' title is missing or blank!" % id
+				title = "** Setup error: '%s' title is missing or blank!" % id
 	if title == "":
 		print "[Setup] Error: Setup ID '%s' not found in setup file!" % id
 		title = "** Setup error: '%s' section not found! **" % id
--- a/lib/python/Screens/Setup.py
+++ b/lib/python/Screens/Setup.py
@@ -150,6 +150,8 @@ def getSetupTitle(id):
 	for x in xmldata.findall("setup"):
 		if x.get("key") == id:
 			title = x.get("title", "").encode("UTF-8")
+			if title == "":
+				title = "** Setup warning: '%s' title is missing or blank!" % id
 	if title == "":
 		print "[Setup] Error: Setup ID '%s' not found in setup file!" % id
 		title = "** Setup error: '%s' section not found! **" % id


IanSav:
I accidentally closed the previously pending Pull Request #1342 .  This is the original pull request together with the change suggested /  requested by Littlesat.

IanSav:
@Littlesat: Based on your suggestion I added an error message for a setup tag that has no title or an empty title.  Is this what you want or would you prefer that the title simply default to "Settings" without any further fuss?

Huevos:
Still no proper reason for deleting this needed exception. If any .py file is broken there is an exception. This forced exception does the same for setup.xml.

I suspect you want to delete this because it is causing a problem in some other code you are writing,

IanSav:
The exception crash is not needed.  No program code is broken.  Any potential non fatal issue with a data file that can be recovered should not cause a crash.  In the new code any such issues with setup.xml are reported appropriately for follow up action.

Your suspicion is incorrect.  I am not having any setup.xml issues.  I have no problems with any of the code I am developing.  My protocol is to verify and test all my code, to the best of my abilities, before making any pull requests.

When I see examples of what I believe to be poor code I think it is in the community's interest to suggest corrections and/or improvements.  Isn't that the point of open source projects?  (This is particularly the case when some OpenPLi members complain that not many outsiders contribute back code to the OpenPLi code base.)

Huevos:
By the way an empty string is not a translation error in OpenPLi.

IanSav:
There are no translations of empty strings.  What prompted the translation information?  Have you found a problem?

Huevos:
You did in paragraph 2 of your proposal. And just adding a random word doesn't make any sense.

littlesat:
Why not simply do this by changing just one line and remove the if title="" stuff completely... (and of course the raise...)

title = x.get("title", "** Setup warning: '%s' title is missing or blank!" % id).encode("UTF-8")

And then show this as title....???

IanSav:
You are absolutely correct in what you suggest.  Unfortunately you have highlighted that there is a line missing from my commit.  :(  There is also meant to be a print statement in that if block that logs the error and makes the current structure make sense.

Pull request updated.

Huevos:
@Littlesat, why would you remove the raise? It is there so a broken setup.xml gets found immediately.

IanSav:
It should be found by the developer making the change when they test their changes.  (Developers are testing their changes - I hope!)  Users should not be exposed to a crash to address developer laziness or error conditions that slip through any testing.  The friendly error messages are a more user friendly way to go.

littlesat:
Or show a warning or raise... when we do a raise we can close this case

IanSav:
The new code shows a warning to the user and logs the error.  I don't understand what you are asking?

Huevos:
Like he says raise should be left in the code. Without the raise there will not be a BSoD when there should be one.

The whole point of the raise is immediate feedback. 

And no idea why you want to remove the raise. A normal user with a production image will never experience it. It is there to help developers. Why do you want to make things more difficult for developers?

IanSav:
That is not what I understand by the post.  It asks for a warning (as I provided in this code update) *OR* a raise (that I propose is too harsh and unwarranted and is in the current code).  If Littlesat only wanted a raise then the code would not be changed from the original and the request would/should be declined.  This is why I asked for clarification from Littlesat.

This condition is not a fatal error, it should not trigger a crash.  The system is fully capable of continuing without any implication from this condition other than the broken menu item will be unavailable.  The cause of this issue should be picked up during development and code testing.  End users should not penalised.

littlesat:
The point is that actually the menu.xml and setup.xml is somehow embedded in the code... instead of xml it is actually easier and more readable to make the table in python... so not having title is not appropriage and normal Users usually do not adapt the xml... so doing the raise as it is sounds reasonable...

littlesat:
When users do edit the file many times then it is require to adapt as the raise is reasobable

IanSav:
Sorry Littlesat I am not sure I understand what you have said but if you are saying that crashing the UI is an appropriate action for a minor error that can easily be corrected in the code then we will have to agree to disagree. 

I have discussed this situation with peer programmers, all with very MANY years of programming experience, like myself, and we are all astounded that crashing the UI is the current and preferred choice.  This is an error condition that shouldn't be allowed to get into production code.  Should an error slip through testing is should be handled and not crash.

I assume that this pull request is to be rejected.  I try to create code that is written for quality, performance and usability.  I will not write or submit code that is unprofessional or contradicts my conscience.  Is there any point in my continuing to develop code for OpenPLi?  If my contributions are going treated like this and be rejected for no technically justifiable reason I will direct my efforts elsewhere.

I read and noted that some OpenPLi members complain that the Enigma2 community don't contribute back to OpenPLi.  If this is how external contributions and corrections are handled then no wonder people don't share code.  :(

IanSav:
By the way, can *anyone* here give me a technical and professional reason why a small, correctable error, should cause a system crash for end users?

(Please don't me that this is to help find bugs.  End users are not developers and should not be exposed to such issues.  Displaying an error message and continuing is a more appropriate approach.)

Huevos:
Ian, end users are not exposed to the raise. It never happens in a production image. So there is no argument that it might upset end users.

IanSav:
The "raise" is in the production code, as such it can be encountered.  I fell over it when experimenting and lost some active recordings.  That is one of the main reasons why I pushed to correct this serious miss-feature.  I would be much happier to see a clear and concise error message to explain the problem.  That is why I wrote this correction.

I am obviously from a different generation of programmers.  I like my code to run without crashing no matter what the provocation.  User experience is very important to me.

If this small change has created so much trauma I wonder and doubt that any of my other changes will be acceptable.  The next change was to take processing of Setup menus from about 500 msec per menu to about 100 msec  per menu while adding significant new features.  I also took the menu generation of setup menu entries from about 50 msec per lookup to about 55 msec for the first lookup and then 1 msec for ALL subsequent lookups.  Is it worth proceeding or am I wasting my time?

Huevos:
So, the reason you want the raise removed is because you are using your regular domestic receiver as a developer box at the same time? And this is true how do you manage to do restarts when testing Python code?

RobvanderDoes:
I get the very strong impression that there is some misunderstanding about what a 'raise' actually does.

IanSav:
I schedule dangerous testing and reboots between recording sessions.

I am an unpaid volunteer, I don't have money to buys spare units for all the development and testing platforms.

Huevos:
A "raise" is there to re-raise the last suppressed exception, or to deliberately invoke a custom exception where one may not already exist.

Huevos:
> I am an unpaid volunteer, 

Aren't we all?

technic:
This error is not due to the incorrect user input (or other actions from the outside), but because of broken xml file which is part of the code base. So, in general, crashing in such situations is a valid choice, and it is becoming one of "matter of taste" discussions. Thus, maintainers' taste should be accepted.

In addition, I think using `assert` statement is the right choice in such situations, although they get removed when we run python with -O flag. How good does it work in enigma2 code then?

technic:
If there is an easy way to trigger python debug switch,  maybe we all will agree on something like:
```python
if __debug__:
  raise Exception("Fix your xml dude!")
else:
  return "Broken Title"
```

IanSav:
>> I am an unpaid volunteer,

> Aren't we all?

I wish I knew how some of you have acquired so much gear without direct income from this hobby.  I would welcome more test and development boxes.

IanSav:
I agree with most of what Technic posted.  Though I still hold that in this particular situation there is no additional benefit from crashing the UI for a minor and programatically correctable issue.

The proposed code clearly shows that the menu entry is in error and provides the menu key for the faulty entry.  The coder should immediately note the error on their first test run with their changes.  In this case no additional information or action opportunity is gained by crashing the UI and there is no risk or issue allowing the UI to continue to run.

Is the `__debug__` syntax supported in Enigma2 development?  That could be an acceptable compromise.

Huevos:
I can't believe you are trying to do development and recordings at the same time. You can't have your cake AND eat it. If you are a Beyonwiz developer don't they supply you with any hardware?

littlesat:
Thanks for the merge suggestion... But I think we simply should keep the raise...

PR #1353, opened 2018-03-17
Add PyLong_FromUnsignedLongLong ePyObject wrapper

technic:
When I was compiling enigma2 on ubuntu, this function was missing. I don't remember the reason, although it's good to have it for completeness. 

Generated by copy & paste, so please look carefully for typos.


littlesat:
Look carefully for typos sounds it is not really tested by yourself.... So I suggest we should not merge this right-a-way....

PR #1356, opened 2018-03-17
fix RGB structure size on 64bit

technic:
We shall stay with platform independent types for color data.


athoik (approved):
I think using types defined on stdint header is the right way to go!

eriksl:
It should have been this way from the beginning, actually...

PR #1358, opened 2018-03-17
Patch reactor

technic:
This needs more discussion. Just a copy of explanation from the commit message:

## Fixed bugs:

In processOneEvent(), when `user_timeout > 0` but `poll_timeout == -1`
(because there were no timers for example)
poll was going to infinite wait, but it should abort after `user_timeout`.

In iterate `to` variable could become 0 due to rounding.
In this case processOneEvent was getting argument `user_timeout=0`,
that was treated as infinite poll waiting time.
Now we pass singed integer to processOneEvent
and zero argument means exiting immediately

## TODO:

eMainLoop::poll() interprets zero timeout as infinite wait.
I think twisted reactor use None for that purpose and 0 means no waiting

### DIFF
--- a/lib/base/ebase.cpp
+++ b/lib/base/ebase.cpp
@@ -170,7 +170,7 @@ void eMainloop::removeSocketNotifier(eSocketNotifier *sn)
 	eFatal("[eMainloop::removeSocketNotifier] removed socket notifier which is not present, fd=%d", fd);
 }
 
-int eMainloop::processOneEvent(unsigned int twisted_timeout, PyObject **res, ePyObject additional)
+int eMainloop::processOneEvent(long user_timeout, PyObject **res, ePyObject additional)
 {
 	int return_reason = 0;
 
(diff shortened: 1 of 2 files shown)


eriksl:
Don't use any of:
- long
- long int
- long long int
It's semantics are varying amongt architectures.

Use either int (if the exact size isn't important) or one of (u)int(8|16|32|64)_t

technic:
In the initial `poll` method we read argument from python with `PyInt_AsLong(timeout)`, that was the reason I used `long`.

eriksl:
Yes that is another league of problems... Is there any documentation as whether to which exact "C" int type the "Long" type of Python maps? If it's really "long", we're screwed as soon as we're moving to 64 bit. I hope it can be defined as "int" or "int32_t".

technic:
Yes. Python has two integer types:

* *PyIntObject* is implemented via "C" `long` and hence should be casted to it.
* *PyLongObject* is implemented via long arithmetic and can have infinite size.

eriksl:
That's not the answer to my question. What exact C types are both Python objects mapped to...

technic:
That's what I have written above: `long` and something more complicated like `long[]` or so. I think details of the latter type doesn't matter here.

PR #1366, opened 2018-03-18
EventView: use StaticText for buttons

Taapat:
This allow hide unused buttons using the objectTypes attribute in skin.


nautilus7:
This change will break all skins.

Taapat:
Yes, break all skins that do not display the buttons correctly.
Now, they show empty buttons, without text, which do not work.

I would recommend introducing dynamic buttons and not worrying about changes in the buttons: https://github.com/littlesat/skin-PLiHD/blob/master/usr/share/enigma2/PLi-HD/skin_templates.xml#L153

PR #1367, opened 2018-03-18
EpgSelection: use StaticText for buttons

Taapat:
This allow hide unused buttons.


nautilus7:
This change will break every skin.

This have been discussed in the openpli forum, and most people agreed that we should keep the changed screens to the minimum in order not to break every third party skin.

I personally don't have any problem, and my opinion is to change all colored buttons to staticText().

Taapat:
As I wrote in the previous pull request.
It will break all skins that do not display the buttons correctly, and I would recommend introducing dynamic buttons and not worrying about changes in the buttons: https://github.com/littlesat/skin-PLiHD/blob/master/usr/share/enigma2/PLi-HD/skin_templates.xml#L153

I think that in screens, where the buttons are not enabled and disabled, you can leave everything as it is.

Taapat:
Or it would be better to implement getBoolean for VariableText to use converter ConditionalShowHide.
Then the buttons can also be unchanged.
I will check this later.

PR #1368, opened 2018-03-18
[Ci] add option "Relevant PIDs routing"

Dima73:


littlesat:
Please explain what this is.
And again there are more changes the. Described... Yes -> yes for example...

littlesat:
And why the open close on the proc, with the try except while you can check the file exists... and why the hasattr is suddenly needed for the highspeed?

Dima73:
config.cimisc.dvbCiDelay  is global option.
It must be initialized only once.
Before it is done 4 times.
Now it is used hasattr.

littlesat:
Is this for the latest vu changes?

Dima73:
Yes

littlesat:
See my comments on the forum

Dima73:
We need another patch.

littlesat:
Please keep it open so we don't forget it....

littlesat:
Merged but with a new patch....

PR #1385, opened 2018-03-26
add hdrtype option for support boxes, remove videomode Ypbpr for boxes which aren't equipt with the YPbPr connector

zgemma-star:


pieterg:
I will cherry-pick the first two commits.
The third commit however needs some work; the e2 aspect ratio settings have turned into a real mess over the last few years. For instance, the 'auto' setting (when found in policy_choices) is not used at all, instead, the 'bestfit' option is written when 'auto' is selected. That needs to be fixed. And the 'scale' option is mapped to 'bestfit' as well. While the translator comments says that this 'scale' option might break the aspect ratio, which would suggest it was intended to do a similar thing as your 'full' (full screen) setting.

BTW, this commit duplicates code and settings, something we would like to avoid.

Enough reasons for a rewrite, I'll see how I can squeeze in your 'full' option with the rewrite

PR #1388, opened 2018-03-26
add video composite mode, set colorformat standard on cvbs, show only the av input which the hw support

zgemma-star:


littlesat:
As far I can see these patches are already in....

PR #1392, opened 2018-03-27
[ConfigList] add sanity check  help_window.instance

Dima73:
  File
"/usr/lib/enigma2/python/Components/ConfigList.py", line 74, in
selectionChanged
  File
"/usr/lib/enigma2/python/Components/ConfigList.py", line 196, in
handleInputHelpers
AttributeError: 'NoneType' object has no attribute
'instance'


Huevos:
Sanity check is a good idea. But this is the cause of the crash...
`if isinstance(self["config"].getCurrent()[1], ConfigText)`

Reason for the crash is subclasses of ConfigText such as ConfigNumber do not have the help window instance. But using "isinstance" includes all the subclasses of ConfigText.

PR #1404, opened 2018-03-29
[Makefile] include by filetype

AbuBaniaz:


AbuBaniaz:
Will submit different request

PR #1408, opened 2018-03-31
[config.py] Fix __getattr_ method error code

IanSav:
NOTE: This pull request should not be merged until a build of OpenWebif built on or after 31-Mar-2018 has been incorporated.  (OpenWebif was incorrectly checking for a KeyError rather than the correct AttributeError.)

The `__getattr__` method in ConfigSubsection did not check if the attribute being retrieved existed.  This caused a KeyError to be raised instead of the expected AttributeError.  This issue means that the getattr() function could not correctly test for the existence of a config item without causing a crash.  This fix will allow the method to return the attribute if it exists or raise an AttributeError.  The AttributeError allows the getattr() function to be supplied with a default that can be returned to the calling code.

For example: getattr(config.usage, "fred", None) would crash rather than return the expected None if config.usage.fred does not exist.  This change fixes the crash and allows getattr() to function as expected.  That is, the definition of "fred" will be returned if it exists or else None will be returned.

Also fixed the last line of the file to have a newline.

### DIFF
--- a/lib/python/Components/config.py
+++ b/lib/python/Components/config.py
@@ -1660,7 +1660,9 @@ class ConfigSubsection(object):
 			value.load()
 
 	def __getattr__(self, name):
-		return self.content.items[name]
+		if name in self.content.items:
+			return self.content.items[name]
+		raise AttributeError(name)
 
 	def getSavedValue(self):
 		res = self.content.stored_values


Huevos:
Ian, OpenPLi builds OWI directly from E2OpenPlugins git so now that PR is accepted by OWI it is already in PLi. IMS says he tested and works for him.

IanSav:
@littlesat I believe this pull request can now be processed.

IanSav:
Is there any reason why this pull request hasn't been processed?

littlesat:
Are you sure we can 'revert this revert' right now? Where is the OWIF "contra" patch...?

Huevos:
Owif was fixed.

IanSav:
@littlesat these are the merged OpenWebif pull requests that address the original problem:

https://github.com/E2OpenPlugins/e2openplugin-OpenWebif/pull/786
https://github.com/E2OpenPlugins/e2openplugin-OpenWebif/pull/787
https://github.com/E2OpenPlugins/e2openplugin-OpenWebif/pull/788
https://github.com/E2OpenPlugins/e2openplugin-OpenWebif/pull/789

I believe it is okay to now merge this pull request.

PR #1415, opened 2018-03-31
ConfigList: use VirtualKeyboard only for ConfigText and ConfigPassword config entries

Taapat:
isinstance return true also if classinfo argument is subclass. Therefore now, after this commit https://github.com/OpenPLi/enigma2/commit/bab01517e80e26dc36d3d42a47b3a0d447f10d6d key_text is displayed and it is possible to call VirtualKeyboard on all configs which has a ConfigText as subclass.
But for example using VirtualKeyboard in ConfigNumber causes a GSOD.
Therefore instead of isinstance use __class__.__name__ to check config entry class name and use VirtualKeyboard only for ConfigText and ConfigPassword config entries.
Also remove try exceptions in HideHelp and ShowHelp. Add aditional check for help_window, this shold fix GSOD and I hope that try exception not need after that.

### DIFF
--- a/lib/python/Components/ConfigList.py
+++ b/lib/python/Components/ConfigList.py
@@ -212,11 +212,11 @@ class ConfigListScreen:
 		self.session.openWithCallback(self.VirtualKeyBoardCallback, VirtualKeyBoard, title = self["config"].getCurrent()[0], text = self["config"].getCurrent()[1].getValue())
 
 	def HideHelp(self):
-		if (self["config"].getCurrent()[1].__class__.__name__ == 'ConfigText' or self["config"].getCurrent()[1].__class__.__name__ == 'ConfigPassword') and self["config"].getCurrent()[1].help_window and self["config"].getCurrent()[1].help_window.instance is not None:
+		if "config" in self and (self["config"].getCurrent()[1].__class__.__name__ == 'ConfigText' or self["config"].getCurrent()[1].__class__.__name__ == 'ConfigPassword') and self["config"].getCurrent()[1].help_window and self["config"].getCurrent()[1].help_window.instance is not None:
 			self["config"].getCurrent()[1].help_window.hide()
 
 	def ShowHelp(self):
-		if (self["config"].getCurrent()[1].__class__.__name__ == 'ConfigText' or self["config"].getCurrent()[1].__class__.__name__ == 'ConfigPassword') and self["config"].getCurrent()[1].help_window and self["config"].getCurrent()[1].help_window.instance is not None:
+		if "config" in self and (self["config"].getCurrent()[1].__class__.__name__ == 'ConfigText' or self["config"].getCurrent()[1].__class__.__name__ == 'ConfigPassword') and self["config"].getCurrent()[1].help_window and self["config"].getCurrent()[1].help_window.instance is not None:
 			self["config"].getCurrent()[1].help_window.show()
 
 	def VirtualKeyBoardCallback(self, callback = None):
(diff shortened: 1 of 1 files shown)


Huevos:
Nice suggestion.

Huevos:
In what situation can "config" not be in "self"?

Taapat:
Check explanation in the second commit.
For example in plugin NetworkWizard.

Huevos:
I still don't understand how "config" cannot be in "self". It is set in the "__init__" on line 159.

Please can you provide a step-by-step of how to reproduce the GSoD.

Taapat:
Are you really writing about plugin NetworkWizard form enigma **OpenPLi**?
https://github.com/OpenPLi/enigma2/blob/rc/lib/python/Plugins/SystemPlugins/NetworkWizard/NetworkWizard.py

Install plugin, start wizard from Network setup.
In the first window, press Ok, in the second press exit, and you will get:
```
  File "/usr/lib/enigma2/python/Components/ActionMap.py", line 65, in action
    return ActionMap.action(self, contexts, action)
  File "/usr/lib/enigma2/python/Components/ActionMap.py", line 46, in action
    res = self.actions[action]()
  File "/usr/lib/enigma2/python/Plugins/SystemPlugins/NetworkWizard/NetworkWizard.py", line 84, in back
    WizardLanguage.back(self)
  File "/usr/lib/enigma2/python/Screens/Wizard.py", line 291, in back
    self.updateValues()
  File "/usr/lib/enigma2/python/Screens/Wizard.py", line 486, in updateValues
    self.configInstance.doClose()
  File "/usr/lib/enigma2/python/Screens/Screen.py", line 111, in doClose
    x()
  File "/usr/lib/enigma2/python/Components/ConfigList.py", line 217, in HideHelp
    if (self["config"].getCurrent()[1].__class__.__name__ == 'ConfigText' or self["config"].getCurrent()[1].__class__.__name__ == 'ConfigPassword') and self["config"].getCurrent()[1].help_window and self["config"].getCurrent()[1].help_window.instance is not None:
KeyError: 'config'
```

As I see in Wizard on close in updateValues "config" from configInstance is removed there: https://github.com/OpenPLi/enigma2/blob/rc/lib/python/Screens/Wizard.py#L485
But after that in next line on configInstance.doClose() is checked "config" value in ConfigList HideHelp.

Huevos:
Ok, it crashes, but I still don't understand why, when self["config"] is assigned on line 159.

Taapat:
I will write the second time:
> Are you really writing about plugin NetworkWizard form enigma OpenPLi?

Show the link to the code where self["config"] is assigned on line 159.

"config" is assigned on Screen Wizard init in line 183: https://github.com/OpenPLi/enigma2/blob/rc/lib/python/Screens/Wizard.py#L183

On close "config" is deleted in Wizard line 485: https://github.com/OpenPLi/enigma2/blob/rc/lib/python/Screens/Wizard.py#L485
But after that is called self.configInstance.doClose() in line 486: https://github.com/OpenPLi/enigma2/blob/rc/lib/python/Screens/Wizard.py#L486
which check now deleted and not existed "config" value in ConfigList HideHelp.

Huevos:
https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Components/ConfigList.py#L161

Ok, so it is a fix in enigma code for a plugin that deletes a variable. Now I understand.

Taapat:
Actually not plugin delete a variable, but enigma in Wizard screen delete variable which is used in the screen ConfigList.

Huevos:
Anyway, obviously necessary to stop the crash.

Thanks for explaining.

Huevos:
BTW... maybe shorten to this.
`self["config"].getCurrent()[1].__class__.__name__ in ('ConfigText', 'ConfigPassword')`

Taapat:
Yes, I also think that it would be better.

WanWizard:
Just leave the current fix in rc, and make improvements in develop please.

PR #1431, opened 2018-04-08
Giga4K MultiBoot - Add Giga4K MultiBoot capability mainly in Tools/Mu…

TwolDE2:
…ltiboot.py with minimal change to FlashImage/SystemInfo

Tested on hd51, gbue4K, gbquad4k with full and empty slots  
Code supports both Flash Image & Multiboot (choose Image) restart.
99% of changes in Tools/Multiboot.py allowing transparent calls from other functions


littlesat:
The feature is multiboot, not multibootgb and the otters ;)... but we can do different stuff depending on other parameters (procs/sys/content of file...)

littlesat:
And how do you handle slots that are empty!

TwolDE:
Littlesat, I am sure with your Python competence its obvious..... either the slot contains the enigma image name (pli, atv or vix) or it says empty slot.... setup in tools/multiboot.py
To be honest I know your python skills are much better than mine, so it should be easily within your abilities to modify this to suit OpenPli.
I was one of the 1st users of the HD51, but taking them as the only way of implementation when thanks to your tools/multiboot idea, it is so easy to keep almost all the changes inside one component, is being a little bit defensive.

TwolDE:
As far as its different, its exactly the same except for additions in the calls to tools/multiboot ... I really don‘t care my code is different anyway, I just tried to meet your requirements, but obviously failed.
Your loos not mine.

littlesat:
What I prefer to see is actually a BSP layer change so we have all possible thinks we could have as startup... and we simply can choose between the option... for HD we have 8 options and for gb 3...

littlesat:
And I’m convinced you did your best... please discuss this further on the forum... your commit works... but actually the multiboot we have now is also not good... we need somewhere all options and modi../

littlesat:
My suggestions.... But actually first thanks for contributing....

1. I like the way you did the empty slot.... :)

2. Let if SystemInfo["canMultiBoot"] be False or a tuple with the addin and lastslot number...
SystemInfo["canMultiBoot"] = HardwareInfo().get_device_model() in ('hd51', 'h7', 'vs1500') and (1, 4) or  HardwareInfo().get_device_model() in ('gbue4k', 'gbquad4k') and (3, 3)

3. Does GB has this sys ? int(open('/sys/firmware/devicetree/base/chosen/kerneldev', 'r').read().replace('\0', '')[-1]) then do the desician GB/HD on this sys... of find another way to decide which STARTUP file "syntax" we need to put there - Or check it by verifying what is in the STARTUP file to determine what to 'write' in STARTUP.... Maybe it is not bad to read-check-modify it on existing file. The current image can also be derived from the existing file.... (then we also do not need point 4)

4. Return int(open('/sys/firmware/devicetree/base/chosen/bootargs', 'r').read().replace('\0', '').split('=')[1].split('p')[1].split(' ')[0]) - 3 /2 instead of having 4 lines (and the same there find a different way to detect gb by sys or procs)

TwolDE2:
2. Love the idea of the tuple in SystemInfo (just didn't know you could do it!!)
4. again, need somebody who really knows python to clean up my code - fine by me.
3. Unfortunately no kerneldev only bootargs (and I have searched I believe all the files to find a better way. e.g. from my gbue4k using slot 3 currently for OpenPli
root=/dev/mmcblk0p9 rootwait rw rootflags=data=journal libata.force=1:3.0G,2:3.0G,3:3.0G coherent_poll=2M vmalloc=525m bmem=529m@491m bmem=608m@2464m\00

So better to work out a nice clean way of using the STARTUP ... it is currently pretty messy on all images. especially with the HD51 complication of having 2 modes (so does the SF4008, but that although promised doesn't have multiboot)

TwolDE2:
Can we use (HD,1,4) and (GB,3,3) as the tuple, so we don't have to worry about detecting box type further, it's all in SystemInfo

TwolDE2:
Using ("HD",1,4) and ("GB",3,3) as the tuple, seems to work well for me without issue.
Have only tested on the HD51, but initial testing is OK.... makes life (and the code) so much easier

TwolDE2:
Using ("HD",1,4) and ("GB",3,3) as the tuple, seems to work well for me without issue.
Have only tested on the HD51, but initial testing is OK.... makes life (and the code) so much easier

TwolDE2:
oops double entry ..... OK, have now tested all your suggestions & changes on hd51 & gbue4k and it all seems to work and ... faster

littlesat:
I want to get rid of GB/HD as this is some kind of 'if GB do X else if HD do something else...' outside SystemInfo.py. Having two different systeminfo keys for one and the same feature is also not nice... Then better when building the STARTUP 'parse' from the previous file with what we should overwrite it...

Thanks for your patience.... I hope I can help you to create better python code... ;)

littlesat:
And actually it will not be faster... only the code looks a bit cleaner...

Probably you can read the STARTUP file and when 'rootflags=data=journal' is inside that file write GB style, otherwise HB style....

Somethin like....
if 'rootflags=data=journal' in open('/usr/media/mmcblk0p0/STARTUP').read():
write GB stuff
else
write HB stuff

littlesat:
And I would do

self.imagelist[self.slot] =  { 'imagename': _("Empty slot")}

to make it translatable and also do it twice not add the extra variable... (but the extra variable is 'taste'.... when doing it then define emptyslot = { 'imagename': _("Empty slot")} and use that so not only include the string)

littlesat:
In short what you now have as SystemInfo["canMultiBootGB"] could be 'somehow replaced by 

if 'rootflags=data=journal' in open('/usr/media/mmcblk0p0/STARTUP').read():

But you need of course always check of SystemInfo["canMultiBoot"] has content

b.t.w. we do in on more places with system info to give it none or values...

TwolDE2:
Most of the suggestions are in plus some little changes and a major fix (current image slot).
The GB selection is done via /dev/mmcblk0p1

littlesat:
Good to know is that also a setup file there... then with os.path.exists or so we can check for which file we need to write for the selection ;)

littlesat:
I cannot merge this one.... It looks almost perfect!!!! 

+	if SystemInfo["canMultiBoot"] and 'rootflags=data=journal' in open('/dev/mmcblk0p1').read():
 +		return (int(open('/sys/firmware/devicetree/base/chosen/bootargs', 'r').read().replace('\0', '').split('=')[1].split('p')[1].split(' ')[0])-3)/2
 +	elif SystemInfo["canMultiBoot"]:

->

if SystemInfo["canMultiBoot"]:
    if 'rootflags=data=journal' in open('/dev/mmcblk0p1').read():
        return (int(open('/sys/firmware/devicetree/base/chosen/bootargs', 'r').read().replace('\0', '').split('=')[1].split('p')[1].split(' ')[0])-3)/2
    else

littlesat:
maybe I close this one and then you push a new merge request... In addition the small suggestion I can also made for you...
Perfect cooperation Thanks!

littlesat:
And one other suggestion

SystemInfo["canMultiBoot"][1] == 4

Sheck here also for the contect of the STARTUP file...

littlesat:
I see why all changes failed... because I did not commit the first thing in the first place... Please reset your git rep and commit the last changes again an do a new merge request..

TwolDE:
OK, I will sort this out over the next day - I shortly have to go out to see my Osteopath (Bad back and left knee!!), and I also want to run a clean build to checkout first. 
1. SystemInfo["canMultiBoot"][1] == 4 agree was just being lazy (knew this worked) but was busy fixing couldn't get current slot to show in multiboot reboot (has to check != 12, as Giga is mode 0)
2. will also ,make other change, test and then resubmit

littlesat:
Oeps... now it really ends up in a mess ;)

TwolDE:
Soory, showing my ignorance!
Which is the best way to proceed.
I can delete this forked version, refork........ what do you suggest...this was my 1st pull and obviously blew it!

TwolDE:
I as just going to revert and add the complete changes again

PR #1437, opened 2018-04-09
ConfigList: try fix GSOD on ShowHelp and HideHelp

Taapat:
Create functions onDeselect and onSelect in ConfigList to use it when call or return from VirtualKeyBoardCallback or MessageBox.
This is the realization of this Huevos idea: https://github.com/Taapat/enigma2/commit/bfb0b008115677c0a50c408b1b3373dde2002877 but in a safer way, and I hope it does not call GSOD.
Uses standard functions that are used in selectionChanged, and therefore they should work not only for HelpWindow, but also for other possible features.

### DIFF
--- a/lib/python/Components/ConfigList.py
+++ b/lib/python/Components/ConfigList.py
@@ -73,6 +73,14 @@ class ConfigList(GUIComponent, object):
 		for x in self.onSelectionChanged:
 			x()
 
+	def onDeselect(self):
+		if isinstance(self.current, tuple) and len(self.current) >= 2:
+			self.current[1].onDeselect(self.session)
+
+	def onSelect(self):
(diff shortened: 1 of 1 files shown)


littlesat:
and I hope it does not call GSOD -> So it can cause GSOD?

zgemma-star:
This commit fixes GSOD on first install wizard and autodiseqc which appeared by the follow commit https://github.com/OpenPLi/enigma2/commit/f9666801efb26e7d25abadcbe875f9f7afd679d5

littlesat:
Sorry... someone who has access to R/C should merge it...

zgemma-star:
I think it was better to push the commit to the develop branch and then cherry pick to the rc branch.

PR #1439, opened 2018-04-11
remove the refresh rate wizard when videomodus is rca

zgemma-star:
all rca modes have static refresh rate so no point to set it
it cannot switch to modus like pal60 (480i)

### DIFF
--- a/lib/python/Plugins/SystemPlugins/Videomode/videowizard.xml
+++ b/lib/python/Plugins/SystemPlugins/Videomode/videowizard.xml
@@ -22,7 +22,7 @@ self["portpic"].hide()
 	</step>
 	<step id="rateselection" timeout="20" timeoutaction="selectnext">
 		<condition>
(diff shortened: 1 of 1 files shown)


lib/python/Plugins/SystemPlugins/Videomode/videowizard.xml
```diff
@@ -22,7 +22,7 @@ self["portpic"].hide()
 	</step>
 	<step id="rateselection" timeout="20" timeoutaction="selectnext">
 		<condition>
-self.condition = (self.port != "DVI" or self.mode == "PC")
+self.condition = (self.port != "DVI" and not "RCA" or self.mode == "PC")
```
athoik, line 25: not "RCA" what is not "RCA"? 


```
>>> not "RCA"
False
```

Do you mean something like: `and not self.port == "RCA"` ?

zgemma-star:
yes, your right I will adjust it

PR #1446, opened 2018-04-12
Giga4K Multiboot - 2nd attempt

TwolDE2:
Hopefully this will work through:-
    Adds multiboot support for gbue4k gbquad4k   
    Adds gbue4k to rc_models.cfg
    Changes a few descriptions in FlashImage


TwolDE2:
Checked out this code on both hd51 & gbue4k with new builds - so hopefully everything covered this time.

lib/python/Tools/Multiboot.py
```diff
@@ -3,17 +3,24 @@
 import os
 
 def GetCurrentImage():
-	return SystemInfo["canMultiBoot"] and int(open('/sys/firmware/devicetree/base/chosen/kerneldev', 'r').read().replace('\0', '')[-1])
+	if SystemInfo["canMultiBoot"]:
+		if 'rootflags=data=journal' in open('/dev/mmcblk0p1').read():
```
zgemma-star: The setting 'rootflags=data=journal' is I think to generic settings and there is a possibility it will been add to other boxes.
Maybe it's easier to use the option which are available I assume GB uses Mode5 and so not Mode12 (as Mode12 is only used by BCM7251s)

if SystemInfo["canMode12"]:
       return	int(open('/sys/firmware/devicetree/base/chosen/kerneldev', 'r').read().replace('\0', '')[-1])
else:
      return (int(open('/sys/firmware/devicetree/base/chosen/bootargs', 'r').read().replace('\0', '').split('=')[1].split('p')[1].split(' ')[0])-3)/2

TwolDE2: Really, I am easy about this. It works for the moment, but there are numerous files you can select on that "may" be used later (by  the hd51/zgemma boxes) but are not at the moment (or visa versa ... like kerneldev, which is not in the Giga proc)).... or I revert to just specifying up front the boxtype in SystemInfo where we detect the boxtype (which Littlesat didn't like) ... really it is up to Littlesat and you guys to decide.... but there are other non currently  Pli supported boxes that have some differences in multiboot.
hd51 may have been 1st, but unfortunately vendors have always done their own thing.

lib/python/Tools/Multiboot.py
```diff
@@ -3,17 +3,24 @@
...
+			return (int(open('/sys/firmware/devicetree/base/chosen/bootargs', 'r').read().replace('\0', '').split('=')[1].split('p')[1].split(' ')[0])-3)/2
+		else:
+			return	int(open('/sys/firmware/devicetree/base/chosen/kerneldev', 'r').read().replace('\0', '')[-1])
 
 def GetCurrentImageMode():
-	return SystemInfo["canMultiBoot"] and int(open('/sys/firmware/devicetree/base/chosen/bootargs', 'r').read().replace('\0', '').split('=')[-1])
+	if SystemInfo["canMultiBoot"] and 'rootflags=data=journal' not in open('/dev/mmcblk0p1').read():
```
zgemma-star: if SystemInfo["canMultiBoot"] and SystemInfo["canMode12"]:

lib/python/Screens/FlashImage.py
```diff
@@ -397,17 +392,19 @@ def getImagelistCallback(self, imagesdict):
...
+				list.append(ChoiceEntryComponent('',((_("slot%s - %s current mode 12 (current image)") if x == currentimageslot and mode == 12 else _("slot%s - %s  restart mode 12 ")) % (x, imagesdict[x]['imagename']), x + 12)))
 		self["list"].setList(list)
 
 	def keyOk(self):
 		currentSelected = self["list"].l.getCurrentSelection()
 		slot = currentSelected[0][1]
 		if currentSelected[0][1] != "Waiter":
 			model = HardwareInfo().get_device_model()
-			if slot < 12:
+			if SystemInfo["canMultiBoot"] and 'rootflags=data=journal' in open('/dev/mmcblk0p1').read():
```
zgemma-star: if not SystemInfo["canMode12"]:

lib/python/Screens/FlashImage.py
```diff
@@ -397,17 +392,19 @@ def getImagelistCallback(self, imagesdict):
...
+				startupFileContents = "boot emmcflash0.kernel%s 'root=/dev/mmcblk0p%s rootwait rw rootflags=data=journal libata.force=1:3.0G,2:3.0G,3:3.0G coherent_poll=2M brcm_cma=764M@0x10000000 brcm_cma=1024M@0x80000000'\n" % (slot, slot * 2 + 3)
```
zgemma-star: This shouldn't belong in e2 but in the bsplayer defconfig as cmdline, it's not realistic to assume all machines uses the memory layout of this specific box.

rootflags=data=journal libata.force=1:3.0G,2:3.0G,3:3.0G coherent_poll=2M brcm_cma=764M@0x10000000 brcm_cma=1024M@0x80000000'

lib/python/Tools/Multiboot.py
TwolDE: Well the GB doesn‘t specifiy a mode with the BCM7252s, only the BCM7251s receivers like hd51, SF4008 and the zgemma‘s specify a dual mode setting and for instance the SF4008 utilises a different option  to change it.... so again there is no standard.

lib/python/Tools/Multiboot.py
TwolDE: OK, that works at the moment

lib/python/Screens/FlashImage.py
TwolDE: Again works

lib/python/Screens/FlashImage.py
TwolDE: Agree, but then that is also true of the following lines for the hd51 (which has changed 2x so far) and the zgemma.
To be honest allowing a mode change (from the defaults) is a luxury not used by most people, and in some situations on some mode 12 receivers (not currently supported by OpenPli) can lead to issues.

zgemma-star:
Can you please post from HD model and GB model the information of:

/sys/firmware/devicetree/base/chosen/kerneldev
/sys/firmware/devicetree/base/chosen/bootargs
/proc/cmdline

TwolDE2:
HD51
Cmdline:
root=/dev/mmcblk0p9 rw rootwait hd51_4.boxmode=1 brcm_cma=440M@328M 
kerneldev:
brcm_cma=192M@768M console=ttyS0,115200
bootargs:
emmcflash0.kernel4#root=/dev/mmcblk0p9 rw rootwait hd51_4.boxmode=1#

**Gbquad4k:**

cmdline:
root=/dev/mmcblk0p7 rootwait rw rootflags=data=journal libata.force=1:3.0G,2:3.0G,3:3.0G coherent_poll=2M vmalloc=525m bmem=529m@491m bmem=608m@2464m

bootargs:
root=/dev/mmcblk0p7 rootwait rw rootflags=data=journal libata.force=1:3.0G,2:3.0G,3:3.0G coherent_poll=2M vmalloc=525m bmem=529m@491m bmem=608m@2464m#

kerneldev: no file

zgemma-star:
bmem and cma settings are you sure this is right??
Can you explain why you use a different memory mapping.
I think the cmd is set in CFE can you confirm this is right.
  
GB:
coherent_poll=2M vmalloc=525m bmem=529m@491m bmem=608m@2464m
your settings:
coherent_poll=2M brcm_cma=764M@0x10000000 brcm_cma=1024M@0x80000000'

TwolDE2:
1. You have raised an excellent point:
2. the hd51 settings are also not correct in flashimage as there is a memory change between mode 1 and mode 12
					if self.optionsList[self.option][0] == "boxmode=1":
						newboot = newboot.replace("520M@248M", "440M@328M")
						newboot = newboot.replace("200M@768M", "192M@768M")
					elif self.optionsList[self.option][0] == "boxmode=12":
						newboot = newboot.replace("440M@328M", "520M@248M")
                                                newboot = newboot.replace("192M@768M", "200M@768M")
and although teh values have changed they are diferent between the 2 modes.
So you are correct gbue4k and gbquad4k are different and I should have checked!
Which means I guess that we need to look at whether we do this mode switch and if so how.
﻿Gbquad4k:

cmdline:
root=/dev/mmcblk0p7 rootwait rw rootflags=data=journal libata.force=1:3.0G,2:3.0G,3:3.0G coherent_poll=2M vmalloc=525m bmem=529m@491m bmem=608m@2464m

bootargs:
root=/dev/mmcblk0p7 rootwait rw rootflags=data=journal libata.force=1:3.0G,2:3.0G,3:3.0G coherent_poll=2M vmalloc=525m bmem=529m@491m bmem=608m@2464m#

STARTUP:
boot emmcflash0.kernel2: 'root=/dev/mmcblk0p7 rootwait rw rootflags=data=journal libata.force=1:3.0G,2:3.0G,3:3.0G coherent_poll=2M vmalloc=525m bmem=529m@491m bmem=608m@2464m'

kerneldev: no file

gbue4K

cmdline:
﻿root=/dev/mmcblk0p9 rootwait rw rootflags=data=journal libata.force=1:3.0G,2:3.0G,3:3.0G coherent_poll=2M brcm_cma=764M@0x10000000 brcm_cma=1024M@0x80000000

bootargs:
root=/dev/mmcblk0p9 rootwait rw rootflags=data=journal libata.force=1:3.0G,2:3.0G,3:3.0G coherent_poll=2M brcm_cma=764M@0x10000000 brcm_cma=1024M@0x80000000#

STARTUP:
boot emmcflash0.kernel3 'root=/dev/mmcblk0p9 rootwait rw rootflags=data=journal libata.force=1:3.0G,2:3.0G,3:3.0G coherent_poll=2M brcm_cma=764M@0x10000000 brcm_cma=1024M@0x80000000'

kerneldev: no file

zgemma-star:
I assume GB uses static cmdline, if so then you can remove in the code :
rootflags=data=journal libata.force=1:3.0G,2:3.0G,3:3.0G coherent_poll=2M brcm_cma=764M@0x10000000 brcm_cma=1024M@0x80000000'

littlesat:
We could also consider to check where the STARTUP file is located.... on 0 or 1....

TwolDE2:
Because you are basically rebooting the GB from STARTUP, I think it is better to combine your idea with the STARTUP_slot e.g.
			if not SystemInfo["canMode12"]:
				startupFileContents = open('/media/mmc/STARTUP_%s').read() % slot
as it removes any ambiguity with GB 4k models.
However, still think hd51 needs code review for different modes.... as far as I am aware not an issue for zgemma or similar receivers. Have always hated the memory size switching by mode on hd51, as it has been changed since 1st release and Octagon SF4008 handles this better and transparently to user.

littlesat:
I do scan now the MCCs where the STARTUP file exists.... probably we can add this also as extra parameter to the CanMultiboot key... so having (1,4,0) or (3,3,1).... The last parameter defines where the STARTUP file is located. Then this can be directly written... especially when the output can be more easier for GB...

zgemma-star:
startupFileContents = open('/media/mmc/STARTUP_%s').read() % slot

Are you sure this is safe to do?
When you use an additional SD card it can been mount as /media/mmc/ 

What is on GB models /media/mmc exactly a mount to /dev/mmcblk0p1 or the rootfs which your using.

Isn't it easier to give the partition a name like GBboot and format it as fat so it's automatic mount as /media/GBboot/

zgemma-star:
@What if manufactures add in /media/"mmc"/ the available optionswith settings
STARTUP should been a symlink to the require multiboot image

Like on zgemma H7 you have 8 options
STARTUP_boxmode1_1
STARTUP_boxmode1_2
STARTUP_boxmode1_3
STARTUP_boxmode1_4
STARTUP_boxmode12_1
STARTUP_boxmode12_2
STARTUP_boxmode12_3
STARTUP_boxmode12_4

If GB use same
STARTUP_boxmode5_1
STARTUP_boxmode5_2
~

TwolDE:
To some extent we are 2nd guessing the manufacturers.
Hd51 used to have STARTUP‘s for mode 1 and mode 12(so 8 like the zgemma), but now just the 4, and the default was changed to mode 1, because thats the only way Kodi works with the hd51.
GB recently had 8 after updating the boot process, alllowing you to use the old system (why??  !!)
Octagon SF4008 had both, but changed to using a setup/menu option - although I am no reference for Octagon this was/is better... as it allows a simple way for a user to change.

To be absolutely safe, I would prefer Littlesat‘s idea to find the volume with the STARTUP (which is typically “at the moment“ mmc, but NOT on other images!!) and use that as the basis.
For the hd51, I would use the appropiate STARTUP and replace the mode and MEM config.
Whichever way is chosen, there is a likelyhood that a manufacturer will change something that could cause an enigma crash - although hopefully the current boxes have reached some stability. 

...... but I am really open to both your ideas!...... as long as I can understand the concept.

TwolDE2:
Changes Multiboot to use "canmode 12" as detection, and FlashImage to following and multiboot reboott appeared to work on both hd51 & gbue4k without issues on mode1.
			model = HardwareInfo().get_device_model()
			for media in ['/media/%s' % x for x in os.listdir('/media') if x.startswith('mmc')]:
				if 'STARTUP' in os.listdir(media):
					break
			if not SystemInfo["canMode12"]:
					startupFileContents = open('%s/%s_%s' % (media, 'STARTUP', slot)).read() 
			elif slot < 12:
				startupFileContents = "boot emmcflash0.kernel%s 'root=/dev/mmcblk0p%s rw rootwait %s_4.boxmode=1'\n" % (slot, slot * 2 + 1, model)
			else:
				slot -= 12
				startupFileContents = "boot emmcflash0.kernel%s 'brcm_cma=520M@248M brcm_cma=%s@768M root=/dev/mmcblk0p%s rw rootwait %s_4.boxmode=12'\n" % (slot, SystemInfo["canMode12"], slot * 2 + 1, model)
			open('%s/%s' % (media, 'STARTUP'), 'w').write(startupFileContents)

However, "canmode12" specifies 192M@768M for the hd51 in mode 12 and I think it should be 200M@768M, 192 is mode 1 ????????
Also if on mode 12 it doesn,t flip the values, just changes boxmode = 1

zgemma-star:
https://github.com/zgemma-star/meta-zgemma/blob/master/classes/zgemma-emmc_image_multi.bbclass#L59

Check available multiboot slots
if mode is available check which mode number are available
When customer select the slot and if available mode number then just copy/save the file as STARTUP

A manufacture can add more options if it's necessary.

TwolDE:
The zgemma is just a copy of the old hd51 class definition..... which works well.
In other images,   /boot contains all the startup files so its dead easy to pick up the startup files, does that mean there is a sym link to the boot volume or what?? (As per your previous post??) Certainly in OpenPli it just contains the bootup images (.mvi).

TwolDE:
.... and if it is a sym link where is it setup? In the build?

littlesat:
Put it in /boot with sim link is not done... thr bsp layer put it on a location and that was also why I made that ‘scanner’ to locate it... then I think it is even better to indicate in the systeminfo where it is located. Not by number but just a string “mcc....0’ or ‘mcc...1”

littlesat:
Sorry I meant mmc iso mcc

TwolDE:
Littlesat, I am happy providing we have a defined “source“  so that we don‘t have to go hunting for the volume with the STARTUP‘s. I even agree with zgemma-star that it would be worth “modifying the  “class“ definitions to add mode 1 & mode12 STARTUP‘s for the hd51.
It makes life easier and prevents issues when switching modes.

TwolDE:
What do we need to add or utilise to add into the SystemInfo?

littlesat:
Just extend the Tupolev with the location of the Startup file... (1,4,”MMC...0) for HD and (3,3,”mcc...1” for gb.... sorry for being cryptic...

TwolDE2:
@zgemma-star 
can you confirm the settings in boxmode 1 & 12 ??
'brcm_cma=440M@328M brcm_cma=192M@768M" is boxmode 12 in OpenPli? 
Is boxmode 1 the same?? 
It is the same in the OE_A.

littlesat:
Boxmode 1 does not require the memory settings in Startup as there the default is fine

TwolDE:
The issue is that you are assuming that OpenPli STARTUPs are the same as OE-A (and they are not).
1) it depends whether the box is flashed 1st with OpenPli or an OE-A image and when you flash a new image into one of the slots, these are not changed only the STARTUP built by the image. 
2) on a mode switch reboot OpenATV reloads the memory configs for mode 1 and mode 12 into the built STARTUP or copied STARTUP_x to ensure integrity.... these settings are not necessary the same as those currently set in OpenPli canMode12.
I am not saying they are correct(and I am probably missing something), but at the moment I don‘t see an agreement on the default memory configs between the OpenPli and OE-A.

So the question is what is the default mode 1 memory config?

I would rather load the memory configs into OpenPli “canMode12“ as a tuple and replace from there when you build the STARTUP for either bootmode 1 or 12.... it would also allow for any difference between hd51, H9 or other mode12 capable box.

littlesat:
The memory settings are default for mode1. they do not need memory extra's....  And when a OE-A image is not satisfy with it they have the default settings not good in their bsp layer. I prefer to have only the real required stuff in the STARTUP file(s). Do we have issue with restarting to e.g .OpenATV? Then I think OpenATV should adapt -or- we should give memory settings that are actually not required at all when the 'other' image does it correctly...

I prefer to put the STARTUP location in the tupple and then also remove the scan for the STARTUP file... Then SystemInfo.py defines what/where the STARTUP file should be located....

Using the canmode12 is misuse of this feature and not the right way to determine HD or GB. Probably we can also return a tupple here for the memory configs instead of giving just one number...

In two weeks I have some time to help here....so please hold on... -or- when you can do another merge request (a clean other one)... I'm happy!!! And later always I can change it a bit...

TwolDE2:
@littlesat - so I have added code to temporarily mount the boot volume from SystemInfo, and write the new STARTUP (code in a tools/multiboot call) - so avoids using mmc - and works on both my HD51 and gbue4k. It removes the need for the STARTUP scan, so much neater.
The only comment here is that both these boxes use mmcblk0p1.
The code currently uses Canmode12 for the GB box, but in some ways this is a good test as it differentiates between  boxes with multiple boot modes (hd51 etc) and those that haven't (e.g.GB).
My question about memory settings, is that OpenATV use different to OpenPli in mode 1/12... but both work. ( I have one version of code that uses tuples to load up the memory configs - I certainly think it's better than hard coding the numbers)
I will load up the changes tomorrow, and then you can review & cleanup etc before applying!!

TwolDE2:
So code added, also includes correction to Tools/multiboot.py (remove line of my code that picks up version info)

littlesat:
I also lost of the complete patch can you please made a complete new merge
 Request

TwolDE:
Both hd51 & GB use same volume for boot/STARTUP, however, I think it is still probably worth keeping the volume in SystemInfo as some of the other multiboot receivers may be different .... I will check tomorrow. I certainly remember some differences in other code for other receivers
Using tools/multiboot for all calls, is to me a really neat trick. Although perhaps not important for OpenPli, it also means I can use the same code in other images/OpenPli modules.... makes life much easier.
I will try tomorrow without the mount, but the code is so fast that purely from a “safety“  point worth keeping perhaps.
I ran a few clean builds today, so plan to test tomorrow on all my boxes, perhaps cleanup a couple of lines of code  and then reapply.
Last time I gave you a heart attack, so is the best approach, to close the pull,  revert the commits and then recommit the complete change with a new pull request???

zgemma-star:
As @littlesat informs mode1 doesn't need memory mapping.

quote: "The issue is that you are assuming that OpenPli STARTUPs are the same as OE-A (and they are not)."

Openpli and OE-A set both the standard mode 1 memory mapping and are so the same!!
Only in mode12 the memory mapping should been adjust.
https://github.com/oe-alliance/oe-alliance-core/blob/4.1/meta-brands/meta-airdigital/recipes-linux/linux-airdigital-4.10.12/h7/defconfig#L613

littlesat:
I thought HD had the STARTUP in partition one... did I made a mistake here... when this is the fact we could consider to simply always use slot 1 for it... When another box comes with a different slot we can still decide to go to extend the tupple....
Thanks for being so patience!!! And thanks to go for Q...

littlesat:
And sorry to say can you please push another brand new merge request... So we can close this one so we get a new fresh overview?

littlesat:
And I do not understand why with cannot mode 12 you do something with a slot or define the content of the STARTUP file.... Couldn't better just write the file with the content... I also do not understand why we need to mount a partition... I think just simply write the file...

littlesat:
Actually I prefer something like this...
@what if manufactures add in /media/"mmc"/ the available optionswith settings
STARTUP should been a symlink to the require multiboot image

Like on zgemma H7 you have 8 options
STARTUP_boxmode1_1
STARTUP_boxmode1_2
STARTUP_boxmode1_3
STARTUP_boxmode1_4
STARTUP_boxmode12_1
STARTUP_boxmode12_2
STARTUP_boxmode12_3
STARTUP_boxmode12_4

If GB use same
STARTUP_boxmode5_1
STARTUP_boxmode5_2
~
But manufacturers should help here....

TwolDE2:
@zgemma-star thanks for the clarification 
Don't have an issue with the h7 (both OpenPli & OE-A use the same - as you have said), but for the hd51 in OpenPli bootmode 12 we have 192M@768 and in OE-A bootmode 12, it is 200M@768M (both h7 and hd51 are treated the same - hardcoded in MultiBootStartup.py).

TwolDE2:
@Littlesat - brand new merge - no problem!!!
1. tools/multiboot.py 
for h7/hd51, the STARTUP is generated in FlashImage.py and written directly to mmcblk0p1
for the gbue4k & gbuhd4k (which have different STARTUP files!!! and only one bootmode) it is safer to read the selected STARTUP (using slot) and writing it back to STARTUP.  
2. In theory using generated STARTUP_boxmode_1 etc files is OK, but assumes same standard, which is not true even for OpenPli at the moment.
For boxes like GB where you only have one bootmode, it is safe to pick up the STARTUP_x and copy to STARTUP. For h7/hd51, its safer to generate.
3. Using mmc! Until yesterday, I might have agreed but whereas media/mmc DID point to the Boot/startup volume - NOW it points to the Image volume. I haven't changed anything its just changed during the last 2 image flashes. I have just done a complete clean flash using ViX rather than OpenPli and the same is true - mmc mount is now different.
So its is safer, to use mmcblk0p1, as we KNOW this is the correct volume.   
Whether we have to mount it first, I will check out in the next 2 hours and report back!!!

TwolDE2:
Sorry, ran code with direct open, no mount and it doesn't work for me, across all boxes!
On the hd51, mmcblk0p1 now appears under media (see mmc comment above) and works.
On the gbue4k, mmcblk0p1 is not mounted and causes crash on Open.
So maybe it can be done, but not by me!
Will post new merge and pull probably later today.

PR #1451, opened 2018-04-13
Update hu.po

mcfly822:
Some more translated lines and minor bug corrctions by wysless & mcfly822.


mcfly822:
Closed because conflict.
(The repo updated just while I uploaded my translation. Will correct in next PR)

PR #1453, opened 2018-04-13
Adapt Makefile.am to allow .po files creation on OSX

pr2git:
On OSX (Mac) the sed command is a BSD one and option are different, for the GNU version gsed must be used. Makefile autoconf create the variable SED that point to the correct "sed" (so on Mac point to gsed) so this is working fine if we use $(SED) in Makefile.am
With this change it is now possible to generate the po files on an Mac without manually editing the enigma2-py.pot file to change the CHARSET.


WanWizard:
@eriksl ?

pr2git:
Since you didn't have problem with this PR:

https://github.com/OpenPLi/enigma2-plugins/commit/03b162baefb39cc830134d2fe9c72341b141003e

Can you please include this one too? 
Since it does exactly the same using $(SED) in for the CHARSET replacement in the Makefile.

PR #1467, opened 2018-04-18
Multiboot -GB  ver 3 

TwolDE2:
OK, this all works for me ... so it's all yours


zgemma-star:
Please, copy your repository on your PC and remove your develop branch.
Fork a new Openpli Enigma2 and make your own branch based on develop branch ( git checkout -b develop-multiboot )
Copy your changes in the map and commit/pull it to your git. Then make a pull request.
Now it's a mess with 16 commits when 2 or 3 is enough.

littlesat:
Deciding GB/HB on CanMode12 is not done....

littlesat:
Can we use the number of slots from that tupple to get the HiSlot value? (and avoid the extra local variable)
Can't we simply write the STARTUP file instead of mounting the STARTUP partition...?

littlesat:
Does '/sys/firmware/devicetree/base/chosen/bootargs' on GB.... this might be a good check to determine between GB/HD instead of using the mode12 key...

littlesat:
Why do we need Startupx parameter... I think we better simply could write the STARTUP file.. I don't like that unmount/mount stuff as it makes it too complicated while you can do it with a one liner that simply writes the file.... And the tupple could be 2 objects the mount path to startup is the same for both systems....

TwolDE2:
@zgemma-star - totally agree with you - I am going to restart, so its just one commit... just my learning process.

TwolDE2:
@littlesat - if you have a nice way of doing this then its fine by me.... I am going to cleanup this mess.
So will disappear for a while.
In between, when you have time, if you can improve the code etc

littlesat:
I did re-open this merge request so it does not disappear from the 'radar'... I think it is mandatory we fix the  multiboot for GB

littlesat:
I suggest we should make it, but the ideal situation of the possibilities and also the check if the feature exists is check if we have a STARTUP file in the first partition.... Then the BSP layer creaters should help...

TwolDE2:
@Littlesat FYI ... before I reboot this fork. I am now using:-
if SystemInfo["canMultiBoot"] and 'coherent_poll=2M' in open("/proc/cmdline", "r").read():
to check for  the Gigablue ... sometimes obviously the SystemInfo["canMultiBoot"]  is not needed, depends on the code section.
As both OpenViX and OpenPli builds are currently messed up, I will concentrate on cleaning this up.

zgemma-star:
maybe add a file called for example like STARTUP_choices
#' available boxmodes
boxmode = "1, 12"
#' available multiboot partitions
nand = ""
emmc = "4"
#' support multiboot from USB (yes/no)
usb = ""
#' support multiboot from SDcard (yes/no)
sdcard = ""
kernel1 = "boot emmcflash0.kernel1"
kernel2 = "boot emmcflash0.kernel2"
kernel3 = "boot emmcflash0.kernel3"
kernel4 = "boot emmcflash0.kernel4"
rootfs1 = "root=/dev/mmcblk0p3"
rootfs2 = "root=/dev/mmcblk0p5"
rootfs3 = "root=/dev/mmcblk0p7"
rootfs4 = "root=/dev/mmcblk0p9"
#' memorymapping if boxmode are available
boxmode1 : ""
boxmode12 : "brcm_cma=440M@328M brcm_cma=192M@768M"

betacentauri:
The number of slots/partitions is not fix. Every image builder can change it, if he/she wants. So e.g in worst case PLi builds a disk.img with 4 slots and ATV one with 3 slots.

TwolDE:
@betacentauri --- now you tell me!!! but obviously in theory you are correct!
Seriously, one would hope that existing layouts will not change - but otherwise apart from counting volume/partition pairs upto the max capacity of the eMMC, how would you suggest doing it??
 - as the number of STARTUP' s is also variable.... and GB don't help by adding in a recovery partition that effects the 1st slot position.
Haven't checked your ofgwrite code recently, but how do you verify the -m(slot no) parameter??

betacentauri:
Good question :smile: 
Currently my tool reads the GPT. So the partition table. The partitions have to be named rootfsX and kernelX (with X a number). If a box use old MBR or use other naming it won't work.
I already thought about it, but I don't think it's easy to implement that in E2.

betacentauri:
And I really hope that it will be theory for all time. Maybe ignore that for now.

TwolDE:
@betacentauri - thanks, you learn something every day - didn't know they were labelled ... so if we had to do it, we could mount each partition, check its label and work out the number of slots from that.
Although, leaving the number of (current) slots in the SystemInfo, means we can change very quickly.

betacentauri:
I'm not sure whether mounting is sufficient to get the partition name. I currently use fdisk code to get the names.

TwolDE:
@betacentauri - I am sure you are correct - hadn't thought enough before responding.
... now if we had Standards for all this  - just joking, would never happen!

PR #1470, opened 2018-04-19
Backport skin patch from OpenBlackHole team

MastaG:
This used to be part of their public CVS which was based on our enigma2.
It basically adds support for the analog clock, which is used in many BlackHole skins.


littlesat:
Sorry I don't like the names Bhclock.py and BhAnalogic.py... Please rename it to e.g. AnalogClock.py -and- Gauge.py

Nice 'gadget' to have an analogue clock on the display of your box when the box is in standby...

MastaG:
Thanks for the feedback littlesat,

I've renamed the files :)

PR #1477, opened 2018-04-20
Multiboot

TwolDE2:
So rebuilt the fork of OpenPli develop, added multiboot branch.
Added code with all changes.
Works as far as I can test on hd51 & gbue4k - unable to run new build at the moment!!
SystemInfo contains multiboot modes, memory configs, STARTUP boot volume, slot differences
FlashImage contains code to detect GB, and send STARTUP_slot to tools/multiboot (where much of the base code resides), to read STARTUP_slot and write to STARTUP.
If mode 1 or mode12 request, creates new STARTUP and sends to tools/multiboot to write.

### DIFF
--- a/data/rc_models/rc_models.cfg
+++ b/data/rc_models/rc_models.cfg
@@ -41,6 +41,7 @@ et7000mini et7000mini
 et11000	et7000mini
 # Gigablue
 gbquad4k gb7252
+gbue4k gb7252
 # Max Digital
 xp1000   xp1000
 # Miraclebox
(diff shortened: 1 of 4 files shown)


littlesat:
Sorry for going again...
Probably in the latest chaos thread we missed something...

Please remove  "mmcblk0p1" from the tuple and don't use the class WriteStartup(): as this makes stuff complicated... Just write directly in the file or in case of GB copy the file or create the content of the file like you did before in the orriginal merge request...... No mount/unmount required...

The current get current image function can be made easier and more general.. by only checking kerneldev

[code]
def GetCurrentImage():
	if SystemInfo["canMultiBoot"]:
		return (int(open('/sys/firmware/devicetree/base/chosen/kerneldev', 'r').read().replace('\0','').split('=')[1].split('p')[1].split(' ')[0])-SystemInfo["canMultiBoot"][0])/2
[/code]

TwolDE2:
kerneldev is not present for GB, which why we have both checks.... that's how this conversation started!
WriteStartup: you will have to clarify what you want instead.
removing mmcblk0p1 from the tuple is not a big issue, it wasn't my suggestion - it's easier on the code.

littlesat:
I simplified it a bit and now it is committed....

PR #1480, opened 2018-04-21
[ActionMap] Improve initialisation and logging, add HelpableNumberActionMap

IanSav:
- Use a slightly longer form of initialisation for the ActionMap `__init__`.  This format avoids an issue where the initialisation parameters become persistent across all invocations of this code.  This is inappropriate and undesirable.

    In essence the problem with the original code is that it appears to be an efficient way to create a default empty list and dictionary.  What actually happens is that these items get created as a *persistent* objects.  Every invocation of the `__init__` method that doesn't specify the parameters will use that *SAME* (persistent) objects.  Any changes to either object will *persist* (be remembered) and be carried forward into every other invocation of this method!  THIS IS NOT WHAT SHOULD HAPPEN!

    Please refer to http://pylint-messages.wikidot.com/messages:w0102 and https://nedbatchelder.com/blog/200806/pylint.html for more details.

- ActionMap / keymap errors are now reported in a single log line rather than taking two lines.

- Add a HelpableNumberActionMap for action map symmetry.

- PEP8 clean-up and fixes.

These changes should be fully backward compatible with all exiting Enigma2 code.  No changes to any existing code should be required required.

This code is based on code written by Prl (Nov-2014 - Jan-2015).

### DIFF
--- a/lib/python/Components/ActionMap.py
+++ b/lib/python/Components/ActionMap.py
@@ -1,9 +1,16 @@
+from Tools.KeyBindings import queryKeyBinding
+
 from enigma import eActionMap
 
+
 class ActionMap:
-	def __init__(self, contexts = [ ], actions = { }, prio=0):
-		self.actions = actions
+	def __init__(self, contexts=None, actions=None, prio=0):
+		if contexts is None:
+			contexts = []
+		if actions is None:
+			actions = {}
 		self.contexts = contexts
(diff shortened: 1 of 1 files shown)


IanSav:
This Pull Request is being replaced by Pull Request #1483 that has been reformatted as requested by Littlesat.

PR #1481, opened 2018-04-21
Added help texts for movieplayer settings menu.

ims21:
- changed "Play audio in background" text to true significance

(PLi skins will be modified soon)


littlesat:
Why on rc as actually this is a new feature

ims21:
I do not understand your notice ... it is for develop, not for rc ... 

"1 commit into OpenPLi:develop from ims21:develop"

PR #1483, opened 2018-04-22
Fix and Improve ActionMap

IanSav:
- Use a slightly longer form of initialisation for the ActionMap `__init__`.  This format avoids an issue where the initialisation parameters become persistent across all invocations of this code.  This is inappropriate and undesirable.

  In essence the problem with the original code is that it appears to be an efficient way to create a default empty list and dictionary.  What actually happens is that these items get created as a *persistent* objects.  Every invocation of the `__init__` method that doesn't specify the parameters will use that *SAME* (persistent) objects.  Any changes to either object will *persist* (be remembered) and be carried forward into every other invocation of this method!  THIS IS NOT WHAT SHOULD HAPPEN!

  Please refer to http://pylint-messages.wikidot.com/messages:w0102 and https://nedbatchelder.com/blog/200806/pylint.html for more details.

- ActionMap / keymap errors are now reported in a single log line rather than taking two lines.

- Add a HelpableNumberActionMap for action map symmetry.

- Use a slightly longer form of initialisation for the HelpableActionMap `__init__`.  This format avoids an issue where the initialisation parameters become persistent across all invocations of this code.  This is inappropriate and undesirable.

  See explanation in ActionMap comment avove.

- Perform a PEP8 clean up of the code and comments.

### DIFF
--- a/lib/python/Components/ActionMap.py
+++ b/lib/python/Components/ActionMap.py
@@ -1,9 +1,13 @@
 from enigma import eActionMap
 
 class ActionMap:
-	def __init__(self, contexts = [ ], actions = { }, prio=0):
-		self.actions = actions
+	def __init__(self, contexts=None, actions=None, prio=0):
+		if contexts is None:
+			contexts = []
+		if actions is None:
+			actions = {}
 		self.contexts = contexts
+		self.actions = actions
 		self.prio = prio
 		self.p = eActionMap.getInstance()
 		self.bound = False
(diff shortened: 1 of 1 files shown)


IanSav:
Why was this version closed?

Where is the discussion regarding any problems here in GitHub or on the forum?

There are five different commits in this pull request.  Are all five unacceptable?  If not, then why weren't the acceptable changes accepted?

EDIT: The old version of this pull request was merged instead of this later one.  That is a problem as this latest version addresses an issue with an old legacy plugin that can't be changed to work with the newer ActionMap code.

PR #1485, opened 2018-04-22
FlashImage - fixed typo

ims21:


littlesat:
Fix is not required the () are in.... As you can see it also failed....

ims21:
then this checks are for nothing ... brackets missing in both actual sources - in rc and in develop too...

PR #1486, opened 2018-04-22
[ConfigList.py] PEP8 clean up

IanSav:
This is simply a PEP8 clean up with no functional changes.


IanSav:
I missed a correction.

PR #1499, opened 2018-04-27
Added several early initialized ConfigYesNo duplicates into UsageConf…

ims21:
…ig.py due graphics switches

### DIFF
--- a/lib/python/Components/ParentalControl.py
+++ b/lib/python/Components/ParentalControl.py
@@ -18,6 +18,7 @@ LIST_BLACKLIST = "blacklist"
 FLAG_IS_PARENTAL_PROTECTED_HIDDEN = 256
 
 def InitParentalControl():
+	# Note: This ConfigYesNo items must be duplicated in UsageConfig.py for graphics switches due early initialization
 	config.ParentalControl = ConfigSubsection()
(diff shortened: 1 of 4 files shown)


littlesat:
I’m afraid this is not the real solution... anywhere else buttons show up without predef in usageconfig

ims21:
Yes, it is "hack", but anytime it can be removed, if will be fixed ConfigBoolean for graphic yes/no for early init.

Btw - then I do not know, why you told me - create PR, If I wroted, how it works for me...

ims21:
Problem is, that skin reading begins after init this items => pictures not exist in this time => ConfirmYesNo is as texts for this items only. 

Then or make graphic on/off not skineable or read skin earlier or use this hack.

littlesat:
What can be done here is load the pixmap in skin.py by putting it directly behind the keys in that dictionary and then don't load the pictures at the __init__ in the config, but simply return them via getMulti in the config class instead of 'caching' them via self.falseIcon and self.trueIcon.

ims21:
I still do not know why you fixed your ability to posibility change images by setting them in skin.xml.
It does not matter if user have to change the picture in the skins or he must add another image and change the skin.

PR #1505, opened 2018-04-27
	fixes gstreamer version detection:   About.py

lprot:


pieterg:
Just one remark, this does not match versions such as 0.10 does it?
I could not quickly find a satisfactionary glob at the time, that’s why I went for the ‘any gstreamer control file’ approach

WanWizard:
I'm not too worried about that, 0.10 is behind us now. But it also doesn't match 6.78. Not sure why the control file is now called "1.0" while the current version is "1.14.0". Where does that name come from?

lprot:
Just a wild idea - process all files but take max version. That will do what you want.

WanWizard:
It gets the version from the package name: `gstreamer1.0_git.bb`. So this should be ok. 

But if needed an optional second digit can be added to the regex, both before and after the the dot, so 0.0 to 99.99 is covered?

PR #1513, opened 2018-04-29
eLabel - added fce for return noWrap status

ims21:


littlesat:
Why needed?

ims21:
If is used noWrap, then must be used diferent procedure for autosize font in label.  
And this info, if is set noWrap is not without this known

For future using ...

PR #1514, opened 2018-04-29
NumericalTextInputHelp-automatically sets the font size by content.

ims21:
- all fields must be the same width.


ims21:
I will put new PR

PR #1546, opened 2018-05-08
fixed pip resize/move for some models

ims21:
- code by littlesat


pieterg:
Are you sure this works on all framebuffer resolutions? (e.g. full hd skin, hd skin, sd skin)

ims21:
I tried it under FullHD, HD and SD and still works.  Why ?

It is limit for 4 boxes only, because PIP finished on it's size, but values could be still increased by pressing ch+. Then will problem with moving or with decreasing...

pieterg:
My gut feeling is that pip dimensions might be scaled with output resolution, and possibly framebuffer resolution (but if you tried with different skins we can rule out the framebuffer resolution).

What about output resolution?
Is the limitation the same for each output mode?

PR #1574, opened 2018-05-21
httpstream: Call getConfigBoolValue from mainloop

betacentauri:
Python code can only be called from the mainloop. This fixes
various crashes when httpstreams were used.


athoik:
🥇 hooray!

PR #1581, opened 2018-05-26
fix typo in Makefile.inc regarding LIBSDL_CFLAGS

technic:
This is a minor commit, but SDL still doesn't work because of enigma2 changes in gMainDC interface.
can we setup travis-ci to build with different configurations? (-with/without-sdl)


technic:
pushed one more commit to make it compile without errors.

athoik:
Those changes look safe. Anyone against merging?

WanWizard:
I don't have the skills to comment. ;)

technic:
This changes doesn't concern any code that is built for stb binary. So
indeed they are safe :)

PR #1589, opened 2018-05-28
Script to update all the po files out of normal build process

pr2git:


WanWizard:
Shouldn't this clean up all temp files created?

pr2git:
It acts exactly the same as when you update the files with the make process. But if you want it is of course possible to clean up the 2 temporary .pot files and the final enigma2.pot (even if it would be nice to have the enigma2.pot file proposed in your github for people willing to add a new language).

pr2git:
rm added to remove the pot files in the PR.

WanWizard:
Thanks !

This probably needs to happen in the other scripts too?

PR #1592, opened 2018-05-29
Update amldecoder.cpp

persianpros:


lib/dvb/amldecoder.cpp
```diff
@@ -249,6 +249,10 @@ RESULT eAMLTSMPEGDecoder::setAudioPID(int apid, int type)
 				m_codec.audio_type = AFORMAT_AC3;
 				t = "AC3";
 				break;
+			case aDDP:
+				m_codec.audio_type = AFORMAT_EAC3;
```
WanWizard: where is AFORMAT_EAC3 defined?

WanWizard, line 253: where is AFORMAT_EAC3 defined?

pieterg:
Can't we just remove all this terrible #if AML stuff which adds non-linuxdvb api workarounds?
It should never have been accepted into e2 in the first place if you ask me...

Is anyone still using it?

mirakels:
This was needed to support Wetek boxes. 
As Wetek apparently is not supported anymore this code could be removed.

littlesat:
+1

PR #1612, opened 2018-06-20
Update encoding.conf

persianpros:
Enable UTF-8 for Persian
"Codepage 1256 (Microsoft Windows Arabic)" supports Persian but it's not defined in enigma2's encoding so enable UTF-8 for Persian.

### DIFF
--- a/data/encoding.conf
+++ b/data/encoding.conf
@@ -17,6 +17,10 @@ slk ISO6937
 chi GB18030
 zho GB18030
 chn GB18030
+fas UTF-8
+per UTF-8
+fa UTF-8
+pes UTF-8
 
 #Fallback encoding when in dvb-text no encoding table is given
 #and no Countrycode  and no transponders configs in this config file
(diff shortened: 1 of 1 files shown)


eriksl:
Rejected in this form. There is a whole bunch of lines that seem to not have changed, but are replaced, maybe whitespace? Whitespace changes must always be in a separate commit.

persianpros:
No github changes the file encoding by default, test it yourself.

WanWizard:
This is the problem: https://github.com/OpenPLi/enigma2/commit/ab58ac20cf6dae53e78841ff8cf97de488a989c6.diff which is an encoding change of the file.

persianpros:
So is my commit right or wrong?

littlesat:
This patch is not good...

persianpros:
When I want to edit this file github says:

We’ve detected the file encoding as ISO-8859-1. When you commit changes we will transcode it to UTF-8.

So would you please do it yourself?

persianpros:
https://github.com/OpenPLi/enigma2/pull/1617

PR #1627, opened 2018-07-04
Add DTS-HD,WMA pro downmix and AC3 plus, AAC transcoding control

persianpros:
Let enigma2 control:

DTS-HD downmix
WMA pro downmix
AC3 plus transcoding
AAC transcoding

for supported STBs, also make choices translatable not like https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Components/AVSwitch.py#L180 for example.

Also I don't get why did you use "downmix_ac3" instead of "downmix_aac" for https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Screens/AudioSelection.py#L265

### DIFF
--- a/lib/python/Components/AVSwitch.py
+++ b/lib/python/Components/AVSwitch.py
@@ -202,6 +202,50 @@ def InitAVSwitch():
 			open("/proc/stb/audio/aac", "w").write(configElement.value and "downmix" or "passthrough")
 		config.av.downmix_aac = ConfigYesNo(default = True)
 		config.av.downmix_aac.addNotifier(setAACDownmix)
+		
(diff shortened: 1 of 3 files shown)


athoik:
> Also I don't get why did you use "downmix_ac3" instead of "downmix_aac" for https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Screens/AudioSelection.py#L265

Because by default the audio menu controls only ac3 downmix. So from audio menu you can globally turn off or on the downmix.  That was the idea few years ago.

PS. What you add in Audio Selection changeAC3Downmix currently is doing nothing!

PR #1630, opened 2018-07-07
Add DTS-HD,WMA pro downmix and AC3 plus, AAC transcoding control

persianpros:
Add DTS-HD,WMA pro downmix and AC3 plus, AAC transcoding control

Let enigma2 control:

DTS-HD downmix
WMA pro downmix
AC3 plus transcoding
AAC transcoding

for supported STBs, also make choices translatable not like https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Components/AVSwitch.py#L180 for example.


littlesat:
Can you please move the system info stuf from av switch to systeminfo with filecheck....

persianpros:
Yeah sure give me some time :)

eriksl:
Which hardware can actually do this (and not only advertise it)?

PR #1637, opened 2018-07-08
More "SystemInfo.py" use.

persianpros:
Move "AVSwitch.py" checks to "SystemInfo.py" as @littlesat suggested before.


persianpros:
That would be great :)

PR #1638, opened 2018-07-08
Let enigma2 manage HDMI audio source.

persianpros:
Manages: /proc/stb/hdmi/audio_source


eriksl:
Note: not every file change needs to be a separate commit. Just split on different functionality. This is all the same feature, so all three files can be in one commit.

PR #1642, opened 2018-07-08
[RecordTimer] add tuner info to the log at start and stop of recordings.

Dima73:
Based on an idea and code supplied by original-birdman


eriksl:
Good idea.

PR #1654, opened 2018-07-15
po files updated from develop at 2018-07-15 11:04

pr2git:


persianpros:
You just added "enigma2.pot" which is wrong.

pr2git:
It doesn't harm to provide the enigma2.pot file, it is even better to have it included so anybody can start a new translation language based on it. On every plugin the .pot file is provided, I already provide it in the past and it was not a problem.

PR #1658, opened 2018-07-28
Fix Zgemma h9 CPU info (undefined problem)

persianpros:
/proc/cpuinfo won't give us what we need for h9 instead /proc/stb/info/chipset give us hi3798mv200
After this you will see 1600 MHz (4 cores) too.
Screenshot: https://forums.openpli.org/topic/28532-merge-requests-for-plis-git/?view=findpost&p=912860

### DIFF
--- a/lib/python/Components/About.py
+++ b/lib/python/Components/About.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import sys, os, time
 from Tools.HardwareInfo import HardwareInfo
+from Tools.Directories import fileCheck
 
 def getVersionString():
 	return getImageVersionString()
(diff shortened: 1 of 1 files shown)


lib/python/Components/About.py
```diff
@@ -76,6 +77,10 @@ def getCPUInfoString():
 	try:
 		cpu_count = 0
 		cpu_speed = 0
+		if HardwareInfo().get_device_model() in ("h9") and fileCheck("/proc/stb/info/chipset"):
+			f = open("/proc/stb/info/chipset", 'r')
+			processor = f.readline().strip()
+			f.close()
 		for line in open("/proc/cpuinfo").readlines():
```
WanWizard, line 84: shouldn't there be an "else", a bit pointless to attempt to execute the for loop...

persianpros: No it's not pointless as we get "cpu_count" from /proc/cpuinfo.
And we're getting "cpu_speed" from /sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq , what a mess LOL

littlesat:
Device model check should not be accepted here!

littlesat:
Also three lines to access the file is not done (I know openatv does it Everywhere to make things extra complicated

persianpros:
This code is not in openatv, I did it myself and tested ok.

WanWizard:
@littlesat Perhaps give (or point to) an example on how it should be done?

PR #1659, opened 2018-07-29
Fix Zgemma H9 CPU info

persianpros:
Each commit has descriptions.

PR #1660, opened 2018-07-31
About.py: Fix zgemma h9 cpu and temp

persianpros:
/proc/cpuinfo won't give us what we need for h9 instead /proc/stb/info/chipset give us hi3798mv200
After this you will see 1600 MHz (4 cores) +cpu temp instead of undefined.
Also for temp we need a new proc file which is /proc/stb/fp/temp_sensor
Related to: https://github.com/OpenPLi/enigma2/pull/1659

### DIFF
--- a/lib/python/Components/About.py
+++ b/lib/python/Components/About.py
@@ -80,7 +80,9 @@ def getCPUInfoString():
 			line = [x.strip() for x in line.strip().split(":")]
 			if line[0] in ("system type", "model name"):
 				processor = line[1].split()[0]
-			elif line[0] == "cpu MHz":
+			else:
+				processor = open("/proc/stb/info/chipset").readline().strip()
+			if line[0] == "cpu MHz":
 				cpu_speed = "%1.0f" % float(line[1])
 			elif line[0] == "processor":
 				cpu_count += 1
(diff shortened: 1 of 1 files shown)


littlesat:
And if you do this....
-if line[0] in ("system type", "model name"):
+if line[0] in ("system type", "model name", "Processor"):
-elif line[0] == "cpu MHz":
+elif line[0] in ("cpu MHz", "cpu max MHz"):

littlesat:
The patch you offer is not good... because everyline that does not parse system type of model name will parse the cpuinfo file... I think the two changes I suggest make the stuff compatible again....

littlesat:
max CPU might not work... but Processor at least should work to get the type....

persianpros:
@littlesat 
if line[0] in ("system type", "model name", "Processor"):

works without

elif line[0] in ("cpu MHz", "cpu max MHz"):

but it shows "ARMv7" for cpu which is not specific ...

littlesat:
First crazy that this box does it differently...

littlesat:
Do (all) other boxes also have the other proc? I have another idea read that proc and don’t parse model when we already have it + add the temp...

persianpros:
@littlesat 
AFAIK most enigma2 STBs have /proc/stb/info/chipset but I think it's not a good idea to change everything for just one model ...

You can merge this for now: https://github.com/OpenPLi/enigma2/pull/1663 at least it solves the info problem without exact cpu model.

PR #1663, opened 2018-08-04
About.py: Fix zgemma h9 cpu/temp info

persianpros:
After this you will see 1600 MHz (4 cores) + cpu temp instead of undefined.
For exact cpu model we need to read /proc/stb/info/chipset which gives us hi3798mv200 instead of /proc/cpuinfo but I didn't find a proper way to do it.
For temp we need a new proc file which is /proc/stb/fp/temp_sensor
Related to: https://github.com/OpenPLi/enigma2/pull/1660

### DIFF
--- a/lib/python/Components/About.py
+++ b/lib/python/Components/About.py
@@ -78,7 +78,7 @@ def getCPUInfoString():
 		cpu_speed = 0
 		for line in open("/proc/cpuinfo").readlines():
 			line = [x.strip() for x in line.strip().split(":")]
-			if line[0] in ("system type", "model name"):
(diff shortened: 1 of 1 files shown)


persianpros:
Yes at least it counts the cores shows the speed and temp.

persianpros:
@littlesat should I close this or is it ok?

I'm cleaning up the enigma2-plugins and it's really a huge work: https://github.com/persianpros/enigma2-plugins/commits/master

I'll send a PR when it's done ...

littlesat:
Please leave this open so I can make it as I have time

PR #1664, opened 2018-08-07
Make front LED configurable

persianpros:
Thanks to Hains

### DIFF
--- a/data/setup.xml
+++ b/data/setup.xml
@@ -212,6 +212,7 @@
 		<item level="0" text="Contrast" description="Configure the contrast level of the front panel display.">config.lcd.contrast</item>
 		<item level="0" text="Standby" description="Configure the brightness level of the front panel display during standby.">config.lcd.standby</item>
 		<item level="0" text="StandbyLED" description="Turn on the power LED during standby." requires="StandbyLED">config.usage.standbyLED</item>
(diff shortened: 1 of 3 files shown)


littlesat:
Why is a restart required.... Can't we make it without restart required???

Dima73:
Maybe this for old dreambox

persianpros:
https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Components/Renderer/FrontpanelLed.py changes won't get applied immediately so we need to restart enigma2.

Dima73:
Need add SystemInfo requires="NumFrontpanelLEDs"

persianpros:
And check what exactly?

Dima73:
this https://github.com/OpenPLi/enigma2/pull/1664/commits/dfe3245d7372ad52dc4517196d41f1bda13dee7e#diff-0fdcfce87c5affad740c8fc9a087f165R215

persianpros:
I mean check what proc file for it ...

Dima73:
setup.xml

littlesat:
And also something with addNotification so a restart of e2 is not mandatory when you change the config... doing it this way is not done...

Dima73:
That's right, it's not necessary for anyone.

PR #1667, opened 2018-08-13
setup.xml: Allow changing Transponder Time

persianpros:
Not in normal mode but in advanced and expert mode let us change Transponder Time.

### DIFF
--- a/data/setup.xml
+++ b/data/setup.xml
@@ -1,6 +1,7 @@
 <setupxml>
 	<setup key="timezone" title="Timezone">
 		<item level="0" text="Timezone">config.timezone.val</item>
+		<item level="1" text="Use transponder time">config.misc.useTransponderTime</item>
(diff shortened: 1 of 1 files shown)


littlesat:
What does it exactly do and why is it required? Does the 'NTF plugin' not make this configurable aswell?

persianpros:
NTP related plugins are not pre-installed in openpli and with this option we can simply disable/enable transponder time, it's something inside enigma2 without any control over it ...

For example I'm using command line ntpd and shell scripts so there is no need for a plugin but transponder time breaks everything.

littlesat:
The I prefer not to put this in... as it makes no sense... when you can add ntp manually you can also add the config defaultly in settings.py...
But actually I think when we make such addition then enable ntp with a setting and when the box has internet use ntp and no transponder time

PR #1670, opened 2018-08-16
Refactor VirtualKeyBoard

IanSav:
[VirtualKeyBoard.py] Refactor code

This is a complete rewrite of the VirtualKeyBoard code.  Changes, in no particular order, are:
- Widen the keyboard grid to 14 characters to allow for a more realistic imaging of real keyboards
- The width is now defined by a single variable, all VirtualKeyBoard calculations are now based on this one value
- If the keymap describes a keyboard layout less than 14 characters the keyboard will be centred in the reserved screen area
- There can now be a large number of shift levels coded though only 4 levels are provided by default
- All text based graphical buttons have been redone as translatable text overlay buttons
- All text buttons on each shift level can now be assigned different foreground colours
- The VirtualKeyBoard now has its own distinct ActionMap called "VirtualKeyBoardActions"
- Help descriptions have been added to all ActionMap entries
- Short text entries can be assigned to any button, they will be shown at a reduced front size so as to fit up to about 4 characters
- All text based buttons are now put through the language translation system
- All keyboard buttons are now defined in unicode so that a UTF-8 aware editor is no longer required
- Any button or function can be assigned to a colour button (Defaults: RED->Exit, GREEN->Save, YELLOW->Locale, BLUE-Shift)
- Any button or function can be assigned to a graphical button (matching images need to be created and added to the code)
- Users may now navigate around the VirtualKeyBoard grid AND the text buffer
- Insert and overwrite modes are now available
- The current insert/overwrite mode status is now shown to the user
- Keyboard layouts can now be separate / independent or based on another keyboard layout
- The VirtualKeyBoard supports the standard screen colour buttons and/or the colour buttons displayed on the VirtualKeyBoard
- The initial locale will be matched to the current operating locale, if a match is not found then en_EN (English - Various) is selected
- Pressing YELLOW will now pop up a menu of available locales rather than randomly moving through the locale list
- All locales now have a description of the language and the country
- The VirtualKeyBoard now accepts input via navigation around the grid, by direct SMS character selection and via an attached keyboard
- Some new languages have been added
- The new code uses the automatic skin button code proposed in /doc/BUTTONGUIDE
- The source code documentation has been duplicated and added to /doc/VIRTUALKEYBOARD
- The new code is PEP8 compliant
- There may be more but I think this list is long enough  ;)

Add documentation for the new VirtualKeyBoard
- This documentation matches the comments / documentation encoded in the new VirtualKeyBoard.

[keymap.xml] Add mapping for new VirtualKeyBoard
- This key map includes definitions for the remote control and for an attached physical keyboard.

### DIFF
--- a/data/keymap.xml
+++ b/data/keymap.xml
@@ -69,6 +69,64 @@
 		<key id="KEY_PVR" mapto="file" flags="m"/>
 	</map>
 
+	<map context="VirtualKeyBoardActions">
+		<key id="KEY_RED" mapto="cancel" flags="m" />
+		<key id="KEY_GREEN" mapto="save" flags="m" />
+		<key id="KEY_YELLOW" mapto="locale" flags="m" />
+		<key id="KEY_BLUE" mapto="shift" flags="m" />
+		<key id="KEY_EXIT" mapto="cancel" flags="m" />
+		<key id="KEY_OK" mapto="ok" flags="m" />
+		<key id="KEY_UP" mapto="up" flags="mr" />
+		<key id="KEY_PREVIOUS" mapto="first" flags="m" />
+		<key id="KEY_REWIND" mapto="prev" flags="mr" />
+		<key id="KEY_LEFT" mapto="left" flags="mr" />
+		<key id="KEY_RIGHT" mapto="right" flags="mr" />
+		<key id="KEY_FASTFORWARD" mapto="next" flags="mr" />
+		<key id="KEY_NEXT" mapto="last" flags="m" />
+		<key id="KEY_DOWN" mapto="down" flags="mr" />
+		<key id="KEY_LAST" mapto="backspace" flags="mr" />
+		<key id="KEY_STOP" mapto="backspace" flags="mr" />
+		<key id="KEY_PLAYPAUSE" mapto="delete" flags="mr" />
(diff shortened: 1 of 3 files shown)


persianpros:
Well done :+1:  :100:

littlesat:
Please put all documentation in a seperate file...  😉 I can remove/change that later myself... for the rest it looks good and of course before merging I need a merge request on the standard skins first

persianpros:
PR for skins: https://github.com/littlesat/skin-PLiHD/pull/685

IanSav:
The documentation is already in a separate file.  Are you asking for the comments to be removed from the source file?  If so, that should not be necessary.  The comments are not stored in the .pyo file (I tested this out) and the .pyo file is the same size with or without the comments.

littlesat:
I prefer to have it on one place... to much comments is also not good! This is my meaning... code should be code... we have a folder on git for readme’s and the documentation is also in the commit... make a oneliner in the code and refer to the big documentations. I think code should not include a bookwork.. it should stay boolwork!

IanSav:
When the next update is made to the code (to add a language or whatever) I will remove the documenation from within the code and replace it with a pointer to the /doc/VIRTUALKEYBOARD version.

PR #1677, opened 2018-08-20
[config] add new ConfigDate - using numer key only

Dima73:


Dima73:
Really not error for bulid

PR #1681, opened 2018-08-21
For series added in favourite strip the ending Sxx Exx to search for …

pr2git:
…picon by name.

Exemple in favourite/bouquet:
MySeries S01 E01

picon by name will be:  myseries.png


littlesat:
I think we should not do this as it is really a work-a-round...

PR #1684, opened 2018-08-22
Cleanup after new virtual keyboard

persianpros:
* Delete vkey_all.png

* Delete vkey_clr.png

* Delete vkey_esc.png

* Delete vkey_ok.png

* Delete vkey_shift.png

* Delete vkey_shift_sel.png

* Update Makefile.am

* Update some pictures for new virtual keyboard


persianpros:
Only https://github.com/OpenPLi/enigma2-plugins/blob/master/unwetterinfo/src/plugin.py uses "vkey_esc.png", see: https://github.com/OpenPLi/enigma2-plugins/blob/master/unwetterinfo/src/plugin.py#L74

But I don't know why and this file didn't get any update since 2011 ...

PR #1698, opened 2018-08-27
Find and detect (skin_default) skin

fairbird:
If we put skin_default in image to make default skin then the (SCOPE_CURRENT_SKIN) can not detected the skin.
So this comment fix it.


littlesat:
Not accepted as we can do this smarter

PR #1704, opened 2018-08-28
Fixed returned path for linked /hdd

ims21:


ims21:
I will rebuilded it, but see Bug / Question in forum

PR #1707, opened 2018-08-30
[VirtualKeyBoard.py] Latest changes catch up

IanSav:
- [VirtualKeyBoard.py] Disconnect function name from button

    Naming a function based on one of the remote control buttons used to commonly activate the function is inappropriate.  This function can also be triggered by the ENTER key.

    The other aspect of this correction is to reinforce that it is the keymap and not the code that creates associations between actual buttons and functions.

- [VirtualKeyBoard.py] Add hard coded title

    This change addresses concerns from some users who don't want the title defined via the title attribute of the screen tag in the skin.  The reason is that it is easier to translate the string if it is in the code than if is in the skin.

### DIFF
--- a/lib/python/Screens/VirtualKeyBoard.py
+++ b/lib/python/Screens/VirtualKeyBoard.py
@@ -38,6 +38,7 @@ class VirtualKeyBoard(Screen, HelpableScreen):
 	def __init__(self, session, title=_("Virtual KeyBoard Text:"), text="", maxSize=False, visible_width=False, type=Input.TEXT, currPos=0, allMarked=False):
 		Screen.__init__(self, session)
 		HelpableScreen.__init__(self)
+		self.setTitle(_("Virtual keyboard"))
 		prompt = title  # Title should only be used for screen titles!
 		self.key_bg = LoadPixmap(path=resolveFilename(SCOPE_CURRENT_SKIN, "buttons/vkey_bg.png"))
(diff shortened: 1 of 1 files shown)


ims21:
Why is added hard coded title, when Tapat solved this problem [here ](https://github.com/OpenPLi/enigma2/commit/32ec6853aad1c8bbb9568fa77b07c2dc06a3f7b6)and it is not needed more ?

ims21:
https://github.com/OpenPLi/enigma2/commit/32ec6853aad1c8bbb9568fa77b07c2dc06a3f7b6

why hardcoded title is added now ?

IanSav:
Others were concerned that skin based titles may not be harvested by the translation system.  Given that skins come from all sorts of places I couldn't offer a workable solution to that problem.  Trust me that this was not a change I wanted to make.  I decided it was better to use the Python defined title rather than make other developer's lives more difficult.

ims21:
"Others were concerned" in ViX , yes ...

IanSav:
The concern was expressed by both OpenPLi and OpenViX.  It was an OpenViX person who made the compelling argument that not all skins are directly accessible in a single repository to have the translation data extracted.  OpenPLi seem to be more insular with their code and skins.  Other images tend to share code and skins between images and hence the sources of skins may vary significantly.

I never understood the OpenPLi concern was for remote skins.  This could be another example where meaning is lost in the language difficulties.  This is why I try the engage people in conversation so that the true meaning of their words can be better understood.

PR #1709, opened 2018-08-30
[VIRTUALKEYBOARD] Update documentation

IanSav:
Add a warning that the skin parameter to change the VirtualKeyBoard text colours MUST be specified as decimal numbers.

### DIFF
--- a/doc/VIRTUALKEYBOARD
+++ b/doc/VIRTUALKEYBOARD
@@ -2,6 +2,7 @@ A Guide to the VirtualKeyBoard Screen
 -------------------------------------
 
 Written by IanSav - 18-Aug-2018
+Updated by IanSav - 31-Aug-2018
 
 This document explains the changes and updates to the VirtualKeyBoard of 
 Enigma2.  The code is located in:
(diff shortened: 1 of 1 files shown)


Taapat:
And you again ignore others.
I changed title="Virtual Keyboard" to title="Virtual keyboard" because in enigma translation is only for "Virtual keyboard": https://github.com/OpenPLi/enigma2/commit/0fed1a3611e4731d71c57e2c25aaefc57d348f1f
Why do you change this and break translation?

P.S.
Have you ever heard about "git cola"?

IanSav:
The term "Virtual Keyboard" is a title not a text element.  I always use capital letters for the first letter of every word in a title.  This is also a way to allow translations to identify and separate titles from other uses of the phrase.

By the way, changing the code here is not changing the master copy of the code.  I post pull requests here as a service to the OpenPLi repository maintainers.

Taapat:
If you have created an example in doc file, this example should be correct so other people can use it.
In this case, the skin example is incorrect. If someone uses it, the title will not be translated.
I see that this PR is megred, so nobody not worried about the correct doc file.
I hope nobody not use your skin example in this doc file.

IanSav:
This section of the document is an example.  It s fully usable.  It is NOT authoritative instructions!  Feel free to write your skins in any way you want.

PR #1710, opened 2018-08-31
EventView.py: create epg_eventname for skins

Hains:
To make it possible to split event name and full description into seperate
widgets in skins.

This gives skin builders the possibility to change color, position, size, aligment and font
of event name without changing full description.

This requires skin changes but it will be still compatible when not.


Hains:
See:
https://github.com/littlesat/skin-PLiHD/pull/693

PR #1720, opened 2018-09-02
resolveFilename: check base path without skin name

Taapat:
This allows use the check to replace images from skin_default with images from skin, or use image from skin, if specified directly.


Taapat:
I agree.
It seems to me that there are too many checks here.
I have a feeling that it can be done more easily, but at this moment I do not see how.

PR #1722, opened 2018-09-03
fr.po file updated from develop at 2018-09-03 21:14

pr2git:


littlesat:
It seems this commit has conflicts

pr2git:
Yes it is normal I split all po, I should have wait for the all po commit before posting the fr.po one.

PR #1737, opened 2018-09-09
networksetup: small cosmetic change

Hains:
Replace spaces with tab and align in correct formatting.


athoik:
The else seems to nested different now! Under "if". Previously was under "for".

So i think it's not only cosmetic change, but a functional one.

Hains:
Yeah you're right.. a for else. I close PR.

PR #1741, opened 2018-09-13
[skin.py] Disallow string parameters

IanSav:
This change reduces the functionality of the previous commit but protects legacy code that only ever expects to see numeric parameters.  This change restores the skin processing "bad parameter" error when non numeric parameters are encountered.  (Note that valid integers, floats, hex values, colour codes and colour names will all resolve to numbers and not trigger the error log.

String parameters can be revisited when a need arises.


IanSav:
The Travis CI build appears to have failed.  Is this something I need to address?

PR #1748, opened 2018-09-30
AudioSelection - added debug messages for audio and subtitle tracks

ims21:


ims21:
Why not ? Where can you see audio/subtitle tracks info in 639 for current service ? Then developer must still install .py, add lines for print, when wants, what happens ?

ims21:
It is displayed on start AudioSelection screen only.

PR #1752, opened 2018-10-06
ScanSetup.py: Support BCM3466 dual DVB-T2

persianpros:
This is for new vuplus drivers (uno4k, uno4kse, ultimo4k)
New changes will be in the BSP layer soon.

### DIFF
--- a/lib/python/Screens/ScanSetup.py
+++ b/lib/python/Screens/ScanSetup.py
@@ -125,7 +125,10 @@ terrestrial_autoscan_nimtype = {
 'TT3L10' : 'tt3l10_t2_scan',
 'TURBO' : 'vuplus_turbo_t',
 'TT2L08' : 'tt2l08_t2_scan'
+'BCM3466' : 'bcm3466'
 }
+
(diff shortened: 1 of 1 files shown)


lib/python/Screens/ScanSetup.py
```diff
@@ -244,9 +247,9 @@ def GetCommand(nim_idx):
 				if nim_name is not None and nim_name != "":
 					device_id = ""
 					nim_name = nim_name.split(' ')[-1][4:-1]
-					if nim_name == 'TT3L10':
+					if nim_name in dualtuners:
```
persianpros, line 250: We could also do this without "dualtuners" list, for example:

if nim_name in ("TT3L10", "BCM3466"):

persianpros:
@littlesat 
Yes I mentioned it :)

We need the tuner name for:

`command = "%s %s" % (terrestrial_autoscan_nimtype[tunerName], device_id)`

I think this is better: https://github.com/OpenPLi/enigma2/pull/1753

PR #1760, opened 2018-10-08
Update de.po thanks angelito + standardize el.po & lt.po

pr2git:


WanWizard:
@pr2git where did this DE translation come from? It looks like not from someone who actually speaks German, translation is a mess, it looks like a poor Google Translate... 

Could you check with the person that provided you with the de.po file?

pr2git:
Hi,

Here is the source of it:
https://forums.openpli.org/topic/21479-translations-for-enigma-2-please-post-them-here/page-79#entry944998

I have no way to check the foreign language translations, Angelito also provides the es.po one.
You can revert the changes in the de.po file, to return to a valid situation, sorry about this I need to trust the proposed translations. :-(

Pr2

WanWizard:
I think we should only accept translations from someone that actually speaks the language, this angelito clearly doesn't, his german translations are so bad, clearly directly from Google Translate (he translated "Enigma" to "Riddle" in German :question:).

@betacentauri has already made an attempt to fix it.

pr2git:
Of course but when someone submit a translation on the forum we should trust him. There is no translation team with reliable translators, so every help is welcome. This is the first time that we experienced problem with translation submitted on the forum. I have no way on my side to check the translation quality. I send a PM to Angelito explaining that its translation was really poor and looks like google translate. I hope the es.po he also submit doesn't suffer from the same problem.

WanWizard:
I agree we need a better system for translations, ideally with one person responsible for each supported language, so you don't have to do any checking.

betacentauri:
Not everything was bad!
But some translations really look like a google translation. If it’s done for a currently not translated string it’s not so bad as changing a already present good translation.

PR #1763, opened 2018-10-10
System temperature (OpenATV Code) Thank you

fairbird:


athoik (changes requested):
Please check Components.About 

We get temperature in getCPUInfoString.

So this code is not acceptable.

PR #1764, opened 2018-10-11
[Picon] Fix for IPTV with buffering enabled

AbuBaniaz:
See thread here
https://forums.openpli.org/topic/41198-serviceapp-gstplayer-and-exteplayer3/page-41#entry945158

### DIFF
--- a/lib/python/Components/Renderer/Picon.py
+++ b/lib/python/Components/Renderer/Picon.py
@@ -79,6 +79,10 @@ def getPiconName(serviceName):
 		#fallback to 1 for TV services with non-standard service types
 		fields[2] = '1'
 		pngname = findPicon('_'.join(fields))
+	if not pngname and fields[9] != '0':
+		#fallback to 0 for iptv buffering
+		fields[9] = '0'
+		pngname = findPicon('_'.join(fields))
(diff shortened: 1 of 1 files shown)


athoik (approved):

littlesat:
Now it searches twice... so )when it’s a stream) make it always zero and just search once...

AbuBaniaz:
I don't know how to do that.

pr2git:
In fact just before your proposed change we already test it it is a stream so place your change there, no need to specifically test if the 9th field is not null since this parameter is only relevant for stream, so you suppress a useless if by doing this that way.

Ev0-BH:
Is this more exceptable ?

[code]
	if not pngname and fields[0] != '1':
		#fallback to 1 for IPTV streams and 9th field to 0 when buffering
		fields[0] = '1'
		fields[9] = '0'
		pngname = findPicon('_'.join(fields))
[/code]

Ev0-BH:
What needs to be changed then ?

What Abu posted, works fine.

Remember not all iptv is service 1:0:1, we also have 5002:0:1, 5001:0:1 and 4097:0:1

PR #1777, opened 2018-10-18
Add Support for Gamma Curve (aka SDR/HDR/HLG)

athoik:
This commit adds support for reporting current gamma curve to Enigma2.
It uses video event with id 17. The framerate property will have one
of the following possible gamma values:
0: Traditional gamma - SDR luminance range
1: Traditional gamma - HDR luminance range
2: SMPTE ST2084 (aka HDR10)
3: Hybrid Log-gamma

### DIFF
--- a/lib/dvb/decoder.cpp
+++ b/lib/dvb/decoder.cpp
@@ -245,7 +245,7 @@ int eDVBVideo::m_close_invalidates_attributes = -1;
 
 eDVBVideo::eDVBVideo(eDVBDemux *demux, int dev)
 	: m_demux(demux), m_dev(dev),
-	m_width(-1), m_height(-1), m_framerate(-1), m_aspect(-1), m_progressive(-1)
+	m_width(-1), m_height(-1), m_framerate(-1), m_aspect(-1), m_progressive(-1), m_gamma(-1)
 {
 	char filename[128];
 	sprintf(filename, "/dev/dvb/adapter%d/video%d", demux ? demux->adapter : 0, dev);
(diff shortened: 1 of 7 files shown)


nautilus7:
May I request to change "gamma" to "eotf" (in comments and perhaps in code too) as this is the proper naming now? In the old days where we only had SD material, the "gamma" term was used, but now the "eotf- electro optical transfer function" is the correct term. Alternative you can use "transfer function" or "transfer characteristics" like in mediainfo application.

Also option 2 should be renamed to 2: SMPTE ST2084 (aka PQ - Perceptual Quantizer)
HDR10 is a more broader term, made by the blu-ray disc forum, and include specs like bt.2020 color, uhd resolution, PQ eotf and more...

Furthermore, what are options `0: Traditional gamma - SDR luminance range` and `1: Traditional gamma - HDR luminance range`? It's not clear to me what these options correspond to...

Thanks for the nice addition anyway!

athoik:
It's cosmetic to rename variable to eotf, if commit accepted please fill free to create a PR.

Regarding the possible values, it's what the Broadcom driver knows.

nautilus7:
Yes, all are cosmetics, no doubt...

Regarding the values 0 and 1, any information (from the driver developers) what exactly the "traditional gamma" is?

athoik:
0: SDR, 1: HDR, 2:HDR10, 3:HLG

nautilus7:
Thanks.
I guess, 1 is some undefined/unknown HDR type and 2 is SMPTE ST2084 (aka PQ - not HDR10).

In wikipedia you can see HDR10 is different than SMPTE ST2084 (aka PQ)
https://en.wikipedia.org/wiki/High-dynamic-range_video#Standards

PR #1779, opened 2018-10-19
[frontend.cpp] Add tuner ID to debug

Huevos:
As suggested by Athoik


athoik:
It's perfect, it will help in the future when debugging with multi frontend receivers.

Maybe we can remove also the extra space?  (replace all   d %d] with d%d]).

See: https://github.com/OpenPLi/enigma2/blob/545449c5a5a213438f0d614de29561fa34a7b6cf/lib/dvb/decoder.cpp#L345

PR #1793, opened 2018-10-27
Update from fr.po

marcodream:
Hi, I have reworked the translation from english to french for all messages in fr.po.
I don't use the word "service" in french because it is not used in french, we usually use "chaîne", "canal" or "transpondeur". 
I have a dreambox DM7080HD and I would be interested to do the same work for this receiver. My OS version is 4.2.11r7 (2016-05-11)  and my installed image of Enigma2 is dated from 2014-10-14. 
Can you help me??
Can you send to me the fr.po of this receiver?
In advance thank you for your help

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -95,7 +95,7 @@ msgid ""
 "Scan for local extensions and install them."
 msgstr ""
 "\n"
-"Scanner les extensions locales et les installer."
+"Rechercher les extensions locales et les installer."
 
 msgid ""
 "\n"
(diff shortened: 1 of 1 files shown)


WanWizard:
We don't make software for dreamboxes, so no, we can't help.

@pr2git can you check this?

pr2git:
Service is not a real problem because it covers both TV and Radio. In french we speak of "chaîne TV" and "station de radio" so there is no real common term for both, so I keep "Service" to refer to both "chaîne or station". A transponder is different from a service! A transponder is a group of several services broadcasted on the same frequency. Moreover we often refer to "service reference" on the forum for exemple for EPG modification, so this is nice, I think, that french people understand what a "service" is the DVB world.
I know it is hard to have a perfect translation so I need to make some compromise between the technical terms and the french equivalent when translating.

I will review your changes and see which one I will integrate in the french translation.

PR #1801, opened 2018-10-31
[VirtualKeyBoard.py] Change "Save" to "Enter"

IanSav:
User feedback has requested that the GREEN button be changed from "Save" to "Enter".  This change makes the change to the button text and, for consistency, the code linked to the button.

### DIFF
--- a/lib/python/Screens/VirtualKeyBoard.py
+++ b/lib/python/Screens/VirtualKeyBoard.py
@@ -61,7 +61,7 @@ class VirtualKeyBoard(Screen, HelpableScreen):
 		self.keyBackgrounds = {
 			"EXIT": self.key_red_bg,
 			"OK": self.key_green_bg,
-			"SAVE": self.key_green_bg,
+			"ENTER": self.key_green_bg,
 			"LOC": self.key_yellow_bg,
 			"SHFT": self.key_blue_bg
 		}
@@ -323,7 +323,7 @@ class VirtualKeyBoard(Screen, HelpableScreen):
 
 		self["actions"] = HelpableNumberActionMap(self, "VirtualKeyBoardActions", {
 			"cancel": (self.cancel, _("Cancel any text changes and exit")),
-			"save": (self.save, _("Save any text changes and exit")),
+			"save": (self.enter, _("Enter text and exit")),
 			"locale": (self.localeMenu, _("Select the virtual keyboard locale from a menu")),
 			"shift": (self.shiftClicked, _("Select the virtual keyboard shifted character set")),
 			"select": (self.processSelect, _("Select the character or action under the virtual keyboard cursor")),
@@ -360,7 +360,7 @@ class VirtualKeyBoard(Screen, HelpableScreen):
 		self["language"] = Label(_("Language") + ": " + self.lang)
 		self["key_info"] = StaticText(_("INFO"))
 		self["key_red"] = StaticText(_("Exit"))
-		self["key_green"] = StaticText(_("Save"))
+		self["key_green"] = StaticText(_("Enter"))
 		self["key_yellow"] = StaticText(_("Select locale"))
 		self["key_blue"] = StaticText(self.shiftMsgs[1])
 		self["key_help"] = StaticText(_("HELP"))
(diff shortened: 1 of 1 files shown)


IanSav:
As I mentioned previously people shouldn't be groping into the middle of the code.  If there is a problem then I will insert capture entry points.  I like to keep the code as clean as possible.  I also want to keep the code as self documenting as possible.

I will add the alternate entry point just to be sure.  My apologies, I should have added it when I made the change.

littlesat:
Sometimes you can go too far... a remote has an ok button, no enter button... so my meaning is we should simply keep ok as it is good as it is.... no change required... has no added value and also not needed to make it more complicated than it is...

IanSav:
The request was specifically for me to change "Save" to "Enter".  The wording was discussed among a number of users and developers and it was decided that most non-technical users would better understand the work "Enter" in the context of a keyboard rather than "Save".  I don't know if "OK" was a candidate.  I am not a fan of "OK".

I used an image of the "Enter" button on the key grid but if the screen is adjusted to use text rather than the icon the the text that will be used is "Enter".  I accept that "Enter" is a better choice for the GREEN button as it does match the visual queue.

I would appreciate you considering that "Enter" will be more meaningful to more users.

littlesat:
Which devs.... note you can’t simply change these things and this devs thinks it really does not have any added value... sorry
Statement.. a remote has an ok button... not an enter button... so using enter is even more confusing...

IanSav:
As you should now note, all the VirtualKeyBoard and NumericalTextInput changes are now live in OpenViX.  The requests came from OpenViX, and I can see and agree with their point.

This pull request was to keep OpenPLi in the loop for code enhancements from downstream as was requested by OpenPLi developers.

IanSav:
The reluctance to accept this clean code submission that keeps the development code in sync is a complete surprise given that the hack botch up of the VirtualKeyBoard addition to TimerEntry.py went in with no obvious scrutiny.

Huevos:
Yep. This hack added in 2014.

https://github.com/OpenPLi/enigma2/commit/207fdd1403a400076d57f14b908126b792831582#diff-afe9bf64da6c0b3a320812ba8016ce77

IanSav:
Yes, but I was just asked what I broke in the new VirtualKeyBoard when this was the actual cause of the bug.

Huevos:
Open virtual keyboard to do an EPG search. Then you have to select "save" to do the search. Makes no sense whatsoever. Virtual keyboard does not "save" anything. It passes the entered text string back to the calling code.

The other option is pass the GREEN text in from the calling function. That way the text will always be suitable for the use.

By the way keyboards have an "enter" button not an "ok" button. This screen is mimicking a keyboard not a remote control. Also this is the GREEN button text. Nothing to do with an "ok" button on the remote control.

RobvanderDoes:
Hence 'enter', as every keyboard has...

nautilus7:
+1 for "enter".

littlesat:
It is NOT a keyboard it is a remote... make no sense this change. Pure technically it can only break stuff... and just because of silly wording.... please iansaf don’t try to change too much with risk of breaking things giving ‘us’ more work or making change of braking things....  the need to add a legacy fallback says enough... you cannot simply thy to change something that in the past was a bad desician.... please ok or save it is just in coding no user is seeing it and a remote is not a keyboard... on a remote it is Ok.  Not enter!!!

For this time I merge it... but please next time avoid these kind of 'useless' changes as they do not have any REAL added value.... The User never sees it and it has a risk of breaking this for just some 'cosmetic wordings'.... Where never someone is right because someone >10 years ago did not made an 100% optimal wording for a specific function....

IanSav:
Thank you for merging the code.

I was careful to not break existing code.  Please also note that this UI is to simulate a physical keyboard and NOT a remote control.  This code is also designed to work with a physically attached PC keyboard. 
 The terminology should reflect the keyboard style environment.

The reason for the code change was to keep the code and functionality matched and readable.  Not keeping the code clean and readable leads to misunderstanding and code bloat / rot.

RobvanderDoes:
The continuation of a lot of nonsense....
1- This is not a remote, but a keyboard, hence the name.
2- There is no keyboard in the world that uses a 'save'-button.
3- A keyboard doesn't 'save' text, it enters text.
All-in-all: a very useful change, as it makes the VK much more logical and hence more user-friendly.

Taapat:
You can change the text on the button, but why change the name of the function?
The okClicked function is already proof that your added functions for the deprecated legacy interface do not work as you imagined.

PR #1814, opened 2018-11-12
Update bg.po

marto74bg:


WanWizard:
no need to provide two PR's, rc changes are merged back into develop.

PR #1834, opened 2018-11-25
osmio4k: support multiboot and online flashing

athoik:
osmio4k uses by default mmcblk1 but FlashImage and Multiboot has
hardcoded values to mmcblk0.

Add the mmc devide used by machine to SystemInfo canMultiboot tuple.
Also add True/False to canMultiboot tuple, once true it will copy
the startup slot, like Gigablue does.


athoik:
> Then line 6 I have doubts... I think we should avoid the +'p' and use "%s%s" % (...., 'p') instead.

Should be fixed via  e2f4cf4

eriksl:
I also think this is better suited for develop.

littlesat:
“%sp” you even make it better!

littlesat:
Sorry I can not merge this as the merge request is agains RC... Please offer a new merge request on Develop.

PR #1835, opened 2018-11-25
fix convertDVBUTF8 error from pull request #1829.

nx111:


lib/base/estring.cpp
```diff
@@ -596,15 +622,24 @@ std::string convertDVBUTF8(const unsigned char *data, int len, int table, int ts
...
-	if (pconvertedLen && *pconvertedLen < len)
-		eDebug("[convertDVBUTF8] %d chars converted, and %d chars left..", *pconvertedLen, len-*pconvertedLen);
+	if (pconvertedLen)
+		*pconvertedLen = convertedLen;
+
+	if (convertedLen < len)
+		eDebug("[convertDVBUTF8] %d chars converted, and %d chars left..", convertedLen, len-convertedLen);
 	//eDebug("[convertDVBUTF8] table=0x%02X twochar=%d output:%s\n", table, useTwoCharMapping, output.c_str());
+
+	if (debugLvl >= 3)
```
athoik: Why debugLvl >= 3 required here? The eDebug already requires debugLv = 4. So what's the point of that if?

littlesat: Is indeed not required in openpli...

littlesat:
Can you please explain the real need for this?

nx111:
Outputing verbose info is very useful when  we need to get more information to resolve string display errors. But   it's not need default because too much infomation.  I want it working just when user set debug level to higher.   

Just use eDebug directly ok. default debugLvl = 3, when change it to 4, eDebug will output message.

littlesat:
Can you please explain the real need for this?... the whole patch I mean. 
- What was the problem it solves?

nx111:
These changes can resolve some satellites display string issue. Such as 88E, the first char is 0x15,but it is encoded by utf16be, by my changes and add config in encoding.conf, it can show string correctly.

littlesat:
The last time such kind of patch (a few days ago) had some side effects... are you sure the side effects are illiminated in this merge request?

nx111:
tested in openatv.

PR #1841, opened 2018-11-27
ChoiceBox: add sanity checks for list

Taapat:
Fix GSOD if list is empty or list value is empty.
https://forums.openpli.org/topic/63893-openpli-7-news-and-information/page-10#entry972505


littlesat:
Can indeed be added as it Will avoid gsod on other places... the ‘trigger’ Mediaplayer did already solve it in a different way as showing an empty playlist list is ugly.

Taapat:
I agree that showing a empty playlist is ugly and in any case fix in Mediaplayer needed.
But this should fix gsod also in any other unexpected case.

PR #1846, opened 2018-11-29
support fallback font to display east asia language

nx111:
if  fallback.font exists, and font code not exits in the current font and
the replacement font, try get glyph from the fallback font.


nx111:
This is best compatible with other languages and does not need to  change  skins.
the fallback.font is a symbolic link to the real font.

otherwise, for chinses,kerean,japanese, need to change skins or replace the existing fonts.
so  can't implement one firmware  works in all languages.

PR #1849, opened 2018-12-02
Reduce convertDVBUTF8 debug in log

Taapat:
Before this half of the log from enigma consists of [convertDVBUTF8] debug rows and it is difficult to review the log.


Taapat:
I agree, use eLog(6, ) is better: https://github.com/OpenPLi/enigma2/pull/1850

Taapat:
It seems that athoik is right.
Everything with log level greater than 4 is removed with compiler: https://github.com/OpenPLi/enigma2/blob/develop/lib/base/eerror.h#L127-L139

PR #1850, opened 2018-12-02
Reduce convertDVBUTF8 debug in log

Taapat:
Use eLog with debug level 6 instead of eDebug.
Before this half of the log from enigma consists of [convertDVBUTF8] debug rows and it is difficult to review the log.
Now you can use lower debug level to disable convertDVBUTF8 messages in log.


littlesat:
Thanks!

PR #1853, opened 2018-12-04
Network: add p2p0 to isBlacklisted devices

athoik:


littlesat:
Why should this be in rc?

PR #1855, opened 2018-12-06
[BoundFunction.py] Pass copy of kwargs in boundFunction calls

IanSav:
This error detection and its correction was supplied by Prl.

### DIFF
--- a/lib/python/Tools/BoundFunction.py
+++ b/lib/python/Tools/BoundFunction.py
@@ -4,6 +4,6 @@ class boundFunction:
 		self.args = args
 		self.kwargs = kwargs
 	def __call__(self, *args, **kwargs):
-		newkwargs = self.kwargs
+		newkwargs = dict(self.kwargs)
 		newkwargs.update(kwargs)
 		return self.fnc(*self.args + args, **newkwargs)


Huevos:
Just to explain this commit...........

Tools.BoundFunction is only 9 lines and it's pretty easy to understand.

And yet, in those short 9 lines, someone found space for a bug (seemingly not knowing the difference between copying a reference and copying an object).

Run the following in enigma2/lib/python:

```
from Tools.BoundFunction import boundFunction

def func(arg1="defaultArg1", arg2="defaultArg2", arg3="defaultArg3"):
	print arg1, arg2, arg3

bf = boundFunction(func, arg1="boundArg1")

bf(arg2="test1Arg2", arg3="test1Arg3")

bf(arg2="test2Arg2")
print "oops! Should be 'boundArg1 test2Arg2 defaultArg3'"

print

class fixedBoundFunction:
	def __init__(self, fnc, *args, **kwargs):
		self.fnc = fnc
		self.args = args
		self.kwargs = kwargs
	def __call__(self, *args, **kwargs):
		newkwargs = dict(self.kwargs)
		newkwargs.update(kwargs)
		return self.fnc(*self.args + args, **newkwargs)

bf = fixedBoundFunction(func, arg1="boundArg1")

bf(arg2="test1Arg2", arg3="test1Arg3")

bf(arg2="test2Arg2")
print "That's better"
```

That prints the following:

> root@osninopro:/usr/lib/enigma2/python# python boundFunctionBug.py
> boundArg1 test1Arg2 test1Arg3
> boundArg1 test2Arg2 test1Arg3
> oops! Should be 'boundArg1 test2Arg2 defaultArg3'
> 
> boundArg1 test1Arg2 test1Arg3
> boundArg1 test2Arg2 defaultArg3
> That's better
> root@osninopro:/usr/lib/enigma2/python# python boundFunctionBug.py

littlesat:
I really don't understand the need as **kwargs should be definition be a dict...? 
How is an issue triggered? (faulty plugin?)... what exactly does it solve?

prl001:
The names of the pull request and the commit are perhaps misleading. The fix isn't to allow the `**kwargs` argument to `boundFunction.__init__()` to be something other than a dict (e.g. an iterable in the right form, for example). As littlesat says, `**kwargs` in both `__init__()` and `__call__()` will always be a `dict`.

A clearer title might have been something like "Pass copy of kwargs in boundFunction calls".

The problem is that calls of the bound function modify the bound function's `self.kwargs`. The original code of `__call__()` does:
```
		newkwargs = self.kwargs
		newkwargs.update(kwargs)
```
This means that when `newkwargs.update(kwargs)` is called, it _also_ adds the call's keyword arguments to `self.kwargs`, since `newkwargs` and `self.kwargs` refer to the same object instance.

Using `newkwargs = dict(self.kwargs)` makes a shallow copy of `self.kwargs`, so that `newkwargs.update(kwargs)` adds the call's keyword arguments to a local _copy_ of `self.kwargs` and doesn't change `self.kwargs`

The use of `dict()` here is to make a copy, not to do any type conversion.

An alternative correct implementation that's closer to the way that positional arguments are handled is:
```
def __call__(self, *args, **kwargs):
	return self.fnc(*self.args + args, **dict(self.kwargs.items() + kwargs.items()))
```
It's probably more Pythonic to do it that way, but I didn't think of it at the time. In this case, dict() is doing both a type conversion, from iterable to `dict`, and making a new `dict` instance that contains both the bound and caller-supplied keyword arguments.

PR #1857, opened 2018-12-06
Greek translation update 7/12/2018

MCelliotG:


MCelliotG:
Problem with empty strings generated after submission to github. This is new.

PR #1858, opened 2018-12-07
Greek translation update 7/12/2018

MCelliotG:
Hope no empty strings this time, this is the first time this ever happened!


littlesat:
I’m afraid this one does not ‘fit’ anymore due to a previous patch...

MCelliotG:
> 
> 
> I’m afraid this one does not ‘fit’ anymore due to a previous patch...

What does not fit exactly? I used the latest el.po generated yesterday. Please tell me what should I use, so we can solve this.

PR #1882, opened 2018-12-16
Update ServiceInfo.py

jpuigs:
Add Orbital Position according to namespace
https://devtools.openpli.org/issues/435

### DIFF
--- a/lib/python/Screens/ServiceInfo.py
+++ b/lib/python/Screens/ServiceInfo.py
@@ -188,8 +188,8 @@ class ServiceInfo(Screen):
 				self.subList = self.getSubtitleList()
 				self.togglePIDButton()
 				trackList = self.getTrackList()
-				fillList = fillList + ([(_("Namespace"), self.getServiceInfoValue(iServiceInformation.sNamespace), TYPE_VALUE_HEX, 8),
-					(_("TSID"), self.getServiceInfoValue(iServiceInformation.sTSID), TYPE_VALUE_HEX_DEC, 4),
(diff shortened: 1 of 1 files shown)


littlesat:
Why do we need a nsOrb variable here?

jpuigs:
fillList = fillList + nsOrb
I could do it without, yes, but as all appends to fillList  done by external functions are done with variables (sublist, tracklist), I did it as well.

ims21:
and why not in HEX ?

jpuigs:
Because the final result has to be decimal (º degrees), so Namespace's 4 first char. have to be converted anytime to decimal.
So I do it initally.

What's the difference between … ??
if posi == 61166  and 
if posi == EEEE

ims21:
better readability !!!  If user see EEEE, is clear for first look, that it is DVB-T !

jpuigs:
Yes but this value is not show on screen,so users don't see it

pr2git:
I agree that using 0xEEEE and 0xFFFF will make the code more readable because viewing the hexa is self explanatory because sorry I don't know by heart that 0xEEEE = 61166 and it is expressed in hexa in the namespace.

The "E" (East) and "W" (West) should be defined to be translatable (if possible).

jpuigs:
Ok. I can change it.
I'm not at home. Later this night I'll do it.
How can I edit this request ? Or do I have to do a new one ?

jpuigs:
I close this request and open a new one.

PR #1883, opened 2018-12-16
all po files updated from develop at 2018-12-16 20:38

pr2git:


Beaving:
I have a question regarding those translations. I was updating the German translation the other day and noticed a lot of weird stuff that is even present in the English translation itself. Like what you just committed:

#, fuzzy
msgid "in menu and plugins"
msgstr "Download plugins"
 #, fuzzy
msgid "in menu only"
msgstr "Main menu"
 #, fuzzy
msgid "in plugins only"
msgstr "This plugin is installed."


The msgstr sometimes have a completely different meaning from the msgid. Is this a relict from old menu structure or where does this come from, and based on what should we translate?

pr2git:
You didn't mention that there is a tag:

#,fuzzy

This tag is added when re-generating the po files on string that no longer properly match there translation.
So all #,fuzzy must be reviewed and corrected.

The fuzzy string are clearly highlighted in PoEdit software. If the translation marked as fuzzy is correct you can also easily switch off the fuzzy flag with PoEdit.

Beaving:
Ok, thanks. I translated it by hand before. What would be the correct translation here? Is "This plugin is installed." correct, or would it be "In plugins only" ?

pr2git:
Better to use PoEdit to translate this will avoid you to screw up the po file. But please go into the Preference - Advanced in PoEdit and disable the wrap line at 79 and tick keep original file format.

The string is msgid is always the latest one. It is coming from the source code founded by the script that generate the updated po file.  
The string in msgstr is the translation that has became fuzzy.

PR #1884, opened 2018-12-16
Add Orbital Position according to namespace on ServiceInfo

jpuigs:
Add Orbital Position according to namespace 
https://devtools.openpli.org/issues/435


jpuigs:
Changed code

PR #1900, opened 2018-12-18
- Make skin xmls to be searched by default in current skin directory

DimitarCC:
@littlesat Try this one. I hope now will be useful for you.
Sometimes github really frustrates me.


littlesat:
Exactly ;)

PR #1912, opened 2018-12-23
Fix creation of ap/sc files for channels using h265

betacentauri:
In this commit the streamtype for h265 was changed from 6 to 7:
https://github.com/OpenPLi/enigma2/commit/9dde0228ca1d85ccb7c7fc9ed2af9ebd08b5c8f4


athoik:
Great! 👍  (and sorry)

Maybe we can avoid harcoded value with eDVBPMTParser::videoStream::vtH265_HEVC instead?

PR #1914, opened 2018-12-23
pvrparse: use video enumeration from eDVBVideo class

athoik:
This will prevent unwanted fails with hardcoded values in the future.

### DIFF
--- a/lib/dvb/decoder.h
+++ b/lib/dvb/decoder.h
@@ -40,7 +40,7 @@ private:
 	int m_width, m_height, m_framerate, m_aspect, m_progressive, m_gamma;
 	static int readApiSize(int fd, int &xres, int &yres, int &aspect);
 public:
-	enum { MPEG2, MPEG4_H264, VC1 = 3, MPEG4_Part2, VC1_SM, MPEG1, H265_HEVC, AVS = 16 };
+	enum { UNKNOWN = -1, MPEG2, MPEG4_H264, VC1 = 3, MPEG4_Part2, VC1_SM, MPEG1, H265_HEVC, AVS = 16 };
 	eDVBVideo(eDVBDemux *demux, int dev);
 	void stop();
 	int startPid(int pid, int type=MPEG2);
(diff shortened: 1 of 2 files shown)


betacentauri:
I only don’t like this “m_streamtype < eDVBVideo::MPEG2” very much. What is smaller than mpeg2? With the “unknown” comment it’s understandable. Maybe use “m_streamtype == -1”, but this also not perfect. 
Let’s see what the others think. 
Thanks for looking into it!

WanWizard:
This is above my "paygrade", I don't have the knowledge to review this.

eriksl:
Agree with betacentauri. A set does conceptually not have an order, so comparing numerically does not have a semantic value. Even more, the contents of the set should not even be treated as having numerical representations, that's just a shortcoming of C(++) ;-)

eriksl:
I see a structure like this:

**if(streamtype < ...::mpeg2) // invalid**
{
    ....
}

switch(streamtype)
{
    case(...):
    ...
    case(...):
}

Why not rewrite this into something like:

switch(streamtype)
{
    case(...):
    ....
    case(...):
    ...
    case(...):
    ...
    **default:**
}

You can even have gcc warn or err on switches missing case statements for certain enum values.

pieterg:
note: the 'if () //invalid' is not part of the switch, it is applied before the switch to fill in the codec id if it is is still unknown at that point. So I'd not move that to a default clause, to maintain readability.

> You can even have gcc warn or err on switches missing case statements for certain enum values.

Indeed, that would help to make sure each newly added codec in the enum gets added into the switch as well.
But only if you omit the default clause.
(so we might have to get rid of that, it doesn't do anything except logging an error at the moment)

eriksl:
Even better, +1.

betacentauri:
Looks ok for me now.

WanWizard:
So this can be merged?

athoik:
Let's wait another ok from @eriksl or @pieterg.

Then squash and merge!

eriksl:
I agree, let's wait for @pieterg

pieterg:
Looks good to me

eriksl:
Nice, thx all!

PR #1932, opened 2018-12-29
libopen: remove hardcoded patch to libc

athoik:
When using multilib the 64bit libc located under /lib64 folder.
This commit removes the hardcoded path to libc, it will be resolved
automatically to the proper location.

### DIFF
--- a/tools/libopen.c
+++ b/tools/libopen.c
@@ -17,7 +17,7 @@ int open64(const char *pathname, int flags, ...)
 	{
 		void *handle;
 		char *error;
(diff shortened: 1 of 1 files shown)


betacentauri:
It‘s path and not patch ;)
According to the man pages absolute path is not needed. So it seems ok for me, but I haven’t tested it.

athoik:
>It‘s path and not patch ;)

Sure it's path, damn typos!

WanWizard:
fixed typo ;-)

PR #1937, opened 2019-01-01
scan: remove CRLF from channel and provider name

athoik:
It seems some channels on 5E include CRLF on channel name.

[eDVBScan] SID 27ce: is scrambled!
[eDVBScan]   name 'Viasat History HD
', provider_name 'AMB'
[eDVBScan]   descr<5d>

Resulting in lamedb corruption, since we expect all data in one line.

This commit replaces CRLF with an empty string to fix that issue.

### DIFF
--- a/lib/base/estring.cpp
+++ b/lib/base/estring.cpp
@@ -1,4 +1,5 @@
 #include <algorithm>
+#include <regex>
 #include <cctype>
 #include <climits>
 #include <string>
@@ -868,4 +869,12 @@ std::string urlDecode(const std::string &s)
 		}
 	}
 	return res;
+}
+
+std::string strip_non_graph(std::string &s)
+{
+	s = std::regex_replace(s, std::regex("[[^:graph:]]"), " ");
+	s = std::regex_replace(s, std::regex("\\s{2,}"), " ");
+	s = std::regex_replace(s, std::regex("^\\s+|\\s+$"), "");
+	return s;
 }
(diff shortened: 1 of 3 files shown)


WanWizard:
My 2cts:

I would prefer a general solution, and replace "\r\n", "\r" and "\n" (in this order) by a space, and then trim the result. So "A\r\nB" becomes "A B" and not "AB".:

littlesat:
The provider makes here the mistake by copy and paste stuff with a Windows pc as the /r/n is at the end... we try to work-a-round here... then better work a round to by just strip the string (remove white spaces and other stuff) no spaces or \n \r \t at front and at the end.... and call it clean service name ‘patch’

athoik:
The most generic approach I can come with is the following: 

```
#include <iostream>
#include <string>
#include <regex>

std::string cleanNonGraph(std::string s)
{
        s = std::regex_replace(s, std::regex("[[^:graph:]]"), " ");
        s = std::regex_replace(s, std::regex("\\s{2,}"), " ");
        s = std::regex_replace(s, std::regex("^\\s+|\\s+$"), "");
        return s;
}

int main()
{
        std::string tmp("   |  test    a     test\r\r\t\t\r\n\n\n  a teare |      Ελληνικά;        <  <  < ");

        std::cout << tmp << std::endl;

        tmp = cleanNonGraph(tmp);

        std::cout << tmp << std::endl;

        return 0;

}
```

```
# g++ test_re.cpp && ./a.out
   |  test    a     test


  a teare |      Ελληνικά;        <  <  <
| test a test a teare | Ελληνικά; < < <
```


I can push function in estring class. I am not sure about the name :-)

littlesat:
I would call it strip.... or clean or so...

PR #1960, opened 2019-01-14
enigma2: control debug level using tmp file

athoik:
Store next debug level on /tmp/enigma_debug_lvl and once gui is respawn "source" the file to get the ENIGMA_DEBUG_LVL.

This makes enigma2 main loop in script not required any more. The inittab will handle the respawn of gui as we used to.

### DIFF
--- a/tools/enigma2.sh.in
+++ b/tools/enigma2.sh.in
@@ -24,6 +24,11 @@ if [ -d /home/root ]; then
 	cd
 fi
 
+# debug file exists?
+if [ -s /tmp/enigma_debug_lvl ]; then
+	source /tmp/enigma_debug_lvl
+fi
+
 # any debuglevel passed?
 if [ -z $ENIGMA_DEBUG_LVL ]; then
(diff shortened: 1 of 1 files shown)


littlesat:
We don’t wanna do it that way... please hold on! I also do not want to continously write a file with tie, but let e2 also create a log when you have a spinner and on demand.

athoik:
En-light me how do you want to do it then?

littlesat:
We’re still discussing... First at least the loop should stay. I think exit it when debug_lvl <= 4.... and no tie... but something else instead....

athoik:
Control the debug level from UI? 
Change the debug level in real time without restarting?
Use syslog?

littlesat:
Control the debug level from UI? -> When it can be done?
But also write a 'spinner log' when we have a spinner and/or write a log on demand....!

Then this whole thing can be reverted

littlesat:
Wanwizard did solve this and also improved the UI.

PR #1974, opened 2019-01-22
Record timer patch

Huevos:
Move self.saveTimer() code back to RecordTimer.py.

Create self.cleanup() and self.cleanupDaily() methods in RecordTimer.py to save to the xml file on clean up.


littlesat:
1000 thanks

PR #1980, opened 2019-01-26
Update .travis.yml

persianpros:
No need to check with old GCC as you're not working on Open PLi 4 anymore.


WanWizard:
@athoik, ok with you ?

athoik:
I would prefer adding gcc-8 first.

Removing older gcc won't harm.

persianpros:
Then you could merge this and do that later ;)
After this all checks will be faster for sure ...
You could remove "release 4" branches from enigma2 and your oe too as people are using them for clones only.

WanWizard:
removing a branch is pointless, the commits themselfs won't disappear.

PR #1993, opened 2019-02-01
fix gsod for satfinder

v8droide2:


athoik:
The scan type can be multistream and or t2mi.

As @Huevos mentioned on forum, the t2mi should not touch existing multistream code.

lib/python/Components/SystemInfo.py
```diff
@@ -63,6 +63,7 @@ def hassoftcaminstalled():
 SystemInfo["3DMode"] = fileCheck("/proc/stb/fb/3dmode") or fileCheck("/proc/stb/fb/primary/3d")
 SystemInfo["3DZNorm"] = fileCheck("/proc/stb/fb/znorm") or fileCheck("/proc/stb/fb/primary/zoffset")
 SystemInfo["Blindscan_t2_available"] = fileCheck("/proc/stb/info/vumodel")
+SystemInfo["canT2MI"] = HardwareInfo().get_device_model() not in ("demo")
```
athoik, line 66: I guess demo it's just a dummy value?

lib/python/Screens/ScanSetup.py
```diff
@@ -719,20 +719,27 @@ def createSetup(self):
 					self.list.append(self.modulationEntry)
 					self.list.append(getConfigListEntry(_('Roll-off'), self.scan_sat.rolloff))
 					self.list.append(getConfigListEntry(_('Pilot'), self.scan_sat.pilot))
-					if nim.isMultistream():
+					if SystemInfo["canT2MI"] or nim.isMultistream():
```
athoik, line 722: How t2mi related with multistream? 

A TP can be multistream with or without PLS. 

A TP is possible to contain a PID that carries T2MI data. 

So T2MI is not multistream. TP is multistream. 

T2MI located on a PID on a TP.

lib/python/Screens/ScanSetup.py
```diff
@@ -860,28 +867,25 @@ def newConfig(self):
 			(self.modulationEntry and self.systemEntry[1].value == eDVBFrontendParametersSatellite.System_DVB_S2 and cur == self.modulationEntry):
 			self.createSetup()
 		elif cur == self.typeEntry:
-			if self.scan_sat.type.value == "multistream":
+			if self.scan_sat.type.value == "multistream" or "t2mi":
```
athoik, line 870: Did you mean `if self.scan_sat.type.value in ("multistream", "t2mi"):` or `if self.scan_sat.type.value == "multistream" or self.scan_sat.type.value == "t2mi":` here?

athoik (changes requested):

Huevos:
The code that was committed should not touch the multistream code. That part of the code should be reverted.

Also a notifier was used for the memory when this is not necessary. The same pattern as the multistream code should have been used. No need to reinvent the wheel.

Here is the code that does the T2MI memory code exactly the same as the Multistream memory code (without the notifier).
https://github.com/Huevos/enigma2/commit/6c21bccb6dff4c71b39d8bb1e288fbdb33f18a18

v8droide2:
The transponder/lamedb should have t2mi_pid and t2mi_plp enlargement.
The t2mi_pid cannot been capture yet from lyngsat or flysat but it's very common pid is 4096, so the transponder grabber should if the tranpsonder is t2-mi set it on 4096. I know it's not in all cases true but for now we have to work with the info which is provide. 

If the t2mi_pid is not 0 then automatic de-encapsulation should start I think this is prefer in userspace (a bit similar as astra-sm does but it seems some companies already choosen for to do it in the drivers.

I understand that T2-MI doesn't mean it's always multistream but I was not aware yet that non DVB-S2 multistreams supported hardware could handle as well PLP_ID

lib/python/Screens/ScanSetup.py
v8droide2: I assumed that only hardware which support multistream could handle plp in dvb-s2 mode.

The relation with multistream is the lenght of the transponder as if it has t2-mi but not multistream the transponder should set for multistream part as 0:0:0
DVB-S2 example
s 12188000:30000000:0:3:48:2:0:1:2:0:2
DVB-S2 with Multistream
s 12188000:30000000:0:3:48:2:0:1:2:0:2:20:0:1
DVB-S2 with only t2-mi
s 12188000:30000000:0:3:48:2:0:1:2:0:2:0:0:0:4096:0
DVB-S2 with Multistream and t2-mi
s 12188000:30000000:0:3:48:2:0:1:2:0:2:20:0:0:4096:0

lib/python/Components/SystemInfo.py
v8droide2: t2mi has no standard of detection, and probably it will never come either.
So probably the best solution is with a proc entry.
Actually T2-MI check is for if the hardware support DVB-T2 PLP and not the de-encapsulation shouldn't been a restriction at all.

This is a dummy and should been replaced in future by an entry, so that the option will only been available on boxes which support PLP in DVB-S2

Huevos:
Can we start again with the original commit by Athoik and then add the T2MI PID stuff so it does not interfere with multistream. Maybe it is a little bit of work now but at least this current spaghetti code would be gone.

Needs to be done a commit at a time, not this whole big commit (that attributes the wrong author). 

Also, what is going to happen about saving the PID in lamedb?

lib/python/Components/SystemInfo.py
athoik: Please see: ETSI 300 468, 6.4.13. 

T2MI_descriptor is how it should be announced.

lib/python/Screens/ScanSetup.py
Huevos: The multistream parameters only appear in lamedb if the IS/PLS values are not the default. How are you thinking of saving the t2mi parameters across reboots?

athoik:
Hi guys,

I think we have taken a wrong path, a path that seems to cause many troubles. 

Let's start over, properly this time!

The original work made for AVL tuners that support T2MI committed on a new branch [t2mi-development](https://github.com/OpenPLi/enigma2/tree/t2mi-development)

@Huevos, can you please push your work on this branch? We need the commits for the UI that you previously made on top of my work.

@v8droide2, can you please gather all PID where T2MI currently works? We can send that list on lyngsat in order to add them on their site. If that happens then satellites.xml can store the extra info t2mi_pid="4096" (or what ever the pid is).

In order to coordinate the development required for T2MI I suggest opening a ticket on gitub (so all interesting can participate) or switching to new infastructure from PLI

@WanWizard what do you think?

Huevos:
@Athoik, PR here: https://github.com/OpenPLi/enigma2/pull/1996

I'll add the indent commit if/when it gets accepted into the main "develop" branch.

Huevos:
Hi, Littlesat. I will reply here as it concerns this PR. 

There is no point merging this because the project has been started again from scratch on a different branch. 

https://github.com/OpenPLi/enigma2/tree/t2mi-development

athoik:
And we are waiting for your actions too!

Are you in?

Huevos:
I'll help with whatever I can.

athoik:
A very crucial task to get the approach with PID and PLP is to have the PID.

So a list should be created and forwarded to lyngsat. I am sure people maintain lyngsat, don't mind adding one more really required information to their website. 

Sending them a list with the extra info required on T2MI enabled TP's is what required.

Anyone willing to handle it? 

I was hoping @v8droide2 could handle it, since he is very warm to support this project.

littlesat:
Please let me also clearly know which multistream related commits we can push

Huevos:
@Littlesat, T2MI has nothing whatsoever to do with multistream, so please don't accept any T2MI commits that touch multistream code.

PR #1997, opened 2019-02-03
[ScanSetup/Satfinder] group multistream ConfigElements with an indent

Huevos:
Affected fields: is_id, pls_mode, pls_code

### DIFF
--- a/lib/python/Plugins/SystemPlugins/Satfinder/plugin.py
+++ b/lib/python/Plugins/SystemPlugins/Satfinder/plugin.py
@@ -124,6 +124,7 @@ class Satfinder(ScanSetup, ServiceScan):
 
 	def createSetup(self):
 		self.list = []
+		indent = "- "
 		self.satfinderTunerEntry = getConfigListEntry(_("Tuner"), self.satfinder_scan_nims)
 		self.list.append(self.satfinderTunerEntry)
 		self.DVB_type = self.nim_type_dict[int(self.satfinder_scan_nims.value)]["selection"]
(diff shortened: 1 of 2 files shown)


Taapat:
Why for this is necessary a variable?
Why not simply use text in getConfigListEntry?

Huevos:
Because hard coding is crap. If you want to change the style of the indent you change it only in one place, not in three. Hard coding just leads to bugs.

That is why we have things like "eDVBFrontendParametersSatellite.No_Stream_Id_Filter" instead of writing "-1" everywhere.

littlesat:
I do not think this is really nessesary to ‘highlight’ these parameters... then better add something to make them viewable and hide it... highlight with ‘- ‘ is ‘ugly’

Huevos:
Littlesat this is exactly how it is done in Satconfig.py for the additional motor options. They are hidden when disabled and indented when enabled. If you don't like the style of the indent use a double space instead. But the indent makes it clear they are a group of subordinate fields.

Huevos:
And this is what "extra motor options" looks like that is already in the image for years.

RobvanderDoes:
But that never worked satisfactory. Once you leave that setup, with extra motor options activated, those menu entries are disabled (ie invisible) when you enter the setup again.

Huevos:
Rob, multistream is different. When the "transport stream type" is "ordinary" the default IS/PLS values are used. Look at the code and you will see.

In the case of "extra motor options" it was just a workaround to hide parameters from the menu that no "normal" user would ever adjust. That was done when "Rhinoceros" created those values for use in the positioner plugin, and rather than having them adjustable from the plugin they were dumped in Satconfig.py.

Anyway, that has nothing to do with whether an indent would be useful for a user to identify a number of fields as a group.

littlesat:
Then use spaces instead and not '- ' please....
And please let me know if I can merge this https://github.com/OpenPLi/enigma2/pull/1993 also?

Huevos:
Ok. I will change this and do again so it follows the current code in satconfig.py

I've answered the other question in the other PR.

Huevos:
Closed...

New request here:
https://github.com/OpenPLi/enigma2/pull/1998/commits/5d8d42d38276eaa213aa616412e8d99118aae6c6

PR #1998, opened 2019-02-04
[ScanSetup/Satfinder] group multistream ConfigElements with an indent

Huevos:
Affected fields: is_id, pls_mode, pls_code

Indent and code is identical to the "extra motor options" in Satconfig.py.


Huevos:
Ok, I will use the indent variable and the format you suggest. OK?

littlesat:
Yep please...

I was planning to use the same mechanism for FullMultiType tuners... (tuners that can do DVB-S2X/T2/C 'together' without a hard switch)... there where you program/config those

PR #2000, opened 2019-02-04
Fixed ConfigText for alternative_imagefeed

ims21:
- when is not set fixed size as False, then is entered char in Numeric input keyboard replaced with next char.


ims21:
About spaces before and after ... I keep using it as is used in whole file.

ims21:
btw - spaces removed

PR #2013, opened 2019-02-14
Update fi.po translation

jmuuriai:
A large(ish) update to the Finnish translation after 5+ years of no updates.

Baseline update merges common strings from the OpenVIX project, which has a much better maintained Finnish translation. Additional OpenPLI specific translations by myself.

### DIFF
--- a/po/fi.po
+++ b/po/fi.po
@@ -5,15 +5,15 @@ msgstr ""
 "Project-Id-Version: enigma2\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2019-02-05 11:25+0330\n"
-"PO-Revision-Date: 2013-10-15 19:32+0200\n"
-"Last-Translator: BoxFreak\n"
-"Language-Team: none\n"
+"PO-Revision-Date: 2018-08-25 10:57+0200\n"
+"Last-Translator: Kojo\n"
+"Language-Team: AndyBlac/timoj/Kojo/Samzam\n"
 "Language: fi\n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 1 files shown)


WanWizard:
You can't just copy a translation file from a different project? It will not be complete, and it will add VIX specific translations to OpenPLi.

WanWizard:
@pr2git could you check if this is mergeable?

jmuuriai:
That is absolutely correct WanWizard. You should not do that. And I did not.

I only merged the strings from OpenVIX that had an exact match to strings in the OpenPLI translation file. This was easy using a combination of msgmerge and msgcat. The result is the first commit in this PR, which contains the original OpenPLI strings, with updated translations where one was found. Strings from VIX which did not match anything in the OpenPLI file were not copied over.

After that, the rest of the translation was business-as-usual - manual.

pr2git:
I had a quick look into it, this seems to have been done properly. I think that you can merge it. Then I can rerun a full po generation and see how fi.po become after this.

WanWizard:
Thanks, merged.

jmuuriai:
Thanks @WanWizard ! Is the process of how commits move from develop onto releases documented somewhere? I'm new to this project, and was wondering whether I need to make further pull requests onto release branches, or will the commits get merged by the core members.

WanWizard:
Commits don't move. 

Releases are made by branching off develop into rc, which is publicly tested, and when ok, it will be branched off into release (and the rc branch will be deleted).

So if develop is up to date, release will be too.

In general we don't accept PR's on release branches, with some exceptions for non-PLi-code related data, such as channel lists, picon packages, the odd 3rd-party plugin update, etc.

PR #2018, opened 2019-02-19
[Standby] check disk  sleepmode only reboot/shutdown

Dima73:
-When not sleepmode set MessageBox timeout -->30 sec. and default for
reboot/shutdown --> Yes

### DIFF
--- a/lib/python/Screens/Standby.py
+++ b/lib/python/Screens/Standby.py
@@ -212,10 +212,11 @@ class QuitMainloopScreen(Screen):
 
 inTryQuitMainloop = False
 
-def getReasons(session):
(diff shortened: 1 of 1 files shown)


pr2git:
Is this related with my bug report?
https://devtools.openpli.org/issues/455
If so please foreseen an option to totally disable this check that is really boring me when I update my boxes. I really don't need this kind of check.

littlesat:
Thanks for the idea... I just comitted a different code by not showing the messagebox at all when the harddisk is running and the reason is related to the harddisk.... (previously it wasn't also never displayed on Harddisk running)

PR #2031, opened 2019-03-02
Add build status to README

persianpros:
This is how it looks: https://github.com/OpenPLi/enigma2/blob/065d979e765fbc221f634080f541314eef2009db/README.md

### DIFF
--- a/README
+++ b/README
@@ -1,1 +1,0 @@
-To build enigma2, start reading here: https://wiki.openpli.org/Information_for_Developers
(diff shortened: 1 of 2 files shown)


WanWizard:
The problem with this is that it hardcodes the develop branch, so every time you branch, you need to change it manually, otherwise it shows the wrong status.

Also, in general PR's that break travis should not be merged, so the chances of that saying something else than passing are quite slim. 

On the other hand, it happens quite often that travis fails (like it does for this PR) for reasons that are related to the travis CI platform, and not the code (this is a readme change!), making it very unreliable.

PR #2033, opened 2019-03-03
[Standby]  Finetune some code (step 2)

Dima73:
-show message only not standby
-set message  default_yes as True
-set timeout message if reasons 30 sec.
!!!Otherwise, we will never turn off the receiver remotely.

### DIFF
--- a/lib/python/Screens/FlashImage.py
+++ b/lib/python/Screens/FlashImage.py
@@ -1,6 +1,5 @@
 from Screens.Screen import Screen
 from Screens.MessageBox import MessageBox
-from Screens.Standby import getReasons
 from Components.Sources.StaticText import StaticText
 from Components.ChoiceList import ChoiceList, ChoiceEntryComponent
 from Components.config import config, configfile
@@ -190,7 +189,6 @@ class FlashImage(Screen):
 		self.downloader = None
 		self.source = source
 		self.imagename = imagename
-		self.reasons = getReasons(session)
 
 		self["header"] = Label(_("Backup settings"))
 		self["info"] = Label(_("Save settings and EPG data"))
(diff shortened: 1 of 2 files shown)


Taapat:
You missed that getReasons() it is used also in FlashImage, so you can't remove getReasons().

littlesat:
Sorry I totally do not understand the issue you try to solve.......

Dima73:
See 
https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Screens/Standby.py#L241
This not check standby mode.
When use remote power control,.e.g. OpenWebif, need timeout MessageBox for run or  execute instantly if in standby mode.

littlesat:
Flashimage needs get reason!!!!!  
The issue you have when the box is in standby and you want to shut down from owif that the message apprears? Sorry I do totally not get the issue you try to solve... please describe that first!
The message should avoid the box does never shut down when a recording is on, someone is streaming...
When we have the issue description we can solve it!
To avoid someone except it by accident I did close this...

littlesat:
The OWIF should be able to set the timeout and default to a specific value so it should go right....
def __init__(self, session, retvalue=1, timeout=-1, default_yes = False):

As you can see the timeout and default is there...!!!

littlesat:
The OWIF should be able to set the timeout and default to a specific value so it should go right....
def __init__(self, session, retvalue=1, timeout=-1, default_yes = False):

As you can see the timeout and default is there...!!!

Dima73:
In any case, we need check of the standby mode this
https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Screens/Standby.py#L241

littlesat:
Why... as even in standby you should not shut down when there is a reason.... where this is called should be made  Correctly... as far I understand the issue is when you want to shutdown or restart from openwebif... then (the same as with flashimage) openwebif should ask before it tries to shutdown.... it also seems your issue was always there as there was never a standby check before... and even in standby you cannot suddenly shutdown/restart etc...

Dima73:
I want to turn off the remote receiver  and I do not care for some conditions, especially since I do not see them.

littlesat:
The conditions are mandatory for client/remote boxes...  It should be solved in OpenWebif. I just pushed a commit so that this is possible via OpenWebIf (but the changes there must still be made including some checks when required).

PR #2066, opened 2019-03-19
[Cosmetic] Add format to the "Show menu or plugin numbers" when enabled.

linuxstv:
Samples:
* 1.
* 1 )
* ( 1 )
Default is the default format as usual.

Preview:

1_0_1_2B_190_2C0_67D0000_0_0_0_20190319200851


WanWizard:
@littlesat ?

Taapat:
Here it is added the ability to change the look of your numbers as users feel better.

Personally, I am against the configuration of all that is possible.
I think it is better to find or discuss about the best solution and use only it.

PR #2077, opened 2019-03-31
Pixmap caching

Huevos:
C++ layer pixmap caching:
loadPNG and loadJPG have been updated to allow caching using an additional defaulted parameter.
loadPNG defaults to cached
loadJPG defaults to not cached
Caching is implemented in a helper class in pixmapcache.cpp. The cache holds a reference to the pixmap in the same manner as the previous caching. Images are reloaded from disk when the file modified time or the file size changes.
If the cache reaches a size of 256 images, the least recently referenced pixmap is removed. This is, I think, the same behaviour.
LoadPixmap has been updated to pass its cached parameter through or to use the default rules if it is not supplied by the caller.

I found that the cache caused protection fault crashes on exiting the enigma2 process when the cache held ePtr<gPixmap>, so I switched to using a gPixmap* and manually managing the refcounts. Since then, no crashes.

2nd commit in the pull request fixes a crash when using LoadPixmap to load JPGs
Verbose logging of pixmap caching has also been removed

This code has been tested for several weeks without any problems being found.

### DIFF
--- a/lib/gdi/Makefile.inc
+++ b/lib/gdi/Makefile.inc
@@ -20,6 +20,7 @@ gdi_libenigma_gdi_a_SOURCES = \
 	gdi/lcd.cpp \
 	gdi/picexif.cpp \
 	gdi/picload.cpp \
+	gdi/pixmapcache.cpp \
 	gdi/region.cpp
 
 gdiincludedir = $(pkgincludedir)/lib/gdi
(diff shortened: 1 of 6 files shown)


athoik (approved):
I would prefer to use size_t instead of uint on variable MaximumSize, since size() returns size_t.

Maybe a cosmetic update for the future.

Patch looks good.

Huevos:
Thanks.

If it is just cosmetic can the PR be accepted and tweaked afterwards please?

littlesat:
That should be adapted later... please do it!

PR #2097, opened 2019-04-07
Fix Latvian layout in VirtualKeyBoard

Taapat:
Layout which is introduced here https://github.com/OpenPLi/enigma2/commit/0f9e71d6560fbe25ec62220f05bf4f20d4f6562c
does not match the Latvian layout in http://kbdlayout.info/ and is absolutely useless.
Restore the Latvian layout as it was before.

### DIFF
--- a/lib/python/Screens/VirtualKeyBoard.py
+++ b/lib/python/Screens/VirtualKeyBoard.py
@@ -322,29 +322,23 @@ class VirtualKeyBoard(Screen, HelpableScreen):
 		]
 		self.latvian = [
 			[
-				[u"", u"1", u"2", u"3", u"4", u"5", u"6", u"7", u"8", u"9", u"0", u"-", u"f", u"BACKSPACEICON"],
-				[u"FIRSTICON", u"\u016B", u"g", u"j", u"r", u"m", u"v", u"n", u"z", u"\u0113", u"\u010D", u"\u017E", u"h", u"\u0137"],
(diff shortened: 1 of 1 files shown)


IanSav:
This change is code bloat.  "Latvian - Standard" has already been provided in the locale table.

Taapat:
Check before write!
Pay attention to alternative layouts.
Do you understand anything in Latvian?
And "Latvian - Latvia" is the keyboard that is selected according to the language by default.

AbuBaniaz:
00000426 Latvian
00010426 Latvian (QWERTY)
00020426 Latvian (Standard)

Current code only shows two. How can the third be added?

IanSav:
It is easy to add as many locales and derivatives as required.

While not specifically requested I think Taapat is asking for Latvian (Standard) to become the default.  (It is already available.)

Taapat:
No Latvia-QWERTY is actually used in Latvia.
But with double Shift without duplicating "|".
And that is what I want to implement.
Extra special letters in real keybord are switched with one key, for both- large and small letters.
For you, they are just incomprehensible signs that you have placed them in different layouts that make this layout unusable.
I not see reason to leave another Latvian layout, but if you feel you need it, you can keep it.
I suggest changing the layout that is offered by default in enigma.

IanSav:
Latvia (QWERTY) is now defined in the pending code update and has been made the default locale for Latvian.

littlesat:
Hopefully all included in #2101

IanSav:
If there are still any issues please report the details in the appropriate forum thread.

PR #2100, opened 2019-04-07
Fix translation in VirtualKeyBoard

Taapat:
You cannot translate a variable because in this case you can create a translated phrase when generating translation files and these words are removed from the translation files.


IanSav:
This change breaks translations of user defined buttons.

Taapat:
And in what buttons need translation?

Taapat:
And how this translation is parsed on po file generation?

IanSav:
Text string translations have been changed in the next version of VirtualKeyBoard.  It will now be a requirement that ALL text strings for use as text buttons must now be individually translated.  This should address the concern / issue of translation string harvesting.

The documentation will be updated with this new requirement.

littlesat:
Hopefully all included in #2101

IanSav:
If there are still any issues please report the details in the appropriate forum thread.

PR #2101, opened 2019-04-08
Update VirtualKeyBoard

IanSav:
* [VirtualKeyBoard.py] Update languages and translations

    The language changes are:
    -  Add a Hungarian layout to make adoption by OpenATV easier.
    -  Add and make "Latvian - QWERTY" the default keyboard for the "lv_LV" locale.

    The translation change removes the automatic translation of text buffer buttons. All text strings defined in the keymaps will be used EXACTLY as defined. If translation of the text is required then translation must be specifically requested. The documentation has been updated with this change.

    The translation change was made to address a concern that translation strings could not be automatically detected during the translation string harvest process.

* [VIRTUALKEYBOARD] Update documentation

    Update the VirtualKeyBoard documentation to match the latest changes to the code.


IanSav:
If there are still any issues with this updated version please report the details in the appropriate forum thread.

PR #2130, opened 2019-05-09
Fix white spaces in LICENSE file

technic:
It is very minor, sorry for bothering you. 
Please check whether this fixes github license badge in the top bar.

There are only white space changes. LICENSING is not changed at all.
We do it for fun. It allows github.com scripts to parse the LICENSE file correctly and put cool label in the repository header.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,12 +1,12 @@
-		    GNU GENERAL PUBLIC LICENSE
-		       Version 2, June 1991
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
 
  Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
-			    Preamble
+                            Preamble
 
   The licenses for most software are designed to take away your
 freedom to share and change it.  By contrast, the GNU General Public
@@ -56,7 +56,7 @@ patent must be licensed for everyone's free use or not licensed at all.
   The precise terms and conditions for copying, distribution and
 modification follow.
 
-		    GNU GENERAL PUBLIC LICENSE
+                    GNU GENERAL PUBLIC LICENSE
    TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION
 
   0. This License applies to any program or other work which contains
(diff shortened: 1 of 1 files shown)


technic:
Did it work though...

WanWizard:
I don't know, as I have no clue what this is about. ;)

I know github has a "view license" button, and that works fine.

technic:
Yes, but it can be more nice like here: https://github.com/technic/make-callv/blob/master/LICENSE

WanWizard:
I understand what you mean.  And no clue why it doesn't work, I have compared the file with the one github copies in when you create a new repo, and they are 100% identical.

WanWizard:
Maybe because there is a second LICENSE file in the root?

technic:
> 
> 
> Maybe because there is a second LICENSE file in the root?

Who knows... well forget about it.

technic:
Got it. It's COPYING file.
```
License:        NOASSERTION
Matched files:  LICENSE, COPYING, LICENSE.GPLv2
LICENSE:
  Content hash:  3becd209e8ed8039656c1debe01dd17b9a79208f
  Confidence:    100.00%
  Matcher:       Licensee::Matchers::Exact
  License:       GPL-2.0
COPYING:
  Content hash:  ff38b05184ad0a71510a62f7f04892c2fea2eb38
  License:       NOASSERTION
LICENSE.GPLv2:
  Content hash:  e56ba662d6a97efc252ccf1e6b01edf20f898a88
  Confidence:    99.66%
  Matcher:       Licensee::Matchers::Dice
  License:       GPL-2.0
  Closest non-matching licenses:
    GPL-2.0 similarity:  99.66%
    MPL-2.0 similarity:  44.60%
```

WanWizard:
Weird, as in clearly detects the other two, one with a 100% match?

technic:
Yes, but since the third one is inconsistent it doesn't show anything. I checked it in the new repository.

WanWizard:
Sounds like a github bug to me.

WanWizard:
Does it work when you remove the LICENSE.GPLv2 file? It is a duplicate anyway, I have no problem with removing that one.

technic:
No it doesn't help.
I think it's not a bug because if there is another License/copyright conditions in the COPYING file then it is better not to assume anything. Removing `COPYING` will fix the issue.

WanWizard:
The COPYING file should not contain any license information. If it does, it should be either COPYING, or LICENSE, but not both. And only a "see LICENSE" is useless as well.

So I'd say both the COPYING and the LICENSE.GPLv2 files can go.

PR #2133, opened 2019-05-10
Fix Sources.List pageUp/pageDown methods

### LINKED ISSUE #2129: Wrong pageUp/pageDown implementation in Sources.List
In this commit:
https://github.com/OpenPLi/enigma2/commit/7d369fb1fb1d5f6d2d554aa4f38c249d397f6f42

https://github.com/OpenPLi/enigma2/blob/52d9717307ab5f6cd11ac49668b857121fc572b4/lib/python/Components/Sources/List.py#L109-L125

Why this magic number `10` in pageUp/pageDown instead of calling `self.master` functions like here:

https://github.com/OpenPLi/enigma2/blob/22601d43f394bc9e043e29970ab705f4a4b03a6c/lib/python/Components/MenuList.py#L47-L53

Also see [dmm implementation](http://git.opendreambox.org/?p=enigma2.git;a=blob;f=usr/lib/enigma2/python/Components/Sources/List.py;h=4b4d80a80683fd56a489a6fe196d06a3398902f7;hb=HEAD#l100)

technic:
I am not sure that this is better. Propose to fix https://github.com/OpenPLi/enigma2/issues/2129

### DIFF
--- a/lib/python/Components/Sources/List.py
+++ b/lib/python/Components/Sources/List.py
@@ -107,22 +107,18 @@ to generate HTML."""
 		self.disable_callbacks = False
 
 	def pageUp(self):
-		if self.getIndex() == 0:
-			self.index = self.count() - 1
-		elif self.getIndex() - 10 < 0:
-			self.index = 0
-		else:
-			self.index -= 10
-		self.setIndex(self.index)
+		try:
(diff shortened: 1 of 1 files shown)


WanWizard:
@littlesat ?

PR #2141, opened 2019-05-17
Add option "Complete disable PiP"

Dima73:
Simple users, especially with children, the picture-in-picture function
is simply not needed.
When "Complete disable PiP" option is activated, the start buttons of
the PIP will be hidden after restarting the GUI.

### DIFF
--- a/data/setup.xml
+++ b/data/setup.xml
@@ -84,6 +84,7 @@
 		<item level="2" text="Show picons in display" description="Configure if service picons will be shown in display (when supported by the skin)." requires="HasFrontDisplayPicon">config.usage.show_picon_in_display</item>
 		<item level="2" text="Infobar frontend data source" description="Configure the source of the frontend data as shown on the infobars. 'Settings' is as stored on the settings. 'Tuner' is as reported by the tuner.">config.usage.infobar_frontend_source</item>
 		<item level="2" text="Show SNR percentage instead of dB value" description="By default, SNR will be shown in dB (when supported by the tuner). This setting forces SNR to be shown as a percentage instead.">config.usage.swap_snr_on_osd</item>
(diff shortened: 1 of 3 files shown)


littlesat:
os system is not done... it forks (doubles) the memory usages of e2! You never should use it! See on other places in e2 how we usually do this (Console().ePopen(..) or so) But parsing the settings file is not required when you properly ‘design’ this feature. Also the reboot ui is not required...
When you do not show the pip options instead of completely killing the feature the config is available... and when you add a parental control not timed out to it you also instantly enable pip features when the parental code is entered. (Add a check function to components.ParentalConytrol that can be called to check if PiP function should be shown) Then also no reboot ui is needed.
You also see it when box does not have PIP feature. And also better to move config to parental setup.

Dima73:
I suggested the easiest option. And it works at 100%.
You offer a rather complicated option, how do we disable the PIP in plugins?
https://github.com/OpenPLi/enigma2-plugins/search?q=SystemInfo%5B%22NumVideoDecoders%22%5D&unscoped_q=SystemInfo%5B%22NumVideoDecoders%22%5D

littlesat:
Your suggestion is sorry to say not good... os system is not done... and reboot ui to enable/disable it for children is also not done and not required. Also (re)parsing config is NOT required... you simply need to add a check where pip options are shown... the it works instantly without reboot and is not extra Complicated. It does also not work practically as the intend was to hide it for children and it does not auto disable (for x time) when adult entered pin...

littlesat:
Why not change at least the systeminfo key via an addNotification in usageconfig.py....?

eriksl:
If this is meant as a parental guidance control, it should be configured there (otherwise the children will happily re-enable it...)

It it's meant as a workaround for buggy drivers, it should be a plugin.

I don't a valid use case where it should be a config option.

Dima73:
Ok,  as plugin
https://forums.openpli.org/topic/33634-enigma2-patches-proposal/?view=findpost&p=1059629

PR #2142, opened 2019-05-18
Testscreens - added 'Pixels' screen with color test screens

ims21:


ims21:
For me it is more comfortable for user. User must not thinking about buttons.
All colors are cyclically switched with left (r,g,b,w,bk,c,m,y) or right(y,m,c,bk,w,b,g,r) and in addition, r,g,b is used for red, green, blue color too (except yellow - it is not basic color).

littlesat:
I think it is better you toggle RGB with the color buttons... and of course use left right...

ims21:
And what is there used ?!

littlesat:
Can be improved later so I merge...

PR #2143, opened 2019-05-19
MovieSelection - partialy fixed local setting per directory

ims21:
- if not exist local setting, use global value for directory
- put all dependent items top of config list
- modified helptext for this item
TODO: solve status, when is local setting enabling only


ims21:
no yet

PR #2146, opened 2019-05-21
all po files updated from develop at 2019-05-21 20:50

pr2git:


persianpros:
If you close this PR and open a new one after https://github.com/OpenPLi/enigma2/pull/2144 that would have new strings as well.

pr2git:
There is no conflict between each others, once your PR and mine will be accepted I will generate again the po files. Generating po is done on a (quite) regular basis.

persianpros:
Yeah I know, just wanted not to have two same string updates right after each other ;)

pr2git:
The all po also include the very latest french translation so would be nice to have it approuved anyway. ;-)

PR #2167, opened 2019-06-15
[Standby.py] Improve readability of exit codes

IanSav:
This change replaces the magic exit code numbers with more descriptive variable names.  These variables can be imported by code across Enigma2 to ensure consistent usage of exit codes across the code base.  This also protects against any future code changes where the exit codes could be renumbered.


littlesat:
Now only the rest...;)

IanSav:
With the base definitions in place all future edits can incorporate the variable names.

I am also considering adding the exit codes from all other images to all other images so that there is less chance of future conflicts.

PR #2169, opened 2019-06-19
[Menu] Rename Service searching

AbuBaniaz:
The current name does not mention anything about tuners. Tuners and scanning covers both terms


WanWizard:
I fully agree with renaming it. 

But does that require a HTML entity for the ampersand? Maybe it is/looks better if you just use "and" instead?

AbuBaniaz:
Needs the HTML entity, otherwise won't load. I used `&` instead of `and` to keep length shorter. I can change it to `and` if you prefer.  However, ViX has new name, pull request submitted to ATV.

WanWizard:
Ok, clear.

AbuBaniaz:
Thanks for merging.

PR #2172, opened 2019-06-25
Create an EPG menu

AbuBaniaz:
For EPG menu and  EPG plugins that want to be there


AbuBaniaz:
Moves the EPG settings out of thr GUI settings menu into a dedicated menu

pli_epg

pli_within_epg

PR #2184, opened 2019-07-07
NimManager.py: add new method "getNimListForSat".

Huevos:
Returns a list of tuners available for the orbital position given as the argument.
e.g. find all the tuners that can tune Astra 1...
nimmanager.getNimListForSat(192)

### DIFF
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -1070,6 +1070,9 @@ class NimManager:
 								list.append(user_sat)
 		return list
 
+	def getNimListForSat(self, orb_pos):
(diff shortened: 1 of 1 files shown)


littlesat:
What Will use it?

Huevos:
Well nothing at the moment because it wasn't present. So just another tool in the NimManager toolbox. Once present it can be used in enigma or plugins.

Huevos:
E.g. a plugin such as CrossEPG could use this to simplify selecting a tuner list.

littlesat:
Or ABM.

Huevos:
Yes, that too. I think it is easier and cleaner to do this function from inside NimManager where all the variables are available, rather then a messy function inside a plugin.

Thanks for merging.

PR #2212, opened 2019-07-22
[ServiceInfo] fix when not play service

Dima73:
File "/usr/lib/enigma2/python/Screens/Screen.py", line 93, in execBegin

File "/usr/lib/enigma2/python/Screens/ServiceInfo.py", line 190, in
ShowServiceInformation
  File
"/usr/lib/enigma2/python/Screens/ServiceInfo.py", line 224, in
getTrackList
AttributeError: 'NoneType' object has no attribute
'getCurrentTrack'

### DIFF
--- a/lib/python/Screens/ServiceInfo.py
+++ b/lib/python/Screens/ServiceInfo.py
@@ -221,8 +221,8 @@ class ServiceInfo(Screen):
 
 	def getTrackList(self):
 		trackList = []
-		currentTrack = self.audio and self.audio.getCurrentTrack()
 		if self.numberofTracks:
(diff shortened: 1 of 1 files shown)


WanWizard:
Isn't it better to move it inside the "if", as that is where it is needed, and assuming NumberOfTracks is 0 if there are no tracks (so also no CurrentTrack)?

Dima73:
I do not see the difference.

WanWizard:
I didn't say the check should go... ;-)

littlesat:
Dima thanks... we just touched this detail because you touched that part of code...

PR #2227, opened 2019-07-30
Update MessageBox input icons

IanSav:
This change updates the look of the MessageBox input_* icons.  All the icons now have a similar look and feel.  Two new icons have been added for the Warning and Message types.  These five icons have been provided in both the 720 and 1080 resolutions.

Both sets of icons have been produced in the standard 53x53 and 80x80 pixel resolutions as per the previous icons so that no changes should be required by any skins that use these icons.


IanSav:
By the way, there will be an updated MessageBox that will use these updated and additional icons.

PR #2229, opened 2019-07-31
[Skin_default] add more buttons

AbuBaniaz:
png files taken from 720 skin, (PLI HD)


littlesat:
Is this still required due to the other previous change

AbuBaniaz:
Yes, otherwise no PNG files will be found.

data/skin_default/buttons/Makefile.am
```diff
@@ -23,19 +24,35 @@ dist_install_DATA = \
...
+	key_text.png \
+	key_tv.png \
+	key_yellow.png \
+	left.png \
+	play.png \
+	rec.png \
 	red-big.png \
 	red.png \
+	repeat-off.png \
+	repeat_on.png \
```
LraiZer: repeat-on.png has hyphen, not underscore ;p

AbuBaniaz: Thanks. Corrected.

However, would be better if wildcards were used like on other images

PR #2236, opened 2019-08-06
[ActionMap.py] Improve initialisation and logging

IanSav:
- Use a slightly longer form of initialisation for the ActionMap __init__. This format avoids an issue where the initialisation parameters become persistent across all invocations of this code. This is inappropriate and undesirable.

    In essence the problem with the original code is that it appears to be an efficient way to create a default empty list and dictionary. What actually happens is that these items get created as a persistent objects. Every invocation of the \_\_init\_\_ method that doesn't specify the parameters will use that SAME (persistent) objects. Any changes to either object will persist (be remembered) and be carried forward into every other invocation of this method! THIS IS NOT WHAT SHOULD HAPPEN!

    See  https://docs.python-guide.org/writing/gotchas/

- Add new logging code to report actions that are not accessible due to no matching keymap entries.  This may not be an error condition!

- Improve variable name clarity.

- Code is now common between main images.

### DIFF
--- a/lib/python/Components/ActionMap.py
+++ b/lib/python/Components/ActionMap.py
@@ -4,7 +4,9 @@ from Tools.KeyBindings import queryKeyBinding
 
 
 class ActionMap:
-	def __init__(self, contexts=[], actions={}, prio=0):
+	def __init__(self, contexts=None, actions=None, prio=0):
+		contexts = contexts or []
+		actions = actions or {}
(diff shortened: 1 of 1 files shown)


IanSav:
@Littlesat, thanks for the further code optimisations.  I have also made the same changes to the other images.

To make merges easier I will try to keep these files in sync.

PR #2241, opened 2019-08-10
AudioSelection.py: Don't hardcode PLi-HD skin

persianpros:
Still I'm not sure about https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Plugins/SystemPlugins/OSDPositionSetup/overscanwizard.py#L131


littlesat:
The one in the overscan wizard is for selecting the default skin

PR #2245, opened 2019-08-23
[enigma2.sh.in] Add option for skin backdrop

IanSav:
This change adds the option for skins to include a backdrop.mvi image at the top level of the skin that will be used when that skin is selected.

The code first checks to see if a skin is defined in the "settings" file and uses it if found. If no skin is identified then the "skin.pyo" file is searched to see if a built-in skin has been defined. If found then that will be used. If no skin associated backdrop is identified then the code will progress to the original search in "/etc/enigma2/" then "/usr/share/" and finally look to use the "bootlogo.mvi" image.

### DIFF
--- a/tools/enigma2.sh.in
+++ b/tools/enigma2.sh.in
@@ -22,7 +22,13 @@ LIBS=@libdir@/libopen.so.0.0.0
 while : ; do
 	# show bootlogo on enigma2 start
 	if [ -x @bindir@/showiframe ]; then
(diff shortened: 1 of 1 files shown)


AbuBaniaz:
Huevos, Shimmelreiter and myself have failed to explain that the user location must have highest priority. I have reverted this on Vix.

IanSav:
This proposed change is not acceptable to all images.  I will revert to the previous mode of operation.

PR #2248, opened 2019-08-26
[NumericalTextInput.py] Correct typo In "7" button

IanSav:
Thanks to thawtes from OpenATV for spotting and reporting the error.


littlesat:
I this something we also need in release?

IanSav:
It is not a fatal issue, look how long it took for the problem to be found.  It would be good for the update to go out to anyone using the version with the error but I wouldn't rush like a critical error.

PR #2251, opened 2019-09-01
config: add lower case to "show on/off button in multiple selections"

Huevos:
This adds lower case versions of the keywords used for displaying the graphical sliders in ConfigSelection. These are found in various plugins, etc.

Also adds "Enable" to complement "Disable".

### DIFF
--- a/lib/python/Components/config.py
+++ b/lib/python/Components/config.py
@@ -367,7 +367,7 @@ class ConfigSelection(ConfigElement):
 		from config import config
 		from skin import switchPixmap
 		if self.graphic and config.usage.boolean_graphic.value and switchPixmap.get("menu_on", False) and switchPixmap.get("menu_off", False):
-			boolvalue = self._descr in (_('True'),_('Yes'),_('Enabled'),_('On')) or (False if self._descr in (_('False'),_('No'),_("Disable"),_('Disabled'),_('Off'), _("None")) else None)
+			boolvalue = self._descr in (_('True'),_('true'),_('Yes'),_('yes'),_('Enable'),_('enable'),_('Enabled'),_('enabled'),_('On'),_('on')) or (False if self._descr in (_('False'),_('false'),_('No'),_('no'),_("Disable"),_('disable'),_('Disabled'),_('disabled'),_('Off'),_('off'),_('None'),_('none')) else None)
 			if boolvalue is not None:
 				return ('pixmap', switchPixmap["menu_on" if boolvalue else "menu_off"])
 		return ("text", self._descr)


persianpros:
I think https://github.com/OpenPLi/enigma2/pull/2252 is better.

littlesat:
Wierd... where are the lower case version used?.... maybe change them there? (So also those are translatable?)

Huevos:
Then you are forcing any plugin writer to conform with your style. All the plugins I write use lower case.

Huevos:
XML update, MisPlsLcnScan, Blind scan, Terrestrial Scan, ABM, Custom Mix Importer, Channels Importer, CrossEPG, HR Tuner Proxy, to name just a few.

Huevos:
I have checked in the OpenPLi enigma2 nl.po file. All of the following exist...

on, On, ON, off, Off, OFF, yes, Yes, YES, no, No, NO, enable, Enable, enabled, Enabled, disable, Disable, disabled, Disabled, true, false, none, None

Some do not even use the same translations, so switching to lower case is never going to work properly.

Huevos:
In fact OpenPLi uses a mixture of upper and lower case through its own enigma2 core screens and plugins. Sometimes even mixing case in the same screen.

So in my opinion this commit is correct.

Or someone can go through the whole OpenPLi enigma2 code base and sort the mess out properly so everything follows one style. And update all the translation files to match.

Huevos:
And translation is so poor that it uses upper case when the English is in lower case.

Huevos:
And, would you believe, sometimes translation of a word using upper case is completely different from the same word in lower case.

littlesat:
Looks like this is terrible (typically enigma2), but it needs to be adapted that at any location it is the same. So I simply prefer the plugins adapted sorry! It makes also the translations easier. To mix it up here is just add another crappy work-a-round spaghetti fix.
Add Enable/Disable is fine..

Huevos:
It is not spaghetti. It is just a tuple with a few extra items in it.

It is not just plugins. It is a lot of OpenPLi screens too. e.g ScanSetup.py, etc.

It is your choice, and your users that will miss out on the feature that you have taken time to code.

Anyway nice feature.

littlesat:
It should be done good... without tricks... this means some stuff can be cleaned aswell...

Huevos:
I will leave that up to you. ;)

littlesat:
Why?!?!

WanWizard:
Because nobody seems to understand what you mean or want, but you? ;-)

Huevos:
I am happy pushing code back to PLi. It is good for the enigma community in general, but I don't have time to go through all your code and translations looking for inconsistencies. I have pointed to a few above but this must represent the entire code base.

littlesat:
I understand... but better to fix it isn’t it... we had an agreement to do it first letter capital the rest small... I think we should avoid to add more work-a-rounds for the mass... maybe think for a different solution... somehow we should find a way to check case unsensitive

littlesat:
menu_on_values = set([x.lower() for x in (_(“Enable”),....”))]

Huevos:
That doesn't fix it. I already tried to explain. 

This is a sample from OpenPLi nl.po.

-------------

msgid "Enable"
msgstr "Aan"

--------------

msgid "enable"
msgstr "Geactiveerd"

-------------

Also, ViX has a lower case policy so all our plugins are going to remain lower case.

WanWizard:
To me in comes across as very logical to standardize in the code, and deal with any case issues in the translations (for example English seems to prefer lower case, USA and Germany capitalize a lot of first letters of words).

Huevos:
The way I suggested works in all situations. All other suggestions so far break for some reason or other. It is just a few extra words in a tuple. Not spaghetti, and not a hack.

I am trying to suggest a way that works for everyone without having to do a massive cleanup of python code and translation files.

WanWizard:
No problem with that, as long as what I said is still applied, and we convert to lowercase when we encounter one that isn't lowercase. So eventually we get to something standardized.

littlesat:
Alternative commit merged... 
https://github.com/OpenPLi/enigma2/commit/9e2451ce70342d119ffb1420f51b1ed477734ac6

PR #2252, opened 2019-09-02
config.py: Ignore case sensitive when checking for answers, this adds…

persianpros:
… lower case versions of the keywords used for displaying the graphical sliders in ConfigSelection. These are found in various plugins, etc.

Also adds enable to complement disable, Thanks to Huevos


Huevos:
So how does this work if "Enable" exists in the .po file and "enable" does not? 

And it does not even take locale into account.

Huevos:
In fact this version does not work at all in other locales.

Huevos:
Tested this in Dutch locale and appears to work.

Huevos:
DO NOT MERGE.

This commit does not work. Tested in Dutch locale.

PR #2256, opened 2019-09-07
[ServiceInfo] Do not embed color in python code

nautilus7:
In service info, when pressing the green button, the ecm info is showing in yellow/orange color. This is because the color in embedded in the python code. So, instead of setting a color in the code, we should allow skins to determine what color to use.

After this change, skins will continue to work perfectly fine and the ecm info will be displayed in whatever color is already selected for the service info screen.

### DIFF
--- a/lib/python/Screens/ServiceInfo.py
+++ b/lib/python/Screens/ServiceInfo.py
@@ -386,8 +386,7 @@ class ServiceInfo(Screen):
 						extra_info = "extra data=%s" % caid[2]
 				from Tools.GetEcmInfo import GetEcmInfo
 				ecmdata = GetEcmInfo().getEcmData()
-				color = "\c00??;?00" if caid[0] == int(ecmdata[1], 16) and (caid[1] == int(ecmdata[3], 16) or str(int(ecmdata[2], 16)) in provid) else ""
-				tlist.append(ServiceInfoListEntry("%sECMPid %04X (%d) %04X-%s %s" % (color, caid[1], caid[1], caid[0], CaIdDescription, extra_info)))
+				tlist.append(ServiceInfoListEntry("ECMPid %04X (%d) %04X-%s %s" % (caid[1], caid[1], caid[0], CaIdDescription, extra_info)))
 			if not tlist:
 				tlist.append(ServiceInfoListEntry(_("No ECMPids available (FTA Service)")))
 			self["infolist"].l.setList(tlist)


Dima73:
What for?
Do you want to force to rewrite all skins?
Example:
green instead of light green

nautilus7:
Where exactly do you see the need for skin re-writing?

Skins will use whatever color they have already selected for `self["infolist"]`: https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Screens/ServiceInfo.py#L96

Now, skin color selection for the "ecm info" option is overridden by the python embedded color!

nautilus7:
No, you are totally wrong!

The color in the "ecm info" tab will now be the same as in the "pids info" tab. And this color is what is defined for the "infolist" in the skin.

Read the python code before commenting the next time. Thanks.

Dima73:
https://github.com/OpenPLi/enigma2/search?q=infolist&unscoped_q=infolist

ecm info one line color?

nautilus7:
I don't understand what you mean by this...

Whatever color is selected by the skin for the `self["infolist"]` is going to be used for ecm info as well.
Do you have any clue which exactly are the screens I am talking about?
Did you bother checking the code?
Did you bother checking how your current skin works with my change?

Dima73:
What is incomprehensible here?
If you add color to the "infolist", the all list will be of this color.

nautilus7:
Even if a skin does not set a color for infolist explicitly, it gets a default color like every skin element.

After my change all possible  infolist elements like "pids", "ecm info" and "tuner values" will have that same color defined in the skin.

Why do we need to force all skins to have the ecm info displayed in orange color? There is no reason for that. So remove the hard-coded color from the python code!

littlesat:
Can you post some screenshots so I can see what’s meant here?

littlesat:
When you want the skin decides the color or overrules the default do something with attributes.

nautilus7:
Go to **menu --> information --> service info**

There you will see the text in whatever color your skin has selected.
Then press the green button to display the "ecm info". The text will be in yellow color BECAUSE this color is defined in the python code!

See how it displays right now:
1_0_19_7531_426_1_C00000_0_0_0_20190909013714
1_0_19_7531_426_1_C00000_0_0_0_20190909013722


The skin cannot change the yellow color in the ecm info!
My change will allow the color to be the same as in the other screen (white).

Dima73:
This two color
1_0_16_70A_12_70_1680000_0_0_0_20190909071536

nautilus7:
OK, I'll add an attribute to the code, so it can change colors

littlesat:
Indeed the skin cannot change the color code that highlights the current 'chosen' ECM. So this change request makes no sense. I agree it should be skinnable. But then we need something to arrange with parameters/attributes that forward the color to python.

nautilus7:
Believe it or not I had never notice that multiple ecm pids could be listed in that screen. I always thought that only the active pid is displayed there. Now I understand the reason behind the hard-coded color.

nautilus7:
@littlesat I can make it read a color from the skin instead of using a hard-coded color from the python.

But, I think it would be better to indicate the active ecm pid using a word (e.g. "active") instead of using a different color. It looks much simpler. I don't see why a different color is needed.

What do you think???

1_0_1_C1D_1E78_71_820000_0_0_0_20190909220715

PR #2261, opened 2019-09-10
[ServiceInfo] Use text instead of color for indicating active ecm pids

nautilus7:
Embedding colors in python means that skins have no control on them.
This change adds a text string next to the active ecm pid, instead of using a different display color.

Skins need no changes as all.

This is how it will look like after this change:

1_0_1_C1D_1E78_71_820000_0_0_0_20190910192138

### DIFF
--- a/lib/python/Screens/ServiceInfo.py
+++ b/lib/python/Screens/ServiceInfo.py
@@ -386,8 +386,8 @@ class ServiceInfo(Screen):
 						extra_info = "extra data=%s" % caid[2]
 				from Tools.GetEcmInfo import GetEcmInfo
 				ecmdata = GetEcmInfo().getEcmData()
-				color = "\c00??;?00" if caid[0] == int(ecmdata[1], 16) and (caid[1] == int(ecmdata[3], 16) or str(int(ecmdata[2], 16)) in provid) else ""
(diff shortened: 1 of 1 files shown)


littlesat:
Not a good idea... better to make the color configurable via skin attribute

littlesat:
But let further discuss... I want to avoid a config... adding text is somehow also good...

nautilus7:
I also thought that having an extra config for color is too much. It adds unnecessary complexity to the skin and python code as well.

If you think the text is too long, we can use just "active" instead of "currently active".

Also bare in mind that there are several other cases in openpli where the color is embedded in the code. If we choose to add new attributes to the skin for each one will complicate things even more.

littlesat:
The space behind is a bit wierd... better create the format string one with and one without (active) text.
Now I need time or you... when you understand what I mean

nautilus7:
Do you mean not having an extra space at end, when the pid in not active?

And what about the string? Should I set it to "(active)" or keep it as "(currently active)" ?

littlesat:
FYI https://github.com/OpenPLi/enigma2/commit/caeefee7ca9989af4b1819a4a2d99a89aed51117

nautilus7:
@littlesat the string "used" cannot be translated! This is wrong!

PR #2272, opened 2019-09-14
Add buttons with comprehensible names

AbuBaniaz:
There are various shapes of button png files. Add them using unique and comprehensible names to avoid confusion


littlesat:
Please not... now we’re adding buttons that really belong in a skin itselves it is getting a never ending story....

PR #2273, opened 2019-09-15
add option "Use margin after/before minutes" yes/no for per zap timer

Dima73:


littlesat:
I’m suggesting to do this differently... just ‘hardcoded’ adjust the start and end time with an switch that is always in the time config and also stored within the timer config... add it to openwebif etc... this is not as simple as you think to do it right.

Dima73:
This is a fully working option.
But if it is not needed, then you can close the comment.
I'm not against.

littlesat:
Ok then I close it... sorry...
It does work but it does not work as it really should be. Please discuss on forum.

PR #2276, opened 2019-09-18
[About.py] Correct a text message in MemoryInfo screen

nautilus7:
The change does not alter the meaning of the text, it just uses more correct English language. So, the existing translations are kept as they are.


persianpros:
Editing po files like this is wrong in my opinion because translators should see the updates of any kind.

They may want to update the translation as well, with these kind of commits they won't.

nautilus7:
If I don't do that, there will be a bunch of people complaining about the lost translations and how much work will be needed by translators to translate the message again.

I don't agree either with editing .po files.

PR #2279, opened 2019-09-20
Fix ChoiceList alignment in mixed lists like the "ChannelContextMenu" screen

nautilus7:
After my previous change https://github.com/OpenPLi/enigma2/commit/5456c1daccb6c47f07e84215da589472970f332d (that allows a different position for choicelists without icons - see PR https://github.com/OpenPLi/enigma2/pull/2278), there is an issue with choicelists where some of the entries have icons and some don't, like in the ChannelContextMenu screen. In such choicelists, the alignment of the list entries is not proper.

This change-set, allows entries of such mixed choicelists to be created with space on the left (when "dummy" key is used), as if all entries had icons, so the choicelist is properly aligned.

See pictures before (pic. 1) versus after (pic. 2) this fix:

1_0_1_28_2_212C_1860000_0_0_0_20190920205420
1_0_1_28_2_212C_1860000_0_0_0_20190920203318


nautilus7:
https://github.com/OpenPLi/enigma2/commit/5456c1daccb6c47f07e84215da589472970f332d

PR #2280, opened 2019-09-23
[ServiceInfo.py] Fixed bugs introduced in previous commit

nautilus7:
This fixes 3 different bugs introduced in commit https://github.com/OpenPLi/enigma2/commit/db4f0533c73e3d55b08982504a048bb2460b0de0#diff-e9865bf9138a460543684c21d2053840
The commit was supposed to do code cleanup and optimization, but it introduced 3 bugs (3 at least - perhaps there are even more):
1. It broke "Editmode" completely.
2. It changed the behaviour of "AudioTracksAvailable" to "true" even if only one audio track was available in a service.
3. It broke the "SubtitlesAvailable" completely.


athoik:
Seems fine! Thanks!

PR #2290, opened 2019-09-29
[ServiceInfo.py] Fix crash in ecm pid info

nautilus7:
The "used" string is meant to replace the last %s, but instead it replaces the first %04X and causes a crash. The bug was introduced in commit https://github.com/OpenPLi/enigma2/commit/7b5907cc9eb5dff63bebe9f8e81e55f760e5cc7a

Reported by @68foxboris 11 days ago...

```
File "/usr/lib/enigma2/python/Components/ActionMap.py", line 57, in action
File "/usr/lib/enigma2/python/Screens/ServiceInfo.py", line 389, in ShowECMInformation
TypeError: %X format: a number is required, not str
```

### DIFF
--- a/lib/python/Screens/ServiceInfo.py
+++ b/lib/python/Screens/ServiceInfo.py
@@ -386,7 +386,7 @@ class ServiceInfo(Screen):
 						extra_info = "extra data=%s" % caid[2]
 				from Tools.GetEcmInfo import GetEcmInfo
 				ecmdata = GetEcmInfo().getEcmData()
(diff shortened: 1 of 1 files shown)


nautilus7:
**Will you also please consider changing the "used" string to "in use"???**

It is more appropriate. It is the ecm pid (currently) in use. It's not something that is not new, but used...

WanWizard:
How about changing it to "active"?

nautilus7:
@WanWizard, no problem, it was my [previous suggestion](https://github.com/OpenPLi/enigma2/pull/2261#issuecomment-530057151) anyway, but littlesat thought it was too long. IMHO though, the "used" he chose was the worst one.

WanWizard:
What is longer or not depends heavily on the selected lanuage. 

"used" is shorter than "active", but in German "verwendet" is longer than "activ". Or in French "utilisee" vs "actif". Or on Dutch "gebruikt" vs "actief"...

imho, change to "active".

PR #2293, opened 2019-09-29
[ServiceInfo.py] Renamed string

nautilus7:
Proposal to rename string as suggested by @WanWizard in https://github.com/OpenPLi/enigma2/pull/2290. Take it, or leave it.

Regards.


WanWizard:
Take it. ;)

PR #2299, opened 2019-10-06
[TimeDateInput.py] Use of proper names for colored buttons

nautilus7:
Colored buttons should be named "key_red" and "key_green" and use StaticText.
Default skin is updated to reflect the changes.
Modern third-party skins using automatic buttons should not have any problems with this change.

In addition, the use of the "TimeDateInput" class is pretty limited (it is used in only 2 or 3 cases) throughout the code of enigma2, so it won't be a disaster for old and unsupported skins.


nautilus7:
@littlesat thanks for merging. This class is only used in 2 EPG screens, when the user wants to enter a date/time to show the EPG for.

Old skins will not be able to show the red and green buttons, but there is the "exit" and "ok" buttons which do the same, so no real problem imho.

PR #2303, opened 2019-10-09
Renamed ALL "upgrade" strings to "update"

nautilus7:
In general, "upgrade" is used for hardware, while "update" is used for software. E.g. when we put a new hard disk into a PC, we do a hardware upgrade, but when we install a newer version of a program, we do a software update.
In openpli, both words and their derivatives are used interchangeably while refering to software updates.

This commit converts all instances of "upgrade" and its derivatives to "update" and its corresponding derivatives and makes sure ALL translations will continue to work like before. For some languages, the translated strings are also updated.


WanWizard:
Too bad opkg still calls it "upgrade". :smirk:

PR #2305, opened 2019-10-13
Greek translation update 13/10/2019

MCelliotG:


po/el.po
```diff
@@ -7751,7 +7751,7 @@ msgid "Show menu paths"
 msgstr "Εμφάνιση διαδρομών μενού"
 
 msgid "Show message when recording starts"
-msgstr "Εμφάνιση μηνύματος στην εκκίνηση της εγραφής"
+msgstr "Εμφάνιση μηνύματος στην έναρξη της εγραφής"
```
nautilus7, line 7754: It should be "εγγραφής" with double "γ".

nautilus7:
Also in line 5131 you forgot to fix another string containing "downmix".

"Multi channel downmix" should become "Αποκωδικοποίηση & μίξη πολυκάναλων ήχων" or Αποκωδικοποίηση & μίξη πολυκάναλων ροών ήχου".

nautilus7:
And another one I have found: In line 11191, "pass" should be "πέρασμα" or something similar. It refers to service scanning, where in some cases we can do one or more scanning passes.

po/el.po
MCelliotG: Yes, it was done the last minute, will be added after the next refresh.

PR #2331, opened 2019-11-03
[SoftwareManager] Removed "\n" from the beginning of some strings

nautilus7:
These strings are the description for the menu options, so the new line is not justified/needed. Also, in one of them there was a full stop missing at the end. All translations are updated, so nothing is lost.


nautilus7:
@littlesat thanks for merging. Well, I know there are too many... but I don't see anyone fixing them, except me!

PR #2335, opened 2019-11-08
[estring] Output as buildShortName

AbuBaniaz:
Remove any Start/End of Selected Area markers before return from convertDVBUTF8()

### DIFF
--- a/lib/base/estring.cpp
+++ b/lib/base/estring.cpp
@@ -640,7 +640,8 @@ std::string convertDVBUTF8(const unsigned char *data, int len, int table, int ts
 		string_to_hex(std::string((char*)data, len < 15 ? len : 15)).c_str(),
 		output.c_str());
 
-	return output;
+// Remove any Start/End of Selected Area markers *now*...
+	return buildShortName(output);
 }
 
 std::string convertUTF8DVB(const std::string &string, int table)


AbuBaniaz:
The data is filtered later on in other files,  so better to do it at source. Below are Vix files which do the filtering, PLI should be similar

```
    Screens/ChannelSelection.py
                            name = name.replace('\xc2\x86', '').replace('\xc2\x87', '')
                            name = name.replace('\xc2\x86', '').replace('\xc2\x87', '')

    Tools/Directories.py:
            # Filter out non-allowed characters.
            non_allowed_characters = "/.\\:*?<>|\""
            basename = basename.replace("\xc2\x86", "").replace("\xc2\x87", "")

    Components/Converter/ServiceName.py
                            name = name.replace('\xc2\x86', '').replace('\xc2\x87', '')

    Components/Converter/ServiceName2.py
                            return name.replace('\xc2\x86', '').replace('\xc2\x87', '')
                                            ret += name.replace('\xc2\x86', '').replace('\xc2\x87', '')
    Screens/TimerEntry.py
                    self.timerentry_name = ConfigText(default = self.timer.name.replace('\xc2\x86', '').replace('\xc2\x87', '').encode("utf-8"), visible_width = 50, fixed_size = False)

    lib/dvb/epgcache.cpp
                                                            // filter short name brakets
                                                            while((pos = name.find("\xc2\x86")) != std::string::npos)
                                                                    name.erase(pos,2);
                                                            while((pos = name.find("\xc2\x87")) != std::string::npos)
                                                                    name.erase(pos,2);
    lib/base/estring.cpp
            static char stropen[] = "\xc2\x86";
            static char strclose[] = "\xc2\x87";
            size_t open = std::string::npos-1;
            while ((open = str.find(stropen, open+2)) != std::string::npos)
            {
                    size_t close = str.find(strclose, open);
                    if (close != std::string::npos)
                            tmp += str.substr(open+2, close-(open+2));
            }
            return tmp.length() ? tmp : str;

    lib/service/servicedvb.cpp
                                                            // filter short name brakets
                                                            size_t pos;
                                                            while((pos = name.find("\xc2\x86")) != std::string::npos)
                                                                    name.erase(pos,2);
                                                            while((pos = name.find("\xc2\x87")) != std::string::npos)
                                                                    name.erase(pos,2);
```

littlesat:
I think this is not good at all... the short names are old names used by a German provider to give short names on the display between brackets. This should not be removed at a high level.... I think vix should also revert it and not always remove the at any string with the convert even when the [] are requored.
Maybe the removal of [] is not required anymore at all... does the German provider still use them (they were in use far back ago for the dbox)

original-birdman:
The characters removed are *not* brackets! They are non-displaying utf8 characters (called Start of Area and End of Area).
Various bits of cache code remove these at a higher level. It surely makes more sense to remove them when received.

They are being removed as *something* is transmitting them and it screws up sorting titles/descriptions as they are non-displaying, but sort after all visible chars.

littlesat:
I thought this was done for displays to shorten channel names.
Sky Germany did this for the dbox at the beginning so channel names could be somehow given properly on the display... as far I know in the current days this isn’t even used yet... so we can totally remove it in python code.... please tel me (with examples, where are they still transmitted?) when I’m wrong.... then better remove it without changing code in cpp!!!!!

AbuBaniaz:
Rob reported a bug which affected how EPG was being sorted.

First fix attempt was trimming the data in py files.
 
Then original birdman came up with fixing data at earliest point. Hence this commit.

original-birdman:
> please tel me (with examples, where are they still transmitted?)

Difficult, as the only person seeing them is Rob on Freesat (in Holland).
From where they show up in the EPG cache it's Now and Next programmes only.

However, this change is at the very end of a function called `convertDVBUTF8()`, and in utf8 these sequences do not display (hence are not in any way "brackets").

WanWizard:
There is no Freesat in Holland, only in the UK, and some Nordic country. There aren't any FTA channels from the Dutch or Flemmish provider.

original-birdman:
To quote Rob:

> This only happens on Freesat services.
> This is with Freesat & OpenTV enabled, but happened already before OpenTV was integrated in E2
 
He is Dutch. I've just assumed he is in Holland.
Mind you, I may have just assumed he was Dutch - just that English isn't his native language and he provides Dutch translations.

RobvanderDoes:
LOL: the usual confusion around nationality/language/location/providers.....
FYI: I'm Dutch, living in the Netherlands & France, speaking (a.o.) Dutch & English, using Dutch, French, German and UK providers.
The only provider named 'Freesat' (I know of) is a UK provider.

Just tested on PLi openpli-7.2-rc-vuultimo-20191109_usb and the same bug is there as I noticed in ViX, and which has been crushed by the files as above. See screenshot

EPG_Alpha_05

in which you see the S-EPG of a Freesat service (Freesat EPG and Now/Next-EIT enabled) being ordered alphabetical; the Now/Next program is listed at the bottom.

RobvanderDoes:
And on a more common service:

EPG_Alpha_06

AbuBaniaz:
Maybe you can post all the steps required to reproduce this bug. I am sorry to say, I tried your description twice and could not reproduce the problem. I am sure PLi will merge the pull request/provide better fix if they can see the bug.

original-birdman:
> then better remove it without changing code in cpp!!!!!

The best place to remove it if it is never going to be used (which it appears not to be, as these chars are stripped out in several other places, and certainly should never get through to a display string) is as soon as it is found.
Whether that is in C++ code or python makes no difference.

PS: The only reason that `buildShortName()` gets called is because it was *already* doing exactly the work that I would have needed to code myself. So I just called what was already there. Don't read anything into the function name.

RobvanderDoes:
> Maybe you can post all the steps required to reproduce this bug.

It's no more complicated then I said before: activate Freesat EPG & Now/Next EPG and open S-EPG on a Freesat service.

BTW: I use ABM to create a Freesat bouquet and choose a service from that bouquet.

WanWizard:
I have to revert this, as it seems to have unexpected side effects.

If I look at the code of buildShortName(), it doesn't seem to strip just the U+0086/U+0087, but it also removes everything in between.

original-birdman:
Do you have an example of such data?

Whomsoever wrote that code originally decided that this was the way to treat U+0086/U+0087 sequences. Given that these are non-printing their presence in a data stream must mean something, so what do *you* think they mean?

WanWizard:
I have no clue what they are meant to do, all I can see is that your commit description says "remove these characters", but the method you used also removes everything in between, which is currently stripping parts of channel names on 19.2e (there is a thread about it in the German section of the OpenPLi forum).

The method you used is 13 years old, so it might originally have another meaning. I'll have a quick check so see if it is used elsewhere, and why...

WanWizard:
epgcache does:
````
	// filter short name brakets
	while((pos = name.find("\xc2\x86")) != std::string::npos)
		name.erase(pos,2);
	while((pos = name.find("\xc2\x87")) != std::string::npos)
		name.erase(pos,2);
````
which might be more appropriate. I'll run a test now.

original-birdman:
So what does the 19.2e broadcaster _mean_ by the U+0086/U+0087 sequences?

original-birdman:
I know `epgcache` does that.
But there seems to be little point in having *multiple* instances of stripping them from data when they might as well just be removed _once_ when getting the original data from a DVB stream.

So instead of calling `buildShortName()` here we should be calling `strip_SOA_EOA()`, which just strips them?

WanWizard:
Your guess is as good as mine. It seems to be in the middle of channel names (for example MotorVision becomes MotorVis). It seems to be inserted at position 9 in the channel name...

WanWizard:
Hmm... Reverted your commit, and rescanned that German transponder, but it remains "Motorvis", so it smells like someone put 1 and 1 together in that forum topic, but didn't quite get to 2....

WanWizard:
Have a look at 19.2e, 12382H. 
All channel names on this transponder are missing a chunk, either in the middle, or at the end:
image

original-birdman:
Hmmm...so it really would appear to be using it to produce short names by taking the long one and marking "bits to be retained" with SOA/EOA markers.

The problem is that some other broadcaster is using the same sequences in EPG (possibly in now&next) to tag programmes that started in the past (so the "now" programme).

original-birdman:
So, for the example of Motorvision => Motorvis, is that any occasion that you can see when this *should* be used?
If not, `buildShortName()` could be changed to just strip U+0086/U+0087.

AbuBaniaz:
As Littlesat mentioned, the function was used to shorten names.

IMO best to leave that function alone and create another function. Provided they don't interfere.

But Wanwizard just said reversing your commit did not solve the issue

WanWizard:
No clue, but I'm an E2 code novice. According to others it the team, it might have been related, long ago, to being able to display the channel name on 8-character VFD's.

AbuBaniaz:
@wanwizard, when you scanned after revert. Did you select "clear before scan".

WanWizard:
> But Wanwizard just said reversing your commit did not solve the issue

Correct. 

I just flashed a box, no restore of config, just configued 1 tuner, and did a direct NIT scan of 19.2e, and the problem remains.

WanWizard:
I have to swallow my words: I made the change but forgot to commit it, so my new build still had this code. Now running a new build with a strip instead of a filter. 

I'll commit if that works, we can have the disussion about whether or not this 13 year old code is still useful, and if so where, later.

AbuBaniaz:
We have reverted this on ViX too

littlesat:
You have to do it later on (in cpp or python) and not in a general convert function. This is asking for problems...

AbuBaniaz:
OK, noted.
We have done this:

https://github.com/OpenViX/enigma2/commit/2420e08c365bd13931941f12f69a61fdab57e499

WanWizard:
@littlesat comments? Can we cherry pick this?

PR #2362, opened 2019-11-30
Remove some strings from translation

nautilus7:
Strings such as kB, MB, GB, ??, etc, are useless for translation, so remove them.

Also, I did some cleanup in the code (removed "object" as "Converter" is already an "object" and fixed indentation).


WanWizard:
Not entirely true, the French for example use giga-octet instead of gigabyte.

nautilus7:
Really? Then reject it. No problem.

I'll create a new PR with the other corrections.

PR #2372, opened 2019-12-08
[NimManager.py] Renamed "Ka-SAT" to "Ka-Band"

nautilus7:
This was in the LNB type selection, where "KA-SAT" did not make any sense.

### DIFF
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -332,7 +332,7 @@ class SecConfigure:
 					sec.setLNBLOFL(10750000)
 					sec.setLNBLOFH(10750000)
 					sec.setLNBThreshold(10750000)
-				elif currLnb.lof.value == "ka_sat":
+				elif currLnb.lof.value == "ka_band":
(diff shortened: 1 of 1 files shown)


athoik:
@Huevos  can you check this one please? 


KAT_SAT committed on 060d14ac55bf6016a87a8179a2eb1934b6a389a9 by you, so you might tell us if that's fine.

nautilus7:
@athoik this is the lnb type selection option under tuner setup. There is "C Band" lnb as well. Ka-Sat does not make sense. It should be named Ka Band instead.

Huevos:
KA-SAT is an LNB type just for provider Saorsat, for use on one transponder on KA-SAT 9.0E. Please don't change the name to Ka-band. There are lots of different lnb's for ka-band. There is no standard and a huge slice of band that would need at least 4 LNBs to cover the whole thing.

Huevos:
If this is to be renamed it should be called "Saorsat LNB".

nautilus7:
OOOOKKKKK, I understand. I am closing it.

PR #2375, opened 2019-12-08
[ServiceInfo.py] Added new converter options for getting video codec

nautilus7:
These boolean tests ("IsVideoMPEG2", "IsVideoAVC", "IsVideoHEVC") can be used to display an appropriate icon (e.g. in the infobar).


athoik:
Thanks!

PR #2379, opened 2019-12-11
[MediaPlayer] do not show playlist

Dima73:
-do not show playlist after play/unpause when the screensaver was
launched before


ghost:
Thanks Bro :)

PR #2381, opened 2019-12-13
[PliExtranInfo.py] Made string available for translation

nautilus7:
This was reverted as part  of https://github.com/OpenPLi/enigma2/commit/ca2a1b142eae4bd131fc1c6724f8420759f50c5c but it didn't cause the problem, as far as I understand.

This time I kept the "N/A" inside the "get" function.

### DIFF
--- a/lib/python/Components/Converter/PliExtraInfo.py
+++ b/lib/python/Components/Converter/PliExtraInfo.py
@@ -173,7 +173,7 @@ class PliExtraInfo(Poll, Converter):
 		return ("SDR", "HDR", "HDR10", "HLG", "")[info.getInfo(iServiceInformation.sGamma)]
 
 	def createVideoCodec(self, info):
-		return codec_data.get(info.getInfo(iServiceInformation.sVideoType), "N/A")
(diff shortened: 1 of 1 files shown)


WanWizard:
description had the same issue, it had to be inside the get(), as that would capture the error and return the default string instead, where "or" only returned the default string if it exists but is empty...

littlesat:
Should we wait for merging this one or are there Some doubts?

Dima73:
This reverts commit 8117865.

It causes a crash when an XML entry has no description

nautilus7:
What reverts commit https://github.com/OpenPLi/enigma2/commit/8117865af6f58b8ee0521a095196c40a46b01c68 ??? That commit has already been reverted.

The present PR, only adds back a part of that commit (not the one that caused the crash), and it also adds it back in a different way. The "N/A" string was already inside the ".get" function and it remains there! I just added the _() function to it.

It should not cause any problem and I don't see where the problem is.

Dima73:
Really.

PR #2382, opened 2019-12-14
Update bg.po

marto74bg:
Amended translation.


WanWizard:
Translations only to the develop branch please.

PR #2391, opened 2019-12-21
[SatelliteEquipmentControl] Further improvements and reverted some previous changes

nautilus7:
1. Restored screen title to the previous/original one that was wrongly changed in previous commit. The plugin is called "Satellite equipment setup" in every string the user sees, including the menu entries. The "Satellite equipment control" name is only used in the python filename and some variables, which is irrelevant. By restoring the original screen title, we keep the existing translations intact as well.

2. Restored green button's text to "OK", which is what openpli uses almost everytime in the "setup" alike screens for saving settings.

3. Removed unused variable in "resetDefaults" function. Just get the 2nd element of the tuple and get done with it...

4. Got rid of the nasty global "session" variable used in plugin startup by defining the "confirmed" function inside the "SecSetupMain".

5. Removed unused variable "show" and made some cosmetic changes in "SecSetupStart" function.


nautilus7:
@Huevos I didn't touch any of the fixes you made, so I hope we are ok.

PR #2392, opened 2019-12-22
[skin.py] Resequence code

IanSav:
This change starts the process of making "skin.py" more comparable between images.

This change includes the following:
- Sort the includes alphabetically.
- Sort the parameters alphabetically.
- Move all global list and dictionary definitions to a group near the start of the file.
- Make "display_skin_id" uppercase to "DISPLAY_SKIN_ID" to indicate that this is a statically assigned value.
- Add "GUI_SKIN_ID" definition to replace the magic number 0 used in the code.
- Improve and add more comments.
- Relocate some methods for clarity or to get them out of the way.

These changes make no functional changes to this code.  This is part of a progression to improving the code and making it more comparable to other images.

### DIFF
--- a/skin.py
+++ b/skin.py
@@ -1,47 +1,69 @@
 from Tools.Profile import profile
 profile("LOAD:ElementTree")
-import xml.etree.cElementTree
 import os
+import xml.etree.cElementTree
 
 profile("LOAD:enigma_skin")
-from enigma import eSize, ePoint, eRect, gFont, eWindow, eLabel, ePixmap, eWindowStyleManager, addFont, gRGB, eWindowStyleSkinned, getDesktop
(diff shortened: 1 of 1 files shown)


skin.py
```diff
@@ -1145,3 +1149,11 @@ def process_panel(widget, context):
 	# things around.
 	screen = None
 	visited_components = None
+
+def dump(x, i=0):
```
WanWizard, line 1153: What is the purpose of `dump()`?

IanSav: I think this was some old debugging code.  I don't think it is used so I moved it to the end of the file.  If it is truly not used it can be pushed off the end.  ;)

PR #2397, opened 2019-12-25
Add option "Default timer type" when create new timer

Dima73:


littlesat:
Please do not commit... just recoding should be default no real need to a a default setting...

Dima73:
Let's discuss it.
My default option is to Zap so as not to miss the event.

littlesat:
Then better have an option to disable recoding from the recoding menu which change the default to zap. (And add that option also to the setup wizard) and the also block all the recoding options. And/or check for a device having more than x Gig and when that is not there change the default without setting (maybe warning that you can ignore)

pr2git:
No please stop thinking of an automatic way to decide what the default action should be. You won't find any valid solution. So please, let to end user the opportunity to choose what is his favorite default action when setting up a recording. Simple and efficient.

Dima73:
<item level="2 -->it means user level Expert

nautilus7:
What changed and you re-opened this?

btw, expert user level should give more control to the user. Not less, by automating settings.

WanWizard:
It was decided to accept this for now, and do a bit of finetuning later.

PR #2405, opened 2020-01-06
[skin.py] Add findSkinScreen() method

IanSav:
- This method allows code to determine if a screen is skinned. If there are multiple screen names possible then the code reports which skin based screen would be used by the skin engine.

- Further improve some of the log messages.


Huevos:
Thanks. This method is extremely useful.

PR #2410, opened 2020-01-11
Directories.py updates

IanSav:
- [Directories.py] Update try/except for Python 3

  Update the try/except syntax to a format ready for any future Python 3 upgrades.

- [Directories.py] Add two new scopes

  Add two new scopes to add more options for developers to abstract their code.

  SCOPE_CURRENT_PLUGIN_ABSOLUTE returns the absolute path of the current plugin if called from within a plugin.  If this scope is called from code other than a plugin it returns None.

  SCOPE_CURRENT_PLUGIN_RELATIVE returns the relative path of the current plugin if called from within a plugin.  The path fragment returned is relative to SCOPE_PLUGINS.  Again, if this scope is called from code other than a plugin it returns None.

  Using os.path.join(resolveFilename(SCOPE_PLUGINS), resolveFilename(SCOPE_CURRENT_PLUGIN_RELATIVE)) the result should be the same as resolveFilename(SCOPE_CURRENT_PLUGIN_ABSOLUTE).

  These scopes can also be used with base data to create paths within the current plugin.

  At the moment SCOPE_CURRENT_PLUGIN returns the same path as SCOPE_PLUGINS.  In an ideal world the new SCOPE_CURRENT_PLUGIN_ABSOLUTE should replace the existing SCOPE_CURRENT_PLUGIN.  This is not currently possible as SCOPE_CURRENT_PLUGIN is in common use.

### DIFF
--- a/lib/python/Tools/Directories.py
+++ b/lib/python/Tools/Directories.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import errno
+import inspect
 import os
 
 from enigma import eEnv, getDesktop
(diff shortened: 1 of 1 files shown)


IanSav:
Is there a reason why this isn't being merged?

PR #2417, opened 2020-01-26
Fix subtitle runahead on recordings

Huevos:
Some receivers have a bug in the file drivers that deliver subtitles on playback of records completely out of sync with the video, making them unusable. This PR slightly adjusts the bad timing handler to treat subtitles that have been delivered too early by the filesystem as valid and to queue them up for showing later at the appropriate time. The handling of subtitles that arrive too late remains unchanged.

### DIFF
--- a/lib/gui/esubtitle.cpp
+++ b/lib/gui/esubtitle.cpp
@@ -119,7 +119,7 @@ void eSubtitleWidget::setPage(const eDVBTeletextSubtitlePage &p)
 
 void eSubtitleWidget::setPage(const eDVBSubtitlePage &p)
 {
-	eDebug("[eSubtitleWidget] setPage");
+	//eDebug("[eSubtitleWidget] setPage");
 	m_dvb_page = p;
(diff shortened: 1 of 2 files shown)


Huevos:
https://mega.nz/#F!B7gVCCBZ!InGsVP5DEBQc1A89NwzEIw

Huevos:
Please look at the video.

https://mega.nz/#F!B7gVCCBZ!InGsVP5DEBQc1A89NwzEIw

littlesat:
Why did you remove the abs? I still don’t understand?

WanWizard:
abs() converts a negative to a positive.

If pos-p.m_show_time is between -1800000 and 0, the abs() is pointless there. If it is < -1800000, the abs() changes the result of the if by suddenly including it.

littlesat:
It checks for out of sync and then both negative and positive make a point.... and now one half is removes. So I still do not understand what is tried to archive here

PR #2422, opened 2020-01-26
Directories.py: Get rid of deprecated scope functions.

persianpros:
for SCOPE_SKIN_IMAGE use SCOPE_CURRENT_SKIN instead.
for SCOPE_ACTIVE_SKIN use SCOPE_CURRENT_SKIN instead.
for SCOPE_ACTIVE_LCDSKIN use SCOPE_CURRENT_LCDSKIN instead.

With this method all the old things work normally as it's backward-compatible but at least we won't have duplicates and check them again and again.

Also I did sort things again.

### DIFF
--- a/lib/python/Tools/Directories.py
+++ b/lib/python/Tools/Directories.py
@@ -14,30 +14,29 @@ SCOPE_TRANSPONDERDATA = 0
 SCOPE_SYSETC = 1
 SCOPE_FONTS = 2
 SCOPE_SKIN = 3
-SCOPE_SKIN_IMAGE = 4  # DEBUG: How is this different from SCOPE_SKIN?
-SCOPE_USERETC = 5  # DEBUG: Not used in Enigma2.
-SCOPE_CONFIG = 6
-SCOPE_LANGUAGE = 7
-SCOPE_HDD = 8
-SCOPE_PLUGINS = 9
-SCOPE_MEDIA = 10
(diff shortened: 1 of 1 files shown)


persianpros:
We use this in Open Vision for months and so far everything is ok.

WanWizard:
@littlesat ?

Huevos:
Please don't accept this commit, thanks. We get things so we can share from all images and now he wants to delete it just because . What is the hurry to deprecate these functions.

This is not backwards compatible. It breaks any plugin still using the old scopes.

persianpros:
@Huevos 

I saw your message that it will break old things, did you test it?

Huevos:
Please do not accept this commit. It renumbers the scopes. Any code using scope numbers and not variable names will be broken by this change.

IanSav:
Removing the old scopes is very unwise.  You don't know what code is out there that may still be using them.

If you want to remove old scopes then prominently publish an alert that the changes will be made at some future time so that all coders have an opportunity to adjust their code before the change happens.  Keep in mind that some source code is closed source and other code is no longer maintained.  Even with a prominent notice there may still be problems if this is done.

IanSav:
Similarly renumbering scopes is dangerous.  If you want to sort the scope names alphabetically then that is one thing.  Renumbering them is something else and also potentially very dangerous.

Issue #2416, opened 2020-01-23
Some leftover /usr/lib/... hard-coded paths - Can these be fixed?


nautilus7:
I am trying to eliminate all `/usr/lib` hard-coded paths from enigma2, so it is compatible with multi lib images (that support 64bit and 32bit binaries and executables and have both `/usr/lib/` and `usr/lib64` directories).

I [fixed several of them](https://github.com/OpenPLi/enigma2/commit/b6de676dab8dab1720e9c32fdb820cadecbb7320) yesterday, but there are 3 cases left with such hard-coded paths:

1. https://github.com/OpenPLi/enigma2/blob/develop/mytest.py#L3 I have no clue what this is and how to fix it. I guess this is not that important for eveyday use. It was added with this commit https://github.com/OpenPLi/enigma2/commit/0e003cd9c9f2f15261eb81d374f99454637a1442

2. https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Components/About.py#L14 I think that ipkg is obsolete and the whole "else" clause can be removed together with the path. So, it seems trivial to fix.

3. https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Plugins/Extensions/TuxboxPlugins/plugin.py#L9 What is this plugin? Is it still used or it's a leftover piece from old code? [Modification history](https://github.com/OpenPLi/enigma2/commits/develop/lib/python/Plugins/Extensions/TuxboxPlugins/plugin.py) of this file seems ancient...

WanWizard:
1. @littlesat ?

1. ipkg is indeed obsolute, the else can be  removed.

1. I think this link is wrong?  Other than that, I did some Googling, and I can only find references to cdk images (Enigma1).

nautilus7:
@WanWizard sorry, link in case 3 fixed.

WanWizard:
Thanks. 

Perhaps @littlesat knows more about the background, but in my 14 years PLi I've never seen an E2 Tuxbox plugin, so no idea why that is still there...

nautilus7:
Yep, same here in my 8 years of using openpli/satdreamgr.

AbuBaniaz:
Does this help?
https://github.com/OpenVisionE2/enigma2-openvision/commit/d876625d80d1fba22768c3465858747a1df30fcc

nautilus7:
SCOPE_LIBDIR would be a nice addition and a useful tool in some cases. Thanks.

[...]

PR #2423, opened 2020-01-26
Remove some ancient things

persianpros:
Read: https://github.com/OpenPLi/enigma2/issues/2416


WanWizard:
@littlesat ? This is above my paygrade...

littlesat:
I'm a bit doupting about ipkg->opkg as it might break some plugins that use it... 
But I think it is indeed better... And I could not find any plugins using it...

PR #2431, opened 2020-01-29
[SkinSelector.py] Adjust resolution search code

IanSav:
Change the way skin resolutions are obtained.  Now we will do an XML lookup for the resolution and if that fails we will revert to a text search.

This code is slower to execute than its predecessor but hopefully addresses Littlesat's reservations about the hack nature of the previous code.

### DIFF
--- a/lib/python/Screens/SkinSelector.py
+++ b/lib/python/Screens/SkinSelector.py
@@ -3,7 +3,7 @@ import xml.etree.cElementTree
 
 from enigma import ePicLoad, getDesktop
 from os import listdir
-from os.path import dirname, exists, isdir, join as pathjoin
+from os.path import dirname, exists, isdir, join as pathjoin, split as pathsplit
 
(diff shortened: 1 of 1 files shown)


IanSav:
I would like to close this pull request.  It is a performance compromise.

Comments on the forum agree that processing the file as XML is very costly and not really justifiable for the purpose of finding the skin resolution.  None of the XML structure will ever be used for any other purpose.  This is a waste of CPU time and temporary memory storage.

IanSav:
I am closing this pull request as I can't justify the waste of resources to process the skin as XML just to get a resolution and then throw the rest of the work away.

I understand the desire for coding purity but we really need to consider system performance is part of the purity consideration.

I will resubmit a new pull request with the original code design but I will remove the 64K file read limit.  I will stop reading the file as soon as I find a resolution.  I will *not* be reading the file a line at a time as this is still a little more costly in terms of system performance.  Block reads are always faster.

PR #2432, opened 2020-01-30
[SkinSelector.py] Optimise resolution search

IanSav:
- Remove all trace of the proposed XML processing of the skin to find the resolution.
- Rework the skin search code to simplify and speed up the search.


littlesat:
Nice!!!!! Conplements

PR #2436, opened 2020-01-31
Fixed default (emergency) skin not showing

nautilus7:
Apparently just loading "/usr/share/enigma2/skin_default.xml" is not enough. The "/usr/share/enigma2/skin.xml" is also needed.
This commit combines these 2 xml files and puts the code inside the "/usr/share/enigma2/skin_default/skin.xml", which is what is loaded by skin.py.
Putting the code directly inside "/usr/share/enigma2/skin_default/skin.xml" is way simpler and less prone to (future) errors than having them as "includes" in it.


AbuBaniaz:
Doesn't the makefile within the skin_default folder need updating to include skin.xml too?

nautilus7:
There is already a skin.xml in there which is also present in the makefile.am

https://github.com/OpenPLi/enigma2/blob/develop/data/skin_default/Makefile.am#L55

nautilus7:
Your confusion proves my point that the current layout (code) is too complicated...

AbuBaniaz:
The idea is to make every skin self-contained including the the emergency skin, which you have so nicely done. On OE-A, the skin is shared among distros. Not all have the the new skin code so cannot be done yet.

littlesat:
Merge?

AbuBaniaz:
Yes please.

IanSav:
I wasn't able to do this combination on the other images as the emergency skin is shared between images.  As OpenPLi has its own, private, emergency skin this should be fine.  In fact the original skin.xml and skin_default.xml in /usr/share/enigma2 can now be deleted!

nautilus7:
I see. Aren't they deleted with my commit or I missing something?

IanSav:
At the time I posted I hadn't looked.  My OpenPLi machine is dead so I have to boot OpenPLi on my OpenViX development machine from time to time to check out changes.

PR #2441, opened 2020-02-01
Multiboot slight rename

AbuBaniaz:


nautilus7:
I am not sure I agree with the renaming, but I would certainly prefer "Multiboot image selector" without upper case first letters (to be consistent with what is used in openpli in general).

nautilus7:
Thanks.

PR #2456, opened 2020-02-16
[SystemInfo.py] PEP8 and other cleanups

IanSav:
- PEP8 clean up of the code.
- Remove unused import of os.
- Add comment to highlight the importance of the current import order.
- Use uniform quotation marks.

### DIFF
--- a/lib/python/Components/SystemInfo.py
+++ b/lib/python/Components/SystemInfo.py
@@ -1,47 +1,52 @@
+import re
+
 from enigma import eDVBResourceManager, Misc_Options, eDVBCIInterfaces, eGetEnigmaDebugLvl
+
 from Components.Console import Console
 from Tools.Directories import fileExists, fileCheck, pathExists, fileHas, resolveFilename, SCOPE_PLUGINS
 from Tools.HardwareInfo import HardwareInfo
-import os, re
 
 SystemInfo = {}
 
-from Tools.Multiboot import getMultibootStartupDevice, getMultibootslots
+from Tools.Multiboot import getMultibootStartupDevice, getMultibootslots  # This import needs to be here to avoid a SystemInfo load loop!
+
+# Parse the boot commandline.
+#
+fd = open("/proc/cmdline", "r")
+cmdline = fd.read()
+fd.close()
+cmdline = {k: v.strip('"') for k, v in re.findall(r'(\S+)=(".*?"|\S+)', cmdline)}
 
 def getNumVideoDecoders():
-	number_of_video_decoders = 0
-	while fileExists("/dev/dvb/adapter0/video%d" % (number_of_video_decoders), 'f'):
-		number_of_video_decoders += 1
-	return number_of_video_decoders
+	numVideoDecoders = 0
+	while fileExists("/dev/dvb/adapter0/video%d" % numVideoDecoders, "f"):
+		numVideoDecoders += 1
+	return numVideoDecoders
 
 def countFrontpanelLEDs():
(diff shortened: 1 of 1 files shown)


IanSav:
Would it be acceptable to also sort the SystemInfo entries alphabetically?

IanSav:
Oops, I forgot to hold back pull requests for OpenPLi.  If this pull request gets lost in the release candidate preparation I can recreate it.

athoik:
Why using open, read, close?

Did close you notice an issue with close missing? Leaking file descriptors?

IanSav:
When I see an open without a close I assume that there is probably going to be a file handle and memory leak.  I applied a least difference fix.

The "with" version of the open has an implied close.  This is a plain open() so there is no automatically associated close().

athoik:
It was suggested long time ago by @MiLo not to use open,read,close. But simple open,read. The file handle will close automatically once python perform garbage collection.

But if you feel that's wrong, create a POC that proves the file handle leak.

IanSav:
SystemInfo is not some transient code.  The module is held open for the entire Enigma2 session.  Why keep the file handle open all that time for nothing?  It is a waste or resources.

Don't rely on garbage collection.  That is not quality code.

Huevos:
Sort alphabetically doesn't make sense to me. This file is grouped by items, e.g. display, etc. Sorting alphabetically will just randomize the content.

IanSav:
You mean it isn't already random?

If there is an order in these entries then perhaps it should be specifically documented.  Otherwise sorting the entries will make them much easier to find.

Huevos:
Currently the items are grouped into features such as "display". Sorting alphabetically would break that. File does need a good clean up with labels, but not an alphabetic sort which would de-group the items further.

athoik:
> Don't rely on garbage collection. That is not quality code.

No you are wrong. open.read is freed immediately the it goes out of scope. 

Use python and do the following:

```
>>> import os
>>> os.getpid()
PID
>>> data = open("/proc/cpuinfo").read()
```

Then run lsof -p PID and see if /proc/cpuinfo is open.

```
>>> f = open("/proc/cpuinfo")
>>> data = f.read()
```

Then run lsof -p PID and see if now /proc/cpuinfo is open.

Finally close handle.

```
>>> f.close()

```

Again run lsof -p PID and check if /proc/cpuinfo is gone. 

In conclusion open.read is just fine. 

Just restore it.

IanSav:
Okay, I ran the tests you suggested and I agree that the open.read does appear to close the file.  I strongly disagree that this is a garbage collection process.  This appears to be the way the nested functions operate.

Garbage collection runs when the method or module exits and not after every command.  So your assertion that the file is being closed by a garbage collection process is what I don't accept.  In the case of this code (SystemInfo) the code is loaded and never exits until Enigma2 itself stops.  Thus the garbage collection won't run to clean up SystemInfo until Enigma2 itself stops.

Given what you have highlighted I will update the code.

Huevos:
Ian, the way I see it you are not assigning a file handle, so nothing to go "out of scope".

It is a valid way to read/write a file, and impossible to call a "close" due to there not being a file handle, so it must close the file immediately. Anything other would be a bug in Python itself, not the script.

athoik:
> Using CPython, your file will be closed immediately after the line is executed, because the file object is immediately garbage collected

https://stackoverflow.com/questions/8011797/open-read-and-close-a-file-in-1-line-of-code/8011836

https://stackoverflow.com/a/8011863

Well I guess MiLo suggest this change in Enigma2 because we are using CPython.

Above comment still mentions gc ;)

IanSav:
I will submit the "with" option and then everyone should be happy.  ;)

PR #2466, opened 2020-02-24
Add debug output to listbox python buildfunc calls

AbuBaniaz:
This change adds exception stack output
when a Python buildfunc throws an exception
when a button press that opens a screen throws an exception

By @simoncapewell


AbuBaniaz:
Not sure why this failed the Travis check. It was fine on Vix. https://github.com/OpenViX/enigma2/pull/501

WanWizard:
It's a problem with Travis CI (the apt-get fails for all tests)

PR #2472, opened 2020-03-01
keyboard.xml - translate OPTION buton to HELP for Gigablue

ims21:
- temporary - it should be solved in driver with manufacturer

### DIFF
--- a/data/keymap.xml
+++ b/data/keymap.xml
@@ -831,4 +831,9 @@
 		<key id="KEY_VOD" mapto="vod" flags="b"/>
 		<key id="KEY_VOD" mapto="vod_long" flags="l"/>
 	</map>
+	<translate>
+		<device name="gigablue remote control">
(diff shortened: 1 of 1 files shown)


littlesat:
Is this if gigablue then do something else thing really needed?

littlesat:
I mean do we really need if device is gigablue?

ims21:
I don't care. I want not to have it in userfile keytranslate.xml, where they added it and what was reverted by WanWizard fortunately (because it replaced user file)

GB has not Help button on RC, afaik.

littlesat:
Other boxes have that other key... then just add it....

ims21:
lets manufacturer change code "option button" to "help button" in driver for e2

ims21:
> 
> 
> Other boxes have that other key... then just add it....

I do not understand ...

littlesat:
Just remove the part that checks for the GB remote controll....

PR #2480, opened 2020-03-07
ImportChannel - cosmetic in dvb-t region code

ims21:


littlesat:
See my first comment.... but I have no time now.... you should involve epg.dat locaction/file here

PR #2488, opened 2020-03-14
[Timezones.py] Refactor and optimise the code

IanSav:
- [Timezones.py] Refactor and optimise the code
  - Clean up and optimise the code.
  - Add access to geolocation data to establish the initial guess of the user's timezone.
  - Don't change the default timezone when navigating through available timezone lists.
  - Ignore the existence of alternate txdata from "posix" or "right" packages.
  - Improve the configuration comments.
  - Correct name display of "Dumont d'Urville".
  - Use the 2019c version of "timezone.xml" file to offer users a classic view of timezone selections.  This list is the default for OpenATV and is available under the "Timezone area" of "Classic" if the new "timezone area" and "Timezone" style UI is adopted by OpenATV.
  - Add callbacks for before and after the timezone is set.
  - Separate the AutoTimer code out of the Timezones code.  Move the AutoTimer code to the end of this code and add the hooks as a template until the AutoTimer code is updated to use the new callback code.

- [Geolocation.py] New tool to provide geolocation data

- [mytest.py] Support refactored Timezones.py
  - Activate geolocation initialisation
  - Create separate timezone initialisation

- [SetupDevices.py] Remove timezone initialisation

- [setup.xml] Add support for refactored Timezones.py

  This chane splits the timezone selection from one massive GMT ordered list into a set or areas or regions in the world followed by a selection list of timezones within that area or region.

  The previous timezone selection list is still available if the user selects the "Classic" area.

### DIFF
--- a/data/setup.xml
+++ b/data/setup.xml
@@ -1,6 +1,7 @@
 <setupxml>
 	<setup key="timezone" title="Timezone">
-		<item level="0" text="Timezone">config.timezone.val</item>
+		<item level="0" text="Timezone area" description="Select timezone area or region.">config.timezone.area</item>
+		<item level="0" text="Timezone" description="Select timezone in the area or region.">config.timezone.val</item>
 	</setup>
 	<setup key="avsetup">
 		<!-- this is just a placeholder, the Videomode plugin implements this submenu -->
--- a/lib/python/Components/SetupDevices.py
+++ b/lib/python/Components/SetupDevices.py
@@ -1,15 +1,14 @@
 from Components.config import ConfigOnOff, ConfigSelection, ConfigSubsection, ConfigText, config
-from Components.Timezones import timezones
-from Components.Language import language
 from Components.Keyboard import keyboard
+from Components.Language import language
 
 def InitSetupDevices():
-	def timezoneNotifier(configElement):
-		timezones.activateTimezone(configElement.index)
+	def languageNotifier(configElement):
+		language.activateLanguage(configElement.value)
 
-	config.timezone = ConfigSubsection()
-	config.timezone.val = ConfigSelection(default=timezones.getDefaultTimezone(), choices=timezones.getTimezoneList())
-	config.timezone.val.addNotifier(timezoneNotifier)
+	config.osd = ConfigSubsection()
+	config.osd.language = ConfigText(default="en_EN")
+	config.osd.language.addNotifier(languageNotifier)
 
 	def keyboardNotifier(configElement):
 		keyboard.activateKeyboardMap(configElement.index)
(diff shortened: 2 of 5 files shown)


WanWizard:
````
  File "/usr/lib/enigma2/python/mytest.py", line 23, in <module>
    import Tools.Geolocation
ImportError: No module named Geolocation
````
what is missing?

WanWizard:
Missing in the makefile. Fix is committed.

WanWizard:
@IanSav can you first fix the autotimer issues in this commit? 

It is a lot more important than LoadSkin(), this is causing an enigma crash...

IanSav:
Can I please have a log file with the errors being experienced.

WanWizard:
Reported on the forum.

change in timezone setting:
````
  File "/usr/lib/enigma2/python/Components/Timezones.py", line 84, in timezoneNotifier
    timezones.activateTimezone(configElement.value, config.timezone.area.value)
  File "/usr/lib/enigma2/python/Components/Timezones.py", line 253, in activateTimezone
    method()
  File "/usr/lib/enigma2/python/Components/Timezones.py", line 313, in autotimerBefore
    self.autotimerPoller.stop()
  File "/usr/lib/enigma2/python/Plugins/Extensions/AutoTimer/AutoPoller.py", line 164, in stop
    self.thread.stop()
AttributeError: 'NoneType' object has no attribute 'stop'
````
Periodicaly cca each 3-4? minuts
````
  File "/usr/lib/enigma2/python/Components/Timezones.py", line 352, in autotimeQuery
    self.autotimerTimer.parseEPG(autoPoll=True)
TypeError: parseEPG() got an unexpected keyword argument 'autoPoll'
`````

IanSav:
Hmmm, is OpenPLi using the OE-Alliance AutoTimer or is there a private version?

IanSav:
This patch should make OpenPLi bootable again.

[https://github.com/OpenPLi/enigma2/pull/2490](https://github.com/OpenPLi/enigma2/pull/2490)

WanWizard:
Also, my test box was changed to Europe/Berlin for some reason, while before the update, it was set to:
````
lrwxrwxrwx    1 root     root            33 Mar 13 21:06 /etc/localtime -> /usr/share/zoneinfo/Europe/Dublin
````

WanWizard:
> Hmmm, is OpenPLi using the OE-Alliance AutoTimer or is there a private version?

No, it is the other way around, we use the original version, OE-A has a forked version.

https://github.com/OpenPLi/enigma2-plugins/tree/master/autotimer

IanSav:
Okay, as a future project hopefully we can bring the two back together.

Now that I know there are two versions I will have to work out the differences.  For now I don't expect any issues with the timezone updates not being communicated to the AutoTimers as this never happened before.

WanWizard:
I wonder why it is needed at all, autotimers in OpenPLi use timestamps, which are always in UTC, no matter what timezone the box is in?

IanSav:
As I didn't realise that the AutoTimers were different I am at a loss to know how they are different.  Maybe the OE-Alliance version is missing some code from the OpenPLi version and hence this code hack.

I will look at this some more when I get some sleep.

IanSav:
All the AutoTimer code has been removed from the OpenPLi version of the code.  This is done in Pull [Request #2491](https://github.com/OpenPLi/enigma2/pull/2491).

PR #2490, opened 2020-03-14
[Timezones.py] Remove AutoTimer integration

IanSav:
This should resolve a crashing issue until I can determine why the AutoTimer interface isn't working on OpenPLi.


WanWizard:
My guess: you're probably looking  at a forked version of Autotimer...

PR #2500, opened 2020-03-21
[RedirectOutput.py] Refactor code

IanSav:
Remove code duplication by using sub classes.

### DIFF
--- a/lib/python/Tools/RedirectOutput.py
+++ b/lib/python/Tools/RedirectOutput.py
@@ -1,20 +1,20 @@
 import sys
+
 from enigma import ePythonOutput
 
-class EnigmaLogDebug:
 
-	lvlDebug = 4
-
-	def __init__(self):
-		self.line = ''
+class EnigmaLog:
+	def __init__(self, level):
+		self.level = level
+		self.line = ""
 
 	def write(self, data):
 		if isinstance(data, unicode):
-			data = data.encode("UTF-8")
+			data = data.encode(encoding="UTF-8", errors="ignore")
 		self.line += data
-		if '\n' in data:
-			ePythonOutput(self.line, self.lvlDebug)
-			self.line = ''
+		if "\n" in data:
+			ePythonOutput(self.line, self.level)
+			self.line = ""
 
 	def flush(self):
 		pass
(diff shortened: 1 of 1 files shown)


lib/python/Tools/RedirectOutput.py
```diff
@@ -1,47 +1,37 @@
...
-			ePythonOutput(self.line, self.lvlError)
-			self.line = ''
 
-	def flush(self):
-		pass
+class EnigmaLogFatal(EnigmaLog):
+	def __init__(self):
+		EnigmaLog.__init__(self, 1)  # lvlError = 1
 
-	def isatty(self):
```
athoik, line 43: I see that isatty removed.

We don't get that error any more? 

```
AttributeError: EnigmaOutput instance has no attribute 'isatty'
```

https://github.com/OpenPLi/enigma2/commit/993e37993b9f4359fdb7498bf1e61821d9e64c6b#diff-b940b4d0317bd2138406237ae954eef4

IanSav: It shouldn't have been removed.  It is part of the parent class.

athoik: Yes you are right.

Just note that any plugin running from enigma2 it will get the modified sys.stdout.

The origin looks more or less like this:
```
>>> dir(sys.stdout)
['__class__', '__delattr__', '__doc__', '__enter__', '__exit__', '__format__', '__getattribute__', '__hash__', '__init__', '__iter__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', 'close', 'closed', 'encoding', 'errors', 'fileno', 'flush', 'isatty', 'mode', 'name', 'newlines', 'next', 'read', 'readinto', 'readline', 'readlines', 'seek', 'softspace', 'tell', 'truncate', 'write', 'writelines', 'xreadlines']
```

Any of the missing functions of modified stdout is possible to cause issues, like the one mentioned on the commit.

IanSav: I simply refactored the code that was there to remove the duplication.  If the new code can be improved please update it.  I will copy any improvements to the other images.

athoik: I mention it since you are up to it.

Plugins might fail silently due to "stealing sys.stdout" function :)

Sure code duplication removed, so that should fine to merge it.

IanSav: When I refactor code I keep the external functionality the same just improve the internal code.

Would the plugins fail or would their output simply go to the log file?  I haven't really thought about it.  Code that outputs to a pipe seems to be working fine.

Anyway if anyone improves this code my image matching system should pick up the change any I will replicate the improvements to the other images.

IanSav: By the way, I do appreciate people checking my code and questioning anything with which they have a concern.  :)

PR #2501, opened 2020-03-22
More time zone code clean up

IanSav:
[Timezones.py] More code clean up
- Allows the time zone in the "settings" file to override the "/etc/localtime" time zone setting.  This will allow users to restore their "settings" file from a backup and have the Linux, glib, TZ and Enigma2 time zones all be matched.  (This was discussed in the forum and seen as desirable.)
- Add "/etc/timezone" to the settings updated when the time zone changes. (This is used by Java and some other software.)
- Correct spelling of "timezone" to "time zone" for displayed text.

[setup.xml] Correct and improve time zone text
- Correct the spelling of "timezone" to "time zone" and make the help text a little more clear.

### DIFF
--- a/lib/python/Components/Timezones.py
+++ b/lib/python/Components/Timezones.py
@@ -10,14 +10,14 @@ from Tools.StbHardware import setRTCoffset
 
 # The DEFAULT_AREA setting is usable by the image maintainers to select the
 # default UI mode and location settings used by their image.  If the value
-# of "Classic" is used then images that use the "Timezone area" and
-# "Timezone" settings will have the "Timezone area" set to "Classic" and the
-# "Timezone" field will be an expanded version of the classic list of GMT
-# related offsets.  Images that only use the "Timezone" setting should use
-# "Classic" to maintain their chosen UI for timezone selection.  That is,
+# of "Classic" is used then images that use the "Time zone area" and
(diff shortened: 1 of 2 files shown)


lib/python/Components/Timezones.py
```diff
@@ -290,6 +284,12 @@ def activateTimezone(self, zone, area, runCallbacks=True):
 		except (IOError, OSError) as err:
 			print "[Timezones] Error %d: Linking '%s' to '/etc/localtime'! (%s)" % (err.errno, file, err.strerror)
 			pass
+		try:
+			with open("/etc/timezone", "w") as fd:
+				fd.write("%s\n" % tz)
+		except (IOError, OSError) as err:
+			print "[Timezones] Error %d: Updating '/etc/timezone'! (%s)" % (err.errno, err.strerror)
+			pass
```
athoik, line 292: Pass is a no-op (no operation). You need to use when you don't have any other commend to use.
Since you are using print above pass, pass I think is useless.

IanSav: You are absolutely correct.  That is a left over from when I thought I would ignore the error.  I will remove the "pass" when I next update the code and if there are no more changes needed I will remove it when I have some free time.

PR #2503, opened 2020-03-22
all po files updated from develop at 2020-03-22 19:15

pr2git:
hr.po and lt.po included


littlesat:
Wierd I cannot merge this one...

WanWizard:
Conflict in po/nb.po

You merged in the wrong order, you should have merged this one first, before the Norsk Bokmal PR...

pr2git:
Yes but then the Norsk Bokmal will also get the problem.
I already create a new PR.

PR #2513, opened 2020-03-29
[esubtitle.cpp] Fix some HTML in subtitles

IanSav:
- Subtitles containing `&gt;` or `&lt;` were leaving behind the semicolon.
- Remove `<font...>` and `</font>` tags.

This fix by Adoxa taken from the Beyonwiz image.

### DIFF
--- a/lib/gui/esubtitle.cpp
+++ b/lib/gui/esubtitle.cpp
@@ -355,8 +355,8 @@ int eSubtitleWidget::event(int event, void *data, void *data2)
 				text = replace_all(text, "&apos;", "'");
 				text = replace_all(text, "&quot;", "\"");
 				text = replace_all(text, "&amp;", "&");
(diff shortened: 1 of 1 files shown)


athoik (changes requested):
Please use size_t instead of unsigned.

http://www.cplusplus.com/reference/string/string/find/

IanSav:
I am not a C++ programmer.  I simply copied the corrections from the Beyonwiz image.  If further changes are required then they will have to be made by someone better able to make and test the changes..

athoik:
1. Make the unsigned type into size_t 
2. Push the commit
3. Check CI build, should be successful

IanSav:
The two instances changed as requested.

I assume that is what you wanted.

athoik:
Yes, now type is the proper one!

PR #2527, opened 2020-04-12
[skin.py] More improvements

IanSav:
- Introduce code to allow a future where the skin can be reloaded without requiring a GUI restart:
  - Add callback infrastructure for skin to notify other code when the skin has changed.
  - Add a method to make the skin reload the currently nominated skin.
- Add a mechanism to update the windowstyle to take account of all skin updates that occur after the windowstyle was first loaded.
- Reduce the scope of the "os" import.
- Provide entry points for both the old and new mytest.py versions.
- Add font recognition to the "parseParameter()" method.  (This is a feature from OpenATV.)
- Add some more animationMode keywords used by OpenATV.
- Add more traps for potential code failure.
- Improve more logging text.
- Add a new method "getSkinFactor()" (from OpenATV) to report the skin scaling factor.
- Apply further code clean ups.
- The strip() method code was accidentally lost from the parameter parsing.  This resulted in parameter lists separated by commas *and* white space would fail.
- Use isfile() as a test for existence because, in context, it is better to know that the path exists and is a file rather than just knowing the path exists.

### DIFF
--- a/skin.py
+++ b/skin.py
@@ -1,16 +1,14 @@
-from Tools.Profile import profile
-profile("LOAD:ElementTree")
 import errno
-import os
 import xml.etree.cElementTree
 
-profile("LOAD:enigma_skin")
 from enigma import addFont, eLabel, ePixmap, ePoint, eRect, eSize, eWindow, eWindowStyleManager, eWindowStyleSkinned, getDesktop, gFont, getFontFaces, gRGB
+from os.path import basename, dirname, isfile
+
 from Components.config import ConfigSubsection, ConfigText, config
 from Components.RcModel import rc_model
 from Components.Sources.Source import ObsoleteSource
 from Components.SystemInfo import SystemInfo
-from Tools.Directories import SCOPE_CONFIG, SCOPE_CURRENT_LCDSKIN, SCOPE_CURRENT_SKIN, SCOPE_FONTS, SCOPE_SKIN, SCOPE_SKIN_IMAGE, fileExists, resolveFilename
+from Tools.Directories import SCOPE_CONFIG, SCOPE_CURRENT_LCDSKIN, SCOPE_CURRENT_SKIN, SCOPE_FONTS, SCOPE_SKIN, SCOPE_SKIN_IMAGE, resolveFilename
 from Tools.Import import my_import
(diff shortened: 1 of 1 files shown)


WanWizard:
@IanSav did you see the comment on the commit?

WanWizard:
OpenPLi already crashes when starting:
````
  File "/usr/lib/enigma2/python/mytest.py", line 26, in <module>
    from Screens import InfoBar
  File "/usr/lib/enigma2/python/Screens/InfoBar.py", line 5, in <module>
  File "/usr/lib/enigma2/python/Screens/MovieSelection.py", line 1, in <module>
  File "/usr/lib/enigma2/python/Screens/Screen.py", line 4, in <module>
  File "/usr/lib/enigma2/python/Components/GUISkin.py", line 1, in <module>
  File "/usr/lib/enigma2/python/Components/GUIComponent.py", line 1, in <module>
  File "/usr/lib/enigma2/python/skin.py", line 4, in <module>
    from enigma import addFont, eLabel, ePixmap, ePoint, eRect, eSize, eWindow, eWindowStyleManager, eWindowStyleSkinned, getDesktop, gFont, getFontFaces, gRGB
ImportError: cannot import name getFontFaces
````
If not fixed in an hour, I'll revert it, so it doesn't break the nightly builds.

IanSav:
Hmmmm,, the font.cpp change must have broken more than I expected.  This appears to work on my version of OpenPLi.  Let me investigate.

Update: Yes the previous version of OpenPLi had the getFontFaces method and it was working!  The font.cpp code has recently been changed.  Has that code been tested?  Those changes are not yet built into code I can test.

IanSav:
This code is still working correctly on the 2020-03-28 build of OpenPLi. There is nothing newer for me to update!

Something has gone seriously wrong with OpenPLi. I believe my code is a victim of the problem and not the cause!

IanSav:
Given that the above issue is not due to my code can this pull request please be restored and activated.

WanWizard is currently unable to revert the reversion.

athoik:
Done!

WanWizard:
Thanks!

IanSav:
Thank you.  :)

PR #2528, opened 2020-04-12
Revert "[skin.py] More improvements"

WanWizard:
Reverts OpenPLi/enigma2#2527

Not tested, crashes Enigma.

### DIFF
--- a/skin.py
+++ b/skin.py
@@ -1,14 +1,16 @@
+from Tools.Profile import profile
+profile("LOAD:ElementTree")
 import errno
+import os
 import xml.etree.cElementTree
 
+profile("LOAD:enigma_skin")
 from enigma import addFont, eLabel, ePixmap, ePoint, eRect, eSize, eWindow, eWindowStyleManager, eWindowStyleSkinned, getDesktop, gFont, getFontFaces, gRGB
-from os.path import basename, dirname, isfile
-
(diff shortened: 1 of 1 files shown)


IanSav:
Hmmmm,, the recent font.cpp change must have broken more than I expected. This appears to work on my version of OpenPLi. Let me investigate.

Update: Yes the previous version of OpenPLi had the getFontFaces method and it was working! The font.cpp code has recently been changed. Has that code been tested? Those changes are not yet built into code I can test.

IanSav:
This code is still working correctly on the 2020-03-28 build of OpenPLi.  There is nothing newer for me to update!

Something has gone seriously wrong with OpenPLi.  I believe my code is a victim of the problem and not the cause!

WanWizard:
Hmm... Could be my bad, I saw the error report, thought I'd try it just before the nightly build run, and it crashed.

I now realize that I'm not home, and the box here runs 7.2-release and not develop, which has probably caused my issue. :cry:

IanSav:
So can this update be restored?

IanSav:
I would appreciate a comment that my code *was* tested and it was okay and the problem was elsewhere.  ;)

WanWizard:
I'm in no position to test, I don't have access to any test equipment. The issue "Boris" reported was definitely elsehwere, as I understand. And my problem was caused by my dumb mistake...

PR #2534, opened 2020-04-21
[Converter/ConditionalShowHide] Allow ConditionalShowHide to work in …

Huevos:
…renderers that currently don't support it

e.g. render=Label".

This means we can now super-impose a text over a background colour and have the colour disappear if no text is present in the text field.

<widget source="key_yellow" render="Label" objectTypes="key_yellow,StaticText" position="710,686" size="240,25" zPosition="1" font="Regular;20" halign="left" transparent="0" valign="center" noWrap="1" foregroundColor="menu-buttons-frgrnd" backgroundColor="green">
	<convert type="ConditionalShowHide"/>
</widget>

Before this commit ConditionalShowHide didn't work. If screen was drawn with no text in the field the screen failed to display any widgets after that point.

Author: @prl001

### DIFF
--- a/lib/python/Components/Converter/ConditionalShowHide.py
+++ b/lib/python/Components/Converter/ConditionalShowHide.py
@@ -14,6 +14,10 @@ class ConditionalShowHide(Converter, object):
 		else:
 			self.timer = None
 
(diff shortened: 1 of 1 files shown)


littlesat:
Nice short merge request! Thanks!

PR #2544, opened 2020-04-29
py files: E101 and E121 PEP8 cleanup

persianpros:
Indentation contains mixed spaces and tabs (E101)
Continuation line under-indented for hanging indent (E121)


persianpros:
@littlesat @WanWizard @IanSav 

Done in all Open Vision repos, if you're up for it let me know to do it for all PLi repos as well.

And later for ViX, ATV and OE-A because it's the standard.

IanSav:
I am more than happy to see all code PEP8 cleaned.

I also standardise quote marks, clean up comments and try to convert variables to camelCase.

littlesat:
I did merge it... But it is such huge it is undoable to check for anomalies. So we might consider to revert it later when it breaks something...
I like the job is done, but it would be nicer that such kind of big topics are better coordinated... :) so we see something like this is coming

persianpros:
@littlesat 

In OV I did this for all repos (except python 3 branches for now) and we're testing it and so far with no problem.

I say accept no more PRs and first test this if it was ok then I will send PRs for all other repos, is it ok?

I hope we could do this once for all.

PR #2575, opened 2020-05-19
Rename Lcd.py to Lcd.py_

builder07:


WanWizard:
Why? It would be handy to provide a decent commit description.

/cc @littlesat

littlesat:
Indeed why? I do not see any reason please explain!

persianpros:
This is 100% wrong.

PR #2577, opened 2020-05-21
Limit geolocation use to first run wizard

IanSav:
- [Geolocation.py] Allow config control of geolocation

  This change adds a config entry "config.misc.enableGeolocation" to control if geolocation is run at startup.  The default is to allow it to run once every boot until the geolocation lookup succeeds and then it will never run again.  It can be enabled for one more run by setting "config.misc.enableGeolocation" to True.  It should be noted that the previous code would not allow geolocation to run more than once in any boot session.  This case is now clearly stated with an appropriate log message.

  In the future "config.misc.enableGeolocation" will not be forced to False on a successful lookup but will allow the user to specifically control if geolocation is permitted to run more frequently.

- [Timezones.py] Update to match changed geolocation

  This change accounts for the change in geolocation execution in that geolocation can now only run once on the first start.  This allows the initial timezone selection to be suggested based on the geolocation data.  If geolocation is unavailable or if a proxy is detected then the image based defaults will continue to be used.

  This change enforces that geolocation can only provide a time zone suggestion once in the first run wizard.

### DIFF
--- a/lib/python/Components/Timezones.py
+++ b/lib/python/Components/Timezones.py
@@ -52,50 +52,49 @@ TIMEZONE_FILE = "/etc/timezone.xml"  # This should be SCOPE_TIMEZONES_FILE!  Thi
 TIMEZONE_DATA = "/usr/share/zoneinfo/"  # This should be SCOPE_TIMEZONES_DATA!
 
 def InitTimeZones():
-	tz = geolocation.get("timezone", None)
-	proxy = geolocation.get("proxy", False)
-	if tz is None or proxy is True:
-		area = DEFAULT_AREA
-		zone = timezones.getTimezoneDefault(area=area)
-		if proxy:
-			msg = " - proxy in use"
-		else:
-			msg = ""
-		print "[Timezones] Geolocation not available%s!  (area='%s', zone='%s')" % (msg, area, zone)
-	elif DEFAULT_AREA == "Classic":
-		area = "Classic"
-		zone = tz
-		print "[Timezones] Classic mode with geolocation tz='%s'.  (area='%s', zone='%s')" % (tz, area, zone)
-	else:
-		area, zone = tz.split("/", 1)
-		print "[Timezones] Modern mode with geolocation tz='%s'.  (area='%s', zone='%s')" % (tz, area, zone)
 	config.timezone = ConfigSubsection()
-	config.timezone.area = ConfigSelection(default=area, choices=timezones.getTimezoneAreaList())
-	config.timezone.val = ConfigSelection(default=timezones.getTimezoneDefault(), choices=timezones.getTimezoneList())
-	if not config.timezone.area.value and config.timezone.val.value.find("/") == -1:
-		config.timezone.area.value = "Generic"
-	try:
-		tzLink = path.realpath("/etc/localtime")[20:]
-		msgs = []
-		if config.timezone.area.value == "Classic":
-			if config.timezone.val.value != tzLink:
-				msgs.append("time zone '%s' != '%s'" % (config.timezone.val.value, tzLink))
+	config.timezone.area = ConfigSelection(default=DEFAULT_AREA, choices=timezones.getTimezoneAreaList())
+	config.timezone.val = ConfigSelection(default=DEFAULT_ZONE, choices=timezones.getTimezoneList())
+	if config.misc.firstrun.value:
+		proxy = geolocation.get("proxy", False)
+		tz = geolocation.get("timezone", None)
+		if proxy is True or tz is None:
+			msg = " - proxy in use" if proxy else ""
+			print "[Timezones] Warning: Geolocation not available%s!  (area='%s', zone='%s')" % (msg, config.timezone.area.value, config.timezone.val.value)
 		else:
-			tzSplit = tzLink.find("/")
-			if tzSplit == -1:
-				tzArea = "Generic"
-				tzVal = tzLink
+			area, zone = tz.split("/", 1)
+			if area != DEFAULT_AREA:
+				config.timezone.area.value = area
+				choices = timezones.getTimezoneList(area=area)
+				config.timezone.val.setChoices(choices, default=timezones.getTimezoneDefault(area, choices))
+			config.timezone.val.value = zone
+			config.timezone.save()
+			print "[Timezones] Initial time zone set by geolocation tz='%s'.  (area='%s', zone='%s')" % (tz, area, zone)
+	else:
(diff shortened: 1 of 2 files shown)


WanWizard:
Too bad this is merged, as I don't see the point.

The only reason for wanting to enable this, as I see it, is when you travel the world with your box and you want the correct timezone everywhere you go. All other users need the correct timezone when they setup their box, and after than never again. And that is if they don't restore a backup, in which case that one-time isn't even needed.

So I maintain: this config key should go, and the geo code should go into the startup wizard, where the one shot is actually needed.

IanSav:
If you would like me to explain this code and what is going on please contact me directly.

WanWizard:
I don't need explanation, I think my point is clear.

IanSav:
No it is not.  There are a number of factors for which I don't think you have made allowances.

WanWizard:
That is why discussing it up front is usually better than arguing after the facts.

IanSav:
> That is why discussing it up front is usually better than arguing after the facts.

As you well know getting agreement for common code across all images is virtually impossible.  It was discussed with a few developers beforehand.  The code also existed for a long time before an agitator started lobbying for attacks against the code.

Anyway, it seems that my efforts are no longer required so I will be taking a long step back from Enigma2 development and repairs.

WanWizard:
> Anyway, it seems that my efforts are no longer required so I will be taking a long step back from Enigma2 development and repairs.

Too bad you feel that way. 

The fact an initial idea doesn't get consensus is a fact of life, and it may sometimes be difficult to find that consensus. It is something you have to live with in a team, and even more so if you try to work across teams. 

I don't care much for political or personal battles, for me it is about the use cases. And if the issues can be addressed without negative effects to those use cases. Which in this particular case I think is possible.

IanSav:
The fact is that there is nothing wrong with the code.  There was no privacy breach or any other *real* issue with the code or the data.  This was a witch hunt scare campaign launched by one disgruntled user who lobbied for others to back his cause.  I trust you noticed that the complaint post on GitHub was from an account created on the day and only used to post complaints about me on multiple images?  The user failed to respond to any requests for more information.

I have taken my development stance as administrators, like yourself, have supported code change based on rumor and innuendo as proof of their case and bullying as a tactic to achieve change.  Every post and objections has been politely handled by me with a simple request to explain the problem.  To date most users can't answer why they object and the few remaining can't provide *any* proof of information or privacy compromise.  This code has been in place for months with very few objectors across all of the Enigma2 images.  Now that the masses have been motivated with claims of privacy violation the issue has become significant and I have been made to be enemy number one.  There is still there is no evidence or proof to sustain the claim.  I have provided many responses to explain the code, explain the privacy consideration upon which I insisted.  I have also have developers from the geolocation organisation confirm all the privacy and protection measures that are in place.  Still facts and evidence apparently don't matter.

The other significant issue was that of the code running on every boot.  I explained why this was done and, after discussions with fellow developers I was made aware of a way to only run the code once during the initial installation.  This solution was coded within minutes of my being made aware of the possibility and the code was distributed to all images as soon as it was tested and verified.  Still that is not good enough.  Your comment "Too bad this is merged, as I don't see the point." shows me that I have lost your trust.

Now with my personal integrity called into question and my attempts to synchronise code between images to make development and merges easier being trashed I don't see a way forward to contribute.  The comments you made indicate to me you also question my efforts.  If that is how you feel then why should I continue to stress myself working up to 20 hours a day to improve Enigma2 and bring the images closer?

My decision is not final but I am very much not in a mind to spend considerable effort when it is not wanted or appreciated.

WanWizard:
>  Your comment "Too bad this is merged, as I don't see the point." shows me that I have lost your trust.

It is always very bad to make assumptions... :wink: 

I have no issue with querying an external service for this information. I would suggest to only query for the information needed, and not more. The same way you would code SQL queries. I have never said I see a privacy issue.

I don't see the point of having it query everytime the box boots, once the timezone is set, it never changes. 

I don't see the point of having that config value. Adding this complexity, and yet another item to the configuration, only for that one user that carries his box with him on world travels and doesn't want to manually change the timezone.

I don't see the point to have the query done at this location in the code. It should be in the setup wizard, which is the only place the user is asked to set a timezone. This will also mean users restoring a backup after flash will never see the question, and the query is never run.

I see this as constructive critisism, and nothing personal. I was no part of any other discussion about this topic before this, nor am I sensitive to privacy zealots (by creating an account that person also gave away his data).

IanSav:
This isn't the place for all of this but it appears you don't use the other chat tools any more.

The problem is that the welcome wizard only runs after all of Enigma2 has started.  The time zone and clock settings happen at the very start of the boot.  By the time the wizard runs all the important time zone decisions have been made and enforced.  To change them in the wizard is not possible because the setting components to make such changes don't exist.  Further such a setting change risks overriding a choice made by the user.  The old code tried to sort this out but, as some people noted, had to err on the side of caution to not use geolocation data if the source of the existing data could not be verified.  That meant that some users considered the code broken because I could not identify the source of the time zone so I refused to change it leaving the incorrect default in operation.  This was promoted to users as I wrote broken code.

The change to which you negatively reacted was to ensure that the geolocation code now only takes effect when running the welcome wizard which is the only time I can ensure that a previous time zone choice has not been made.  Once I get a valid geolocation look up the code locks out the function to never run again.  That switch mechanism can later be adjusted when there is code and facility to better handle geolocation access permission from the user and when the time setting code can correctly process an option to source the data via geolocation.  As I said *none* of that infrastructure currently exists.

IanSav:
By the way, if you have a solution that allows what you want to happen to be done and it doesn't require major surgery to Enigma2 then I welcome any suggestions.  (I already have the major surgery solution but it is way to big and too much to be done quickly.)

oldsat:
> by creating an account that person also gave away his data

Very wrong.
I just created an account here... I give you some hints. Public internet space, so not really my PC or IP, disposable email, plus could also use Tor network... any geolocation information or other personal data microsoft have about me is wrong.
Most people that use this services to report issues/bugs do the same.

> a way to only run the code once during the initial installation. This solution was coded within minutes 

@IanSav 
I read countless discussions across different images about this E2 geolocation, to the date of this PR and like 3 months before some people asked you like hundrends of times to not run this at every boot, you always tried to shake off the request by lecturing people on how internet works, sometimes almost saying it was impossible to do, or that it was not open to discussion, here you say it took "minutes" to code. Well...
It seems to me that you actually wanted that cr*ap to run at every boot despite it was unwanted and not needed. Why wanting to keep that behavior at all costs, remains the question.

PR #2581, opened 2020-05-25
Include CI assignment - support two modules

Dima73:
-[CIHelper] support more modules,optimized check for assigned service
-[Alternatives] added function to search for a free tuner for
alternative services with a setting "Alternative services tuner
priority"
-[RecordTimer] automatic search for a recording tuner when using CI
modules
-[Navigation] update play service for new  'Alternatives' function
-[TimerSanityCheck] update for new 'CIHelper' tools

Attention!
Checking multi-decoding modules is now disabled.At the moment, it’s only
"AlphaCrypt".
If this is still necessary, then several lines must be uncommented.


littlesat:
Why don’t you make the alphacrypt stuff detectable and/or configurable

PR #2614, opened 2020-06-18
About - box uptime to description

ims21:


littlesat:
Nice idea... but why is this info not added in the about text instead... why at description?

ims21:
Because I think, this info is more about working as about box info - I did not want add it to scrolllabel, where is a lot of informations yet.
And under PIG is so much unused space :-)
My first variant was add it as "About" + "Uptime: .... to Title, but it was ugly. This is more decent...

I have prepared skins ... :-)

68foxboris:
I add this line AboutText += _("Uptime: ") + about.getBoxUptime() + "\n"
1_0_16_2C1_7_70_1680000_0_0_0_20200618185530

ims21:
Yes, but when are more HDDs, then this next info adding scrollbars and I did not want it - scroll in info ... STUPID...
Under PIG is cca 1/4 free whole screen ...

persianpros:
This is how I did it 2 months ago: https://github.com/OpenVisionE2/enigma2-openvision/commit/7fdb7c95b206fce1a715d737b06dd9b773f36547

ims21:
obrazek

Decent imho ...  but I want not spend full 7 pages in discussion about one screen ;-)

WanWizard:
I think the entire screen should be redesigned, because this looks absolutely weird, that one item on the left. 

I may be a better idea to move the entire Enigma section to the left, and keep the right side for everything hardware / OS related. It will also prevent a scrollbar when you have more tuners, storage devices and network interfaces (an ET10000 with wifi has 3).

persianpros:
In OV it's different, we have more menu sections in About: https://github.com/OpenVisionE2/enigma2-openvision/blob/develop/data/menu.xml#L25~L36

ims21:
> 
> 
> I think the entire screen should be redesigned, because this looks absolutely weird, that one item on the left.
> 
> I may be a better idea to move the entire Enigma section to the left, and keep the right side for everything hardware / OS related. It will also prevent a scrollbar when you have more tuners, storage devices and network interfaces (an ET10000 with wifi has 3).

Agree ... I have more HDDs and why must I then use scrollbar, when I want see IP ? Due It I moved it left, where is free space. And it is not true 'box info', imho

ims21:
I don't care - added it to scrollbar

PR #2620, opened 2020-06-20
[Screen.py] Fix a schoolboy coding error

Huevos:
Did someone really think it was a good idea to try to re-assign "False"?

How did this get past the infallible quality control radar?

Original commit: OpenPLi/enigma2@bb28004#diff-685ef552d889a7ab8ab8df5a1bb08bacR12


WanWizard:
It is a 2009 commit, so that radar would have been Dream Multimedia?

PR #2623, opened 2020-06-22
Improve the menu path code

IanSav:
[GUISkin.py] PEP8 clean up

[GUISkin.py] Tidy up the code
- Sort and move imports to the top of the file.
- Add proposal to address the baseRes FIXME comment.
- camelCase variables.
- Tidy up the comments.
- Tidy up the debug log prints.

[GUISkin.py] Clean up old dead code
- The getSkinTitle() method appears to be unused.
- The "skin_title" variable is simply a duplicate of self["Title"].
- The skin "title" attribute is already translated by the skin load process.

[GUISkin.py] Optimise menu path code
- Add a new skin widget "ScreenPath" that can be used and shared by skins for multiple images.  This widget follows the style of its sibling widget "Title".
- The previous widget is still supported.  It should be removed when all skins are updated.
- Eliminate the "screenPathMode" variable and its setScreenPathMode() method.  This can more simply managed with an easier to follow argument "addToPathList=True" that has been added to the setTitle() method.  Any titles should not be added to the history should add the argument "addToPathList=False" to the setTitle() method call.
  This is easier to follow in the code rather than having setScreenPathMode() method calls somewhere else in the code.
- The  setScreenPathMode() method is still defined but does nothing and should be removed when calls to it are removed from the rest of the code.

[GUISkin.py] Optimise code
- Add code optimisation as suggested by Prl.

NOTE: This code should not be merged until reviewed by Littlesat and some other code adjustments are made elsewhere in OpenPLi (as described above).

UPDATE: From my scans it appears that  setScreenPathMode() should be removed from Menu.py, MessageBox.py, ChannelSelection.py and InputBox.py.  The "addToPathList=False" argument should be added to any/all setTitle() method calls in MessageBox.py, ChannelSelection.py and InputBox.py as required.  In practice these are mostly end-point screens so they should not need protection from being added to the menu list as they don't spawn other screens.

### DIFF
--- a/lib/python/Components/GUISkin.py
+++ b/lib/python/Components/GUISkin.py
@@ -22,7 +22,6 @@ class GUISkin:
 	def __init__(self):
 		self["Title"] = StaticText()
 		self["screen_path"] = StaticText()
-		self.skin_title = ""
 		self.onLayoutFinish = []
 		self.summaries = CList()
 		self.instance = None
@@ -92,7 +91,6 @@ class GUISkin:
 
 	def setTitle(self, title):
 		pathText = ""
-		self.skin_title = title
 		if self.screenPathMode is not None and title and config.usage.menu_path.value != "off":
 			if self.screenPathMode and not screenPath.pathList or screenPath.pathList and screenPath.pathList[-1] != title:
 				self.onClose.append(self.removeScreenPath)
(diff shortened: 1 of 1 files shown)


IanSav:
This pull request will be replaced with a significantly improved version.

PR #2666, opened 2020-07-13
[Screen.py] Protect against screens with no self or session

IanSav:
This change protects the code from tasks started at system startup that don't have a self or session.


IanSav:
There is an error in the pull request.  It will be replaced with a corrected version.

PR #2672, opened 2020-07-19
[Screen] dont use menu path for ChannelSelection

Dima73:
-fix title translation if use title from skin attributes

### DIFF
--- a/lib/python/Screens/Screen.py
+++ b/lib/python/Screens/Screen.py
@@ -154,10 +154,11 @@ class Screen(dict):
 		except AttributeError:
 			pass
 		self.screenTitle = title
-		if config.usage.menu_path.value == "large":
+		dont_use_menu_path = "ChannelSelection" in self.__class__.__name__
+		if config.usage.menu_path.value == "large" and not dont_use_menu_path:
 			screenPath = ""
 			screenTitle = "%s > %s" % (self.screenPath, title) if self.screenPath else title
-		elif config.usage.menu_path.value == "small":
+		elif config.usage.menu_path.value == "small" and not dont_use_menu_path:
 			screenPath = "%s >" % self.screenPath if self.screenPath else ""
 			screenTitle = title
 		else:
@@ -228,8 +229,8 @@ class Screen(dict):
 				if title:
 					self.skinAttributes[skinTitleIndex] = ("title", title)
 				else:
-					self["Title"].text = value
-					self.summaries.setTitle(value)
+					self["Title"].text = _(value)
+					self.summaries.setTitle(_(value))
 			elif key == "baseResolution":
 				baseRes = tuple([int(x) for x in value.split(",")])
 			idx += 1


IanSav:
Why was this code changed?
1. It is not good practice to translate variables as the original strings are *not* harvested for translation.
2. Please explain why the code is not working correctly such that it needed to be patched.  I suggest that the cause of the problem be addressed rather than bloating this new code.

If you don't want titles in ChannelSelection then just remove the call to setTitle() there!

Dima73:
1)Do you always decide for everyone?

For your information, there are alternative translations of the image, it includes the translation of some plugins.
And such translations are very popular on local forums, for example Russian.

https://github.com/OpenPLi/enigma2/commit/32ec6853aad1c8bbb9568fa77b07c2dc06a3f7b6#diff-5c23dd9a148dfc142501202a9111a03f

2)https://forums.openpli.org/topic/77618-manual-satellite-scan-path-remains-forever-on-top-of-the-screen/

IanSav:
I don't decide anything for anyone.  I contribute code and them am happy to discuss and workshop any change.  Did you consult me about any perceived issues with the code?  Don't you think that might have been appropriate?

Regarding point 1, from skin.py:
`
def title(self, value):
    self.guiObject.setTitle(_(value))
`
The text is already translated.  This now creates a translation of a translation!  That is wrong.

As for point 2, I will look at the thread in more detail shortly to investigate the issue.  The changes made to Screen.py are just a hack to potentially cover up the real issue.

EDIT:  I can't make the code fragment indent but I am sure you can see what I am showing you.

Taapat:
> Regarding point 1, from skin.py:
> `def title(self, value): self.guiObject.setTitle(_(value))`
> The text is already translated. This now creates a translation of a translation! That is wrong.

And where the title() is used here?
Here is used setTitle() where the title is not translated.
UPD.
For text() of course you don't need a translation.

IanSav:
If code want's to set a title viat the self.title mechanism then the title fed into the applySkin() method should already be translated.

The translation code can *NOT* find text to add to the .po files via the variables that convey the text.  That is

`     text =  _("Translate me")`

will be added to the .po file but

`    text = "Translate me"
    _(text)`

will not be harvested.  It is better to translate text directly.

IanSav:
On a related note, it is better to use the setTitle() method rather than sharing titles via a shared variable.

Taapat:
Looks like you don't understand how translations are parsed in enigma.
Translations are taken not only from _() but also from xml files: https://github.com/OpenPLi/enigma2/blob/develop/po/xml2po.py
If you want to translate text from an xml file you need to use _ (text)

IanSav:
How is what I said in any way incorrect?  What I said was that translated variables for variables defined within the code are not collected for translation.  I even gave an example of exactly what I meant.  I never said that selected attributes weren't collected directly from the XML files.

So, based on your comment, why are we translating text obtained from XML that has already been translated?

I suspect that the difficulty here is that we are talking about generic comments.  This conversation would be far more meaningful if we could talk about specific instances in the code.

Taapat:
You clearly state:
> The translation code can _NOT_ find text to add to the .po files via the variables that convey the text

But now you write that you haven't said that the translation can't be added in .po files in any other way. Very interesting.


Why do you assume that text obtained from xml has already been translated?
The fact that it is added in po does not mean that it is translated.
In applySkin() you take the title from untranslated skinAttributes that come from xml files and use in setTitle().
At what point title is translated here?

IanSav:
Yes, I wrote that and I stand by that.  I even gave an example.

I didn't say this:

>  But now you write that you haven't said that the translation can't be added in .po files in any other way. Very interesting.

In fact I have no idea what this is meant to mean.

> Why do you assume that text obtained from xml has already been translated?
> The fact that it is added in po does not mean that it is translated.
> In applySkin() you take the title from untranslated skinAttributes that come from xml files and use in setTitle().
> At what point title is translated here?

Because the "title" attributes in the skin are translated on skin load.  I quoted the code from skin.py that does it.

IanSav:
If you think there is an error in my logic, or an error in the skin load code then please show me the problem or evidence of the problem.  I am happy to fix errors but I want to fix them properly and not via ill conceived hacks.

Taapat:
So setTitle(_ (value)) in title() is not a hack but in applySkin() is. Interesting logic.


Here self.summaries.setTitle() is used if screen does not use setTitle() but title is specified in the skin.
You haven't answered me at what point in this case "title" attribute in the self.skinAttributes are translated on skin load?
It seems to me that here skinAttributes are obtained using collectAttributes() without translation and after is used in self.summaries.setTitle().
Without translation here, an untranslated title appears on the receiver's display.

IanSav:
I know you just want to argue and find fault in my logic but I will try to continue responding politely.

Doing the translation once at the first opportunity after the skin or menu or setup file is loaded is the appropriate place to do the translation.  These are the first places where the data is created.

If you would like to provide me with exact steps to demonstrate an issue I will look at it.

As for the hack to "fix" ChannelSelection, I have found the real problem and will submit a fix in a few minutes.  The error was in Screen.py and it was my mistake but the hack completely misses the bug and, in fact, hides it.

Taapat:
You as always continue to ignore what someone writes to you because it does not match your assumptions.
As usual, a lot more comments will be needed here, so at the end you probably read what is written in the first comments.
I have already clearly indicated the conditions under which an untranslated title from skin is used here to display the title in SummaryScreen on the receiver display.

Taapat:
Of course, you can make a separate exception for the title in collectAttributes() and do the translation there.
But in that case also the class AttributeParser to what you pointed is not the right place to do the translation because it also uses existing data from skinAttributes.

PR #2673, opened 2020-07-20
[Screen.py] Revert hack and apply appropriate correction

IanSav:
- Revert the previous hack commit as it doesn't fix the problem.  It only hides it.
- Apply the appropriate correction to the screen path history buffer not being cleared when the screen path stack is cleared.

### DIFF
--- a/lib/python/Screens/Screen.py
+++ b/lib/python/Screens/Screen.py
@@ -148,17 +148,18 @@ class Screen(dict):
 		try:  # This protects against calls to setTitle() before being fully initialised like self.session is accessed *before* being defined.
 			if self.session and len(self.session.dialog_stack) > 2:
 				self.screenPath = " > ".join(ds[0].getTitle() for ds in self.session.dialog_stack[2:])
+			else:
+				self.screenPath = ""
 			if self.instance:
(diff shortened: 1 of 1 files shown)


Taapat:
Why are you in this commit revert title translation in SummaryScreen on the receiver display?
This does not affect the ChannelSelection problem.

IanSav:
It should not be under a commit claiming to fix ChannelSelection title issues.

I am yet to see any proof that the translation is required here.  If you can demonstrate an issue with translation with reproducible steps then I will investigate that issue as well and apply appropriate corrections.

Taapat:
The commit what you cancel was not only about ChannelSelection.
Why in the fix for ChannelSelection you also revert the translation?
I have already indicated the circumstances in which the text is not translated.
How many times I need to write this?

IanSav:
Writing it once with clear steps on how to reproduce the problem is all I require.  Saying the same thing repeatedly without the reproduction steps is not going to help.

The bug report on the ChannelSelection issue gave *exact* steps on how to trigger the error and a screen capture showing the exact problem.  When you can offer me something similar then I will be able to investigate the issue.

PR #2678, opened 2020-07-24
Translate title and text when collect attributes from skin

Taapat:
Title is used not only in AttributeParser but also in applySkin() to use for summaries.
This will fix the translation when the title is obtained from the skin instead of using setTitle().
It will also fix double translation in AttributeParser when you're using already translated text from setTitle() in applySkin() add it to the skin attributes and translate again in AttributeParser.

### DIFF
--- a/skin.py
+++ b/skin.py
@@ -370,6 +370,8 @@ def collectAttributes(skinAttributes, node, context, skinPath=None, ignore=(), f
 			elif attrib == "font":
 				font = value.encode("utf-8")
 				skinAttributes.append((attrib, font))
+			elif attrib in ("title", "text"):
+				skinAttributes.append((attrib, _(value.encode("utf-8"))))
 			else:
 				skinAttributes.append((attrib, value.encode("utf-8")))
 	if pos is not None:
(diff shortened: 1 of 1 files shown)


IanSav:
This pull request should *NOT* be merged until the exact problem it is solving is documented with steps required to reproduce the issue are given.

I have tried to get this information on the forum with no success.

Taapat:
How can you suggest any changes if you can't understand what's going on in some lines of code in a try this?
I've pointed this to you several times but you still ignored it.

IanSav:
If you think I don't understand the problem did you ever consider that you aren't explaining it well?  How about document the issue clearly so that *everyone* can understand the issue.

It is interesting how I worked day and night to try and understand this issue and all you can say is I am ignoring it.  If you worked as hard to explain the problem as you do to criticise me then we would probably be much further down the track to addressing whatever it is that you think is broken.

littlesat:
Indeed why is it needed? Eg how could we verify the need?

Taapat:
https://forums.openpli.org/topic/77618-manual-satellite-scan-path-remains-forever-on-top-of-the-screen/page-3#entry1235378


https://github.com/OpenPLi/enigma2/pull/2672

Taapat:
On skins that do not use Title in windowstyle (such as OpenPli skins) Title will not be translated also on screen in TV.

IanSav:
I have already offered a better solution, offered in the forum topic, and I am waiting for instructions on how the correction should be submitted.

Taapat:
Where is your solution in the forum for PLi-HD skins?

IanSav:
I only have a sample of the correction required for the GigaBlue UE 4K front panel skin.

My presumption was that once the code was corrected and available to developers and users that the skins could follow and be fixed by the skin maintainers.

littlesat:
Currently I have no time to fix the standard pli skins....

littlesat:
Can we close this one -or- merge?

IanSav:
In my opinion this should be closed.

PR #2698, opened 2020-08-07
[keymap.xml] Organise and document the keymap file

IanSav:
- Start moving map blocks into more logical groupings.
- Start documenting the keymaps.
- Use a space before the self close tag for readability.

I hope that this tidy up will assist in comparing keymaps between images.  This is only a start of what will probably be a long term clean up process.


IanSav:
There are other errors that should be corrected and provided in a single clean commit.

PR #2701, opened 2020-08-09
GigaBlue PiP

AbuBaniaz:
Translate remote code so PiP works on GigaBlue remote.
Key translation while we await  a remote that sends the correct code.


AbuBaniaz:
See this thread
https://forums.openpli.org/topic/77898-gigablue-trio-4k-pip-key-is-not-mapped/

littlesat:
This can only be merge as temporary solution... actually we should get rid of such if gigablue exception....

PR #2705, opened 2020-08-13
Offline decoding - select path for decoded file

ims21:
- Timer recording path (as original behaviour)
- Current directory
- Select path...


littlesat:
What added value does this have... I think we should keep it in the same directory and when a file is fully encoded then remove the coded one....
adding more not required features makes it too complicated.

ims21:
I do not know, what is complicated. Existing code is divided to 2 parts, where in first part is obtained destination folder (in checkbox) used then in 2nd part. Added are things around checkbox only.

PR #2707, opened 2020-08-14
Support advanced sat position depends in use rotor

Dima73:
-for  additional cable of motorized LNB witch twin/quattro LNB using
advanced satellite parameters.


littlesat:
Can you explain the goal... I see also it is a big impact in code..: and bools and exceptions and old_ variables....

Dima73:
more info:
https://forums.openpli.org/topic/72892-unicable-lnb-with-usals-trouble-recording-while-watching/?hl=unicable

This is actually a very interesting option.
For example, I use a twin LNB on the motor.
Set one tuner as "Second cable of motorized LNB" too wasteful.
But it is quite possible to use it in conjunction with DiSEqC.
You will always have fallback tuner to record when needed.
For two weeks I checked this option, when everything was already written, there were no errors when using the simple twin LNB.
Let's just check it out,especially when using Unicable LNB, you can always undo/revert it.

PR #2708, opened 2020-08-17
Update unicable.xml

kueken:


kueken:
Maybe you like it that way?

WanWizard:
Think in your original commit I saw Jultec's with 2 UB's, which means "JPSxx/JRSxx 2 UB" is missing?

kueken:
Yes, but in the forum the request was made not to change Jultec. So I made it back.
I then took care of the syntax error at Kathrein.
greetings

persianpros:
You could use our updated version based on PLi: https://github.com/OpenVisionE2/enigma2-openvision/blob/develop/data/unicable.xml

Simply replace it.

WanWizard:
Also missing the "JPSxx/JRSxx 2 UB" :grin:

persianpros:
@WanWizard 

Where is "JPSxx/JRSxx 2 UB" that you said?

WanWizard:
It was in the original PR.

PR #2709, opened 2020-08-17
Add option "Auto play last service in radio mode"

Dima73:
-Only option "Hide channel list in radio mode" set disable.


littlesat:
Sorry I also do not understand the goal for this one....’

Dima73:
more info:
https://forums.openpli.org/topic/78198-tv-radio-mode-switch/
Indeed, when using a motor, this is inconvenient.
And by the way, I also asked myself this question, if the list of radio channels is immediately visible, then give me the opportunity to choose a service for zap(play)  myself.

PR #2718, opened 2020-08-23
[Satconfig] show entry "Externally powered" when this not FBC tuner

Dima73:
SystemInfo["LnbPowerAlwaysOn"] = HardwareInfo().get_device_model() in
("vusolo4k", "vuduo4k", "vuultimo4k", "vuuno4k", "vuuno4kse")

I checked my solo4k.
In standby:
FBC tuner(A, B ) voltage 13.68v is tuner not use.
Dual DVB-S2 tuner(I,J) voltage 0v is tuner not use.


littlesat:
Even fbc tuners can be externally powered...
That vu is doing it not correctly is a different story

Dima73:
> That vu is doing it not correctly is a different story

And what to do?
At the moment, SystemInfo["LnbPowerAlwaysOn"] only has VU+ models.

littlesat:
Sorry now I got it....

PR #2721, opened 2020-08-29
Improve recent skin/screen commits

IanSav:
[skin.py] Improve recent commit
- Improve the mandatory widget code.  Find all widgets in the screen including all widgets in panels before deciding if the screen is acceptable.
- Improve the templated embedded skin processing by using the existing "getSkinFactor()" method.
- Improve the text of the messages in the code above.

[Screen.py] Improve mandatoryWidgets default
- Using an empty list in code default calling parameters can have unexpected results.

### DIFF
--- a/lib/python/Screens/Screen.py
+++ b/lib/python/Screens/Screen.py
@@ -16,7 +16,7 @@ class Screen(dict):
 	ALLOW_SUSPEND = NO_SUSPEND
 	globalScreen = None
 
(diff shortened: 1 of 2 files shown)


littlesat:
Merge because it is a clear improvement...
But we need further discuss the 'version' idea... Which has no 'for' loops is cleaner and quicker... only skinners need to adapt as they do not want to show the embeded skin.

PR #2723, opened 2020-08-30
Clean up screen path code

IanSav:
[Screen.py] Remove defunct "screen_path"
- All official OpenPLi skins have now been updated to use the newer "ScreenPath" widget so the defunct "screen_path" widget can now be removed.

[UsageConfig.py] Transition screen path settings
- This temporary code will find the user's current value for the old "config.usage.menu_path" setting and move it to the new "config.usage.showScreenPath".  The code will then remove the old "config.usage.menu_path" entry from the settings file.

- This code is only run once such that when the setting is transferred the old setting is no longer used.  It is intended to remove this code after one or two production releases of OpenPLi when it will be assumed that most users have had the setting transferred.  When this code is removed the only implication for users will be that the setting will not be migrated and the "config.usage.showScreenPath" default value will be used.  Also the old "config.usage.menu_path" setting, if not at default, will not be removed.

[setup.xml] Use the new showScreenPath setting
- This change uses the new "config.usage.showScreenPath" setting and improves the label and description text to better reflect the functionality.

[Screen.py] Use new "config.usage.showScreenPath"
- This change replaces the old "config.usage.menu_path" setting with the new and proposed common "config.usage.showScreenPath" setting.

[UsageConfig.py] Remove transition code
- This change was requested by Littlesat and Persian Prince as OpenPLi production releases are too far apart such that this code could hang around for a very long time.  It is not considered an issue for this setting to be reset to default and the old value to be left in the setting file.

### DIFF
--- a/data/setup.xml
+++ b/data/setup.xml
@@ -9,7 +9,7 @@
 	<setup key="usage" title="Customize">
 		<item level="1" text="Sort order for menu entries" description="This option allows you to hide and sort menu entries. When selecting user you can change order and hide menu items via the blue button. With the user defined hidden the blue button is not shown in the menus">config.usage.menu_sort_mode</item>
 		<item level="1" text="Sort order for setup entries" description="This option allows you to alphabetically sort setup menu entries.">config.usage.sort_settings</item>
-		<item level="1" text="Show menu paths" description="This option allows you to show menu paths.">config.usage.menu_path</item>
+		<item level="1" text="Show screen path" description="This option allows you to show the full screen path leading to the current screen.">config.usage.showScreenPath</item>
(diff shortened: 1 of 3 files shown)


littlesat:
I suggest we remove the temporary code.... We can put it at the release notes that user need to reconfigure this setting... I just suggest to change the default to small instead.

IanSav:
I would prefer that we try to rescue the settings for users.  It will make my change less disruptive.  I will ensure this code is removed after update after the next update.  This code will *not* last long.  It only ever runs once and will clean up the user's setting file.  Without this code the settings file may retain a stale entry.

persianpros:
develop branch won't be used for normal releases so users might see this for example 1 month later, then we need to wait till then so it's better to not have it :+1:

IanSav:
So how can we clean up the old setting value?

IanSav:
Pull request updated as requested.

PR #2729, opened 2020-08-31
[config.py] Allow non string values in ConfigSelection

Huevos:
This commit allows non-string values to be used in ConfigSelections, e.g. booleans and integers. These values can now be safely saved to the settings file. On reading the settings file the correct data type is retrieved.

This code ensures the correct functioning of the many instances of ConfigSelection that are already using non-string values, and offers greater flexibility to developers writing new code.

Commit author: PRL

### DIFF
--- a/lib/python/Components/config.py
+++ b/lib/python/Components/config.py
@@ -216,7 +216,7 @@ class choicesList(object):  # XXX: we might want a better name for this
 
 	def index(self, value):
 		try:
-			return self.__list__().index(value)
+			return list(map(str, self.__list__())).index(str(value))
 		except (ValueError, IndexError):
 			# occurs e.g. when default is not in list
 			return 0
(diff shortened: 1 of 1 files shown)


littlesat:
Are you sure this is a good idea.....? As it has also limitations and could break thinks in case somewhere mixed types were used as everything is forced to a string and mapped.....

Huevos:
Please read the comment in the commit. ConfigSelection item must be unique. That has always been the case. Everything that worked before will continue working + we can now use integers and booleans.

Also there is no problem with mixed cases, unless someone wants to do something stupid like this, 
choices = ["False", False]

But that is an error with the coder, not the code.

littlesat:
It does something with index and when the choices are changed (is possible) then this does not work. I also do not see the added value strings were always mandatory here. There is no need at all... it does not solve anything as far I can see... choices should have strings...

Huevos:
If you look through the code base you will find examples of booleans and integers being stored in ConfigSelections. This code ensures those ConfigSelections are handled properly and correctly read and saved in the settings file.

The "string" only requirement is just because the code was not able to handle this, and for no other good reason.

Why would you not want code that adds sanity?

littlesat:
We still do not get the added value... We are are not convinced... it makes the code more complicated without any added value.

Huevos:
Ok, if you can't see the value don't merge it.

And all the code that contains integers and booleans in ConfigSelections won't work properly, including OpenPLi code.

Integers have been used in ConfigSelections since before OpenPLi existed.

Why do you want to limit variables to strings when there are other variable types?

Huevos:
By the way, OpenPLi is full of ConfigSelections that contain integers. Takes just seconds to find them.

https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Screens/ScanSetup.py#L1036-L1080

https://github.com/OpenPLi/enigma2-plugins/blob/master/blindscan/src/plugin.py#L549-L553

https://github.com/OpenPLi/enigma2-plugins/blob/master/blindscan/src/plugin.py#L583-L589

https://github.com/OpenPLi/enigma2-plugins/blob/master/blindscan/src/dmmBlindScan.py#L668-L675

https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Screens/ScanSetup.py#L1027-L1029

prl001:
> it makes the code more complicated without any added value.

The added complexity is quite small, and IMO the gain in flexibility is large in comparison: pretty much any type that can be converted to a string that's unique among the choices can be used in ConfigSelections. The largest part of the change is to add a comment detailing how it works.

prl001:
Here's one of my test examples that shows how the feature might be used on a slightly more structured data type:
`config.movielist.my_corners = ConfigSelection(default=(-1, -1), choices=[((-1, -1), "bl"), ((-1, 1), "tl"), ((1, 1), "tr"), ((1, -1), "br")])`

And another on a mixed-type ConfigSelection (coincidentally, taken from PLi's UsageConfig.py, though it has since been converted to only use strings):
`config.movielist.my_boolean_graphic = ConfigSelection(default=True, choices={False: _("no"), True: _("yes"), "only_bool": _("yes, but not in multi selections")})`

littlesat:
And what is the added value? What is the real need? What does not work here?
Plus when the choices changes the index does not work as it could get wrong...

Huevos:
Littlesat, here is a line from satfinder...

`self.feid = int(self.satfinder_scan_nims.value)`

Don't you think it is better to do that one place at source than copying that code all over the image.

Even worse in ScanSetup.py where the integer conversion is not done at all.

littlesat:
Yes... as a config 'API' for selection box is always a "string"...

When you try to improve that part then better work with...

self.feid = self.satfinder_scan_nims.getIntegerValue()
-or-
self.feid = self.satfinder_scan_nims.int_value

You also want to to avoid that self.feid might become a string.. This is guaranteed here and not anymore after this commit suggestion!... And it might cause crashes later... And when keeping it as is while you read the code you 'know' that it is an integer...

PR #2738, opened 2020-09-12
[NimManager] add option ignore delsys info

Dima73:
This will help you use the correct tuner type when reading only
/proc/bus/nim_sockets
For example, Sundtek applies patch in  file /usr/bin/enigma2.sh line
LIBS="/opt/lib/libmediaclient.so /usr/lib/libopen.so.0.0.0"
But this leads to completely incorrect information from API 5.10

### DIFF
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -515,6 +515,7 @@ class NIM(object):
 		self.i2c = i2c
 		self.frontend_id = frontend_id
 		self.__is_empty = is_empty
+		self.ignore_delsys_info = "Sundtek" in description
 
 		self.compatible = {
 				None: (None,),
@@ -530,7 +531,7 @@ class NIM(object):
 
 		# get multi type using delsys information
 		self.combined = False
-		if self.frontend_id is not None:
+		if self.frontend_id is not None and not self.ignore_delsys_info:
 			types = [type for type in nim_types if eDVBResourceManager.getInstance().frontendIsCompatible(self.frontend_id, type)]
 			if "DVB-T2" in types:
 				# DVB-T2 implies DVB-T support


Dima73:
`root@vusolo4k:~# cat /proc/bus/nim_sockets
NIM Socket 0:
        Type: DVB-S2
        Name: Vuplus DVB-S NIM(7376 FBC)
        Frontend_Device: 0
NIM Socket 1:
        Type: DVB-S2
        Name: Vuplus DVB-S NIM(7376 FBC)
        Frontend_Device: 1
NIM Socket 2:
        Type: DVB-S2
        Name: Vuplus DVB-S NIM(7376 FBC)
        Frontend_Device: 2
NIM Socket 3:
        Type: DVB-S2
        Name: Vuplus DVB-S NIM(7376 FBC)
        Frontend_Device: 3
NIM Socket 4:
        Type: DVB-S2
        Name: Vuplus DVB-S NIM(7376 FBC)
        Frontend_Device: 4
NIM Socket 5:
        Type: DVB-S2
        Name: Vuplus DVB-S NIM(7376 FBC)
        Frontend_Device: 5
NIM Socket 6:
        Type: DVB-S2
        Name: Vuplus DVB-S NIM(7376 FBC)
        Frontend_Device: 6
NIM Socket 7:
        Type: DVB-S2
        Name: Vuplus DVB-S NIM(7376 FBC)
        Frontend_Device: 7
NIM Socket 8:
        Type: DVB-S2
        Name: Vuplus DVB-S NIM(AVL6222)
        Frontend_Device: 8
        I2C_Device: 4
NIM Socket 9:
        Type: DVB-S2
        Name: Vuplus DVB-S NIM(AVL6222)
        Frontend_Device: 9
        I2C_Device: 4

NIM Socket 10:
        Type: DVB-T
        Name: Sundtek DVB-T (III) (1/0) (USB):VTUNER
        Frontend_Device: 10
NIM Socket 11:
        Type: DVB-S2
        Name: Sundtek DVB-S/S2 (IV) (0/0) (USB):VTUNER
`

Dima73:
1new
1old
2old

littlesat:
Please move the if sundtek check as feature to SystemInfo.py....

Dima73:
This is nim description,don't know any other way.

WanWizard:
> Please move the if sundtek check as feature to SystemInfo.py....

This is a typical example of where the current SystemInfo design falls short: it only does static variables, not runtime information.

Dima73:
This patch is a simple and easy workaround.
I understand that this does not quite meet the requirements of openPli, but we are not an automatic machine and not a blind Themis of justice, so we can accept this option.
Sundek has been trying to fix the bug since 2016, but it didn't work out.

athoik:
Add an option on frontend? Ignore delsys default False.

People with that tuner will have to enable value to True.

Document this one on wiki.

littlesat:
The patch as is should be avoided...

littlesat:
Add an option to frontend might also not be a good idea.... maybe add to systeminfo a function that could retrieve the info could be considered. So instead of getting a static global value call a function...

PR #2740, opened 2020-09-12
EPGCache - lock on save

ims21:
it should prevent crash on epgcache saving


WanWizard:
I sure hope so, these crashes drive me nuts!

PR #2752, opened 2020-10-04
[listboxservice] set celServiceNextInfo priority higher than celServi…

Dima73:
…ceInfo

Because this is probably a feature(?) such that the default color of the
progress bar is taken from the color celServiceInfo.


Dima73:
It is necessary to carry out additional verification that this does not violate anything.

PR #2757, opened 2020-10-07
[listboxservice] force set default color progress bar as celServiceInfo

Dima73:
For next event.
Because this is probably a feature(?) such that the default color of the
progress bar is taken from the color celServiceInfo.

### DIFF
--- a/lib/service/listboxservice.cpp
+++ b/lib/service/listboxservice.cpp
@@ -707,6 +707,7 @@ void eListboxServiceContent::paint(gPainter &painter, eWindowStyle &style, const
 		bool serviceAvail = true;
 		bool serviceFallback = false;
 		int isplayable_value;
+		gRGB EventProgressbarColor = 0xe7b53f;
 
 		if (!marked && isPlayable && service_info && m_is_playable_ignore.valid())
 		{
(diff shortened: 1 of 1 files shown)


littlesat:
Why is this needed.... what issue might happen.... How do we trigger it?

Dima73:
` How do we trigger it?`
It's very simple, switch to the PLi-HD skin and look at the color of the progress bar, it will be yellow.
Now turn on two lines with the next event. The color of the progress bar will be gray, as well as the color of the next event.
This patch fixes this, the color of the progress bar, if it is not specifically indicated in the skin, will be the same as the color of the current event.

littlesat:
Ah... so actually it is a patch for A side effect of the two lines Channelselector with next epg...
You now first I thought it was a good idea that two line but at this moment I do not like it at all... unless I did start developing it...

Dima73:
` I did start developing it...`
I do not advise you to transfer painter to Python.
Firstly, it just greatly slows down the data processing speed, c++ against Python, this is heaven and earth.
Secondly, nobody needs it.
And I advise you not to listen to talkers, for whom the main thing is to scratch with tongue.

PR #2759, opened 2020-10-11
Auto detect region for DVB-T/T2 tuners

Dima73:
Attention!
This check will only work when zap to the service.
Availability in the channel list is not checked to reduce the load on
the system.

### DIFF
--- a/lib/dvb/dvb.cpp
+++ b/lib/dvb/dvb.cpp
@@ -840,7 +840,7 @@ RESULT eDVBResourceManager::allocateFrontend(ePtr<eDVBAllocatedFrontend> &fe, eP
 {
 	eSmartPtrList<eDVBRegisteredFrontend> &frontends = simulate ? m_simulate_frontend : m_frontend;
 	eDVBRegisteredFrontend *best, *fbc_fe, *best_fbc_fe;
-	int bestval, foundone, current_fbc_setid, c, system;
+	int bestval, foundone, current_fbc_setid, c, tuner_type;
 	bool check_fbc_leaf_linkable, is_configured_sat;
 	long link;
 
@@ -896,8 +896,8 @@ RESULT eDVBResourceManager::allocateFrontend(ePtr<eDVBAllocatedFrontend> &fe, eP
 			}
 			else
 			{
-				feparm->getSystem(system);
-				if (system == eDVBFrontend::feTerrestrial)
+				feparm->getSystem(tuner_type);
+				if (tuner_type == eDVBFrontend::feTerrestrial)
 					is_configured_sat = true;
 			}
 			c = i->m_frontend->isCompatibleWith(feparm, is_configured_sat);
--- a/lib/dvb/dvb.cpp
+++ b/lib/dvb/dvb.cpp
@@ -840,7 +840,7 @@ RESULT eDVBResourceManager::allocateFrontend(ePtr<eDVBAllocatedFrontend> &fe, eP
 {
 	eSmartPtrList<eDVBRegisteredFrontend> &frontends = simulate ? m_simulate_frontend : m_frontend;
 	eDVBRegisteredFrontend *best, *fbc_fe, *best_fbc_fe;
-	int bestval, foundone, current_fbc_setid, c;
+	int bestval, foundone, current_fbc_setid, c, system;
 	bool check_fbc_leaf_linkable, is_configured_sat;
 	long link;
 
@@ -893,6 +893,12 @@ RESULT eDVBResourceManager::allocateFrontend(ePtr<eDVBAllocatedFrontend> &fe, eP
 							is_configured_sat = true;
 					}
 				}
+			}
+			else
+			{
+				feparm->getSystem(system);
+				if (system == eDVBFrontend::feTerrestrial)
+					is_configured_sat = true;
 			}
 			c = i->m_frontend->isCompatibleWith(feparm, is_configured_sat);
 		}
--- a/lib/dvb/frontend.cpp
+++ b/lib/dvb/frontend.cpp
@@ -2871,6 +2871,22 @@ int eDVBFrontend::isCompatibleWith(ePtr<iDVBFrontendParameters> &feparm, bool is
 			/* prefer to use a T tuner, try to keep T2 free for T2 transponders */
 			score--;
 		}
+		/* auto detect region - check if this frequency is in the frontend list */
+		if (is_configured_sat)
+		{
+			char configStr[255];
+			snprintf(configStr, 255, "config.Nims.%d.smart_region_list", m_slotid);
+			std::string smart_region_list = eConfigManager::getConfigValue(configStr);
+			if (!smart_region_list.empty())
+			{
+				std::stringstream ss;
+				ss << parm.frequency;
+				std::string freq;
+				freq = ss.str();
+				if (smart_region_list.find(freq) == std::string::npos)
+					return 0;
+			}
+		}
 	}
 	else if (type == eDVBFrontend::feATSC)
 	{
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -4,7 +4,7 @@ from SystemInfo import SystemInfo
 from Tools.HardwareInfo import HardwareInfo
 from Tools.BoundFunction import boundFunction
 
-from config import config, ConfigSubsection, ConfigSelection, ConfigFloat, ConfigSatlist, ConfigYesNo, ConfigInteger, ConfigSubList, ConfigNothing, ConfigSubDict, ConfigOnOff, ConfigDateTime, ConfigText
+from config import config, ConfigSubsection, ConfigSelection, ConfigFloat, ConfigSatlist, ConfigYesNo, ConfigInteger, ConfigSubList, ConfigNothing, ConfigSubDict, ConfigOnOff, ConfigDateTime, ConfigText, NoSave
 
 from enigma import eDVBFrontendParametersSatellite, eDVBSatelliteEquipmentControl as secClass, eDVBSatelliteDiseqcParameters as diseqcParam, eDVBSatelliteSwitchParameters as switchParam, eDVBSatelliteRotorParameters as rotorParam, eDVBResourceManager, eDVBDB, eEnv
 
@@ -138,7 +138,8 @@ class SecConfigure:
 
 		nim_slots = self.NimManager.nim_slots
 
-		used_nim_slots = [ ]
+		used_nim_slots = []
+		terrestrial_nim_slots = []
 
 		for slot in nim_slots:
 			if slot.type is not None:
@@ -168,6 +169,10 @@ class SecConfigure:
 					if connto not in self.satposdepends:
 						self.satposdepends[connto] = []
 					self.satposdepends[connto].append(x)
+			elif slot.isCompatible("DVB-T"):
+				if nim.configMode.value != "nothing":
+					terrestrial_nim_slots.append(slot)
+				nim.smart_region_list.setValue("")
 
 		for slot in nim_slots:
 			x = slot.slot
@@ -245,6 +250,11 @@ class SecConfigure:
 								diseqc13V = nim.diseqc13V.value)
 					elif nim.configMode.value == "advanced": #advanced config
 						self.updateAdvanced(sec, x)
+			elif slot.isCompatible("DVB-T"):
+				if nim.configMode.value != "nothing" and len(terrestrial_nim_slots) > 1 and nim.smart_region.value:
+					region = self.NimManager.transpondersterrestrial[nim.terrestrial.value]
+					if region:
+						nim.smart_region_list.setValue(';'.join([str(x[1]) for x in region]))
 		print "[SecConfigure] sec config completed"
 
 	def updateAdvanced(self, sec, slotid):
@@ -1546,6 +1556,8 @@ def InitNimManager(nimmgr, update_slots = []):
 		list = [(x[0], x[0]) for x in nimmgr.terrestrialsList]
 		nim.terrestrial = ConfigSelection(choices = list)
 		nim.terrestrial_5V = ConfigOnOff()
+		nim.smart_region = ConfigYesNo(default = False)
+		nim.smart_region_list = NoSave(ConfigText())
 
 	def createATSCConfig(nim, x):
 		list = [(x[0], x[0]) for x in nimmgr.atscList]
--- a/lib/python/Screens/Satconfig.py
+++ b/lib/python/Screens/Satconfig.py
@@ -289,6 +289,15 @@ class NimSetup(Screen, ConfigListScreen, ServiceStopScreen):
 				self.terrestrialRegionsEntry = getConfigListEntry(self.indent % _("Region"), self.terrestrialRegions, _("Select your region. If not available change 'Country' to 'all' and select one of the default alternatives."))
 				self.list.append(self.terrestrialCountriesEntry)
 				self.list.append(self.terrestrialRegionsEntry)
+				nimcount = 0
+				for slot in nimmanager.nim_slots:
+					if slot.isCompatible("DVB-T") and slot.config.configMode.value != "nothing":
+						nimcount += 1
+				if nimcount > 1:
+					self.list.append(getConfigListEntry(self.indent % _("Auto detect region"), self.nimConfig.smart_region, _("If you use more than one DVB-T tuner with different regions, you can enable this option to select the correct tuner when zap in service.")))
+				elif self.nimConfig.smart_region.value:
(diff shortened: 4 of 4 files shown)


WanWizard:
What exactly is the point of this, as you already have to define the region before you scan, and regions don't tend to change overnight?

Dima73:
As soon as I have both enabled, whenever I tune a DVB-T local channel, it tries to tune the UK server and fails. It never tries the second tuner, unless I setup a PiP or recording etc. I guess this needs a support inside enigma2, maybe to give IDs to DVB-T tuners and this tuner ID should follow the service scanned, in order to only tune to this ID. Something like "satellite position" for DVB-T

littlesat:
Sorry I also do not get the purpose.... how can you have two networks on one location?

Dima73:
Please see this topics
https://forums.openpli.org/topic/40109-slow-zap-for-ci/?view=findpost&p=1152263
https://forums.openpli.org/topic/78881-satip-support-improvements/?view=findpost&p=1259129

If I had a more powerful antenna directed to Lithuania, then I would be able to receive the Lithuanian network, and not only the Liepaja / Latvia network.

WanWizard:
> Sorry I also do not get the purpose.... how can you have two networks on one location?

You can have the same in the UK and NL too, different transmitters are close enough to receive with a second antenna / tuner, either different regional ones, or in the case of NL, German or Belgian transmitters...

Huevos:
If it was a proper system those 2 aerials would be combined onto one cable and available to all tuners. Having different aerials plugged into different tuners is a broken system.

This is a workaround for badly set up hardware.

Dima73:
There can be several use cases for this option.
This is not a workaround at all, but a real binding to the region (list of frequencies).
Approximately as it is implemented in the satellite system, there the check is carried out by satellites.

WanWizard:
> If it was a proper system those 2 aerials would be combined onto one cable and available to all tuners. Having different aerials plugged into different tuners is a broken system.

Why is that a broken system? You do the same for two different sats?

What you suggest is technically impossible, same as connecting two LNB's directly on one cable, because the transponder frequencies of both are the same or overlap.

Ideally you should have a numbering of configs (aka a fake sat position), and use the namespace to point to the config that needs to be used, exactly how it works for sat positions. But that means a really major change of Enigma (where FFFF and EEEE is hardcoded virtually everywhere).

Huevos:
If 2 signals are on the same frequency you will never be able to receive both even if you have a big yagi. For anything else you should use a proper mixer and bring all signals down the same cable.

WanWizard:
If the transmitters are in line, yes, but not if they are not, you could receive them just fine. 

Espeicially for people living close to a border, the "local" transmitter and the one across the border will virtually never be in line.

Huevos:
I've worked professionally as an antenna installer for over 20 years and I have never come across the situation you describe. In the DVB-T world if 2 transmitters are on the same frequency it would be a single frequency network, i.e. both masts would be carrying identical signals. In all other cases transmitters are never on the same frequency in overlap areas. Why would they be? It would only cause chaos, shrinking the usefulness of both transmitters.

WanWizard:
As an expert, how would you install this scenario for a client:

Have a look at https://nl.wikipedia.org/wiki/DVB-T-frequenties

You'll see for the locations Almere and Amersfoort there are a lot of overlapping channels (which are not important because they carry the same programmes) and one channel which is different for each, and which carries the public and regional channels.

If you live in Hilversum, Almere is 15km North, Amersfoort is 25km South-East. 

The goal is to receive all encrypted channels (from either location as they are on the same channels 24/27/49/57), public channel 26 (Almere) and channel 50 (Amersfoort).

If you would try to receive both using a single omni-directional antenna and single tuner, would you be able to do a scan, given the fact most frequencies of both transmitters are the same? And see all 6 unique channels?

And how would you receive both if the signal is so weak that you need two directional antenna's (for example to cover longer distances, say cross-border)? With a single tuner?

If it can be achieved without a second tuner, this is not needed, and the user can use your method instead.

Huevos:
@WanWizard, bouquets 2, 3, 4, and 5 are a single frequency network. The frequency, ONID, TSID, and SIDs are identical, as is the data. The whole transport stream is identical and transmitted in phase. What you are suggesting is impossible because the service refs would be identical (even subnet),. That is how a single frequency network works. Everything is identical.

In the case of bouquet 1, the ONID, TSID and SIDs are identical. The local channels are using identical SIDs even though the data on the transport stream is different. They only possible way to receive both using enigma would be by enabling namespace subnet, otherwise, you would have service ref collision.

The LCNs for all muxes are carried in the SI tables of bouquet 5 mux, and on no other muxes. That is a national level SI table. The local channels drop into there correct spots because their SIDs are identical, even though their data stream is not.

But anyway as bouquet 1 muxes are on different frequencies they could easily be mixed onto the same cable without problems.

For your infomation...

Bouquet 1, Almere TSID: 0x08A6
Bouquet 1, Amersfoort TSID: 0x08A6

Bouquet 2, Almere TSID: 0x08A3
Bouquet 2, Amersfoort TSID: 0x08A3

Bouquet 3, Almere TSID: 0x08A4
Bouquet 3, Amersfoort TSID: 0x08A4

Bouquet 4, Almere TSID: 0x08A5
Bouquet 4, Amersfoort TSID: 0x08A5

Bouquet 5, Almere TSID: 0x08C4
Bouquet 5, Amersfoort TSID: 0x08C4

Hopefully the above notes demonstrate that it is either a single frequency network, or the transport streams are on different frequencies and could be mixed onto the same cable.

In my opinion this is just a nasty workaround for people with incorrectly set up antenna systems.

Dima73:
Ok my example,  I was writing this code.
NIM Socket 10:
        Type: DVB-T
        Name: Sundtek DVB-T (III) (2/0) (USB):VTUNER
        Frontend_Device: 10
NIM Socket 12:
        Type: DVB-T
        Name: A835:VTUNER

	<terrestrial name="Latvia - Liepāja, Embūte, Ēvarži, Lielauce bezmaksas (Europe DVB-T)" flags="0" countrycode="LVA">
		<transponder centre_frequency="474000000" bandwidth="0" constellation="3" system="0"/>
	</terrestrial>
	<terrestrial name="Latvia - Liepāja (Europe DVB-T)" flags="0" countrycode="LVA">
		<transponder centre_frequency="474000000" bandwidth="0" constellation="3" system="0"/>
		<transponder centre_frequency="490000000" bandwidth="0" constellation="3" system="0"/>
		<transponder centre_frequency="514000000" bandwidth="0" constellation="3" system="0"/>
		<transponder centre_frequency="570000000" bandwidth="0" constellation="3" system="0"/>
		<transponder centre_frequency="586000000" bandwidth="0" constellation="3" system="0"/>
		<transponder centre_frequency="618000000" bandwidth="0" constellation="3" system="0"/>
	</terrestrial>

I took a simple signal splitter and split it into two tuners.
frequency="474000000" it is a free  network and has a very strong reception signal.
All other frequencies on the tuner A835 'tune failed'.
But these frequencies work fine for tuner Sundtek DVB-T.
I bought a signal booster on purpose, but it doesn't help.

Then I selected different regions and it works great.
A835 -- "Latvia - Liepāja, Embūte, Ēvarži, Lielauce bezmaksas (Europe DVB-T)" 
 Sundtek DVB-T (III)  --"Latvia - Liepāja (Europe DVB-T)"

Why not?

WanWizard:
@Huevos ,

If you connect both directional antanna's to the same tuner, would transmission of the same frequencies from both sides give interference? We're not taking about analog radio here, but about a digital signal.

Your explanation describes the restrictions placed on us because of the way it is currently implemented, with a single hardcoded namespace. ID values, frequencies and LCN entries would all not have been relevant if both transmitters could be given a unique namespace (like the orbital position does for SAT).

Huevos:
@WanWizard, "_Your explanation describes the restrictions placed on us because of the way it is currently implemented_", No, there are no restrictions placed on us due to the software. 

Also, you don't just connect 2 aerials together. You put a mixer inline so the impedance and isolation stays correct.

"_would transmission of the same frequencies from both sides give interference?_" No. That is how a single frequency network works. DVB-T is specially designed to deal with reflections. So one signal comes straight from the transmitter to your aerial, while at exactly the same time the same signal leaving the transmitter doesn't come to your aerial directly but bounces off some other object before reaching your aerial. So we now have 2 paths of slightly different lengths. That means the signal from 1 path arrives at the aerial slightly earlier than the signal taking the other path. This is known as an echo. DVB-T is hardened to deal with echos. This is what the guard interval is for. So the receiver accepts the main signal and ignores the echo.

A single frequency network is a special kind of echo. 2 transmitters sending the exact same transport stream, and due to the different distances between the transmitters and the user, 1 signal arrives a fraction later than the other. The receiver just treats the delayed signal as an echo and ignores it. This is only possible because both transmitted signals are absolutely identical. This is why the muxes of regional channels are not transmitted on the same frequency.

The reason the booster does not work is because it is a cheap piece of junk. A proper booster only amplifies one mux. And you have one booster for each mux and turn them up individually until all signals are at a balanced level (not some muxes stronger than others).

The most likely cause of the problem is that the tuner is flooded by the strong mux on 474 and this is making it deaf to the other weaker muxes. Adding a booster is just going to make this worse. On the other hand, adding an attenuator to the problem tuner might even improve the problem. Anyway, to me it is obvious it is a hardware problem because one tuner works and the other is stone deaf.

Should we really be trying to fix this hardware problem by hacking the enigma2 software? This is "fix" that only one person on the planet is ever going to use and should be in a personal build.

WanWizard:
Ok, thanks for enlightening me.

Huevos:
There is a better explanation of SFN on wikipedia.

https://en.wikipedia.org/wiki/Single-frequency_network#DVB-T_SFN

Dima73:
`Should we really be trying to fix this hardware problem by hacking the enigma2 software? This is "fix" that only one person on the planet is ever going to use and should be in a personal build.`
This situation arose completely by accident when I was checking the already written code.
This code is not written for this situation.
That's a coincidence.
I just started checking it and it came in handy, at least by accident.
And I think my case is no exception.
I believe that a lot of people will find the use of this option.
You are trying to show people how to connect two different antennas.
But if they do not want / cannot, for some reason?
This option is for them.

Huevos:
This is not a real-world situation. The main tuner is crap. This is obvious because the Sundtek works fine. 

A can't believe you want to hack enigma to cure a (personal) hardware problem, and then dress it up to look like it is a useful feature. If the tuner doesn't work properly disable it.

Or alternatively, if you want to hack enigma to work around your personal hardware fault do it in a personal build.

Dima73:
`hack enigma`
How easily you all judge that this is a hack.

For me, this is a very simple and easy solution for binding to a region, through a frequency check.
I repeat, you simply do not take into account / do not imagine all the situations when this opportunity will be needed by the user.

So you decided for me that I would have to delete this tuner.
But this was a simple example, situations can be different.

Huevos:
And the other thing I can't believe is the hacks that you have added to enigma and surrounding plugins for Sundtek. No idea why PLi has allowed this.

Huevos:
> 
> 
> `hack enigma`
> How easily you all judge that this is a hack.
> 
> For me, this is a very simple and easy solution for binding to a region, through a frequency check.
> I repeat, you simply do not take into account / do not imagine all the situations when this opportunity will be needed by the user.

And that is something you want to do because you have a hardware fault, not because it is a good idea.

Dima73:
That's funny.
The openVIX image that you supervise is half composed of these exceptions, which you call hacks.

Huevos:
> 
> 
> That's funny.
> The openVIX image that you supervise is half composed of these exceptions, which you call hacks.

How is the content of OpenViX enigma2 repo related to whether or not you should hack OpenPLi?

But if you want to start that discussion you are welcome to on the OpenViX forum. This is not the place.

Dima73:
This is called double standards.

Note that I did not start this.
What is a hack for openPli is just the norm for other images.
And so you do not argue,  calling a 'dirty hack' with this terrible word.
Everything is ok there.
This is ridiculous.

Huevos:
NimManager.py, SatConfig.py, dvb.cpp and frontend.cpp are identical to PLi, including the hacks.

EB-TNAP:
Having a way to control 2 or more separate terrestrial tuners in OpenPLi would be a Very Good option for me. The closets thing to this right now is the slowzap plugin. 
Today, I have one large terrestrial Yagi up around 15 meters, and another UHF multi element antenna below it. 
Terrestrial Transmitters are in several different directions for my location. The top Yagi is rotatable for 360 degrees, and the bottom uhf antenna is fixed. A combiner is not a good option for this antenna system for more than one reason, so I use separate tuners for each antenna.

I also enjoy satellite fta receivers that have more than one tuner because I have separate antennas for c and ku bands. It would be Very Nice for to use 2 terrestrial tuners in OpenPLi in a similar way that 2 satellite tuners can be used for c and ku bands.

EnoSat:
The situation is similar with us.
CZ mux - 26 channel horizontally (antenna 1), SK mux 26 channel vertically (antenna 2). Ideally, use two tuner inputs and not have to buy an expensive programmable amplifier with channel conversion - since classic merging is not possible. Therefore, I find this idea useful, although for most users inland it will not use it (perhaps only with the exclusion of transmitters).

Otherwise, it is probably unnecessary to produce two-input tuners when only one input is to be used and LOOP is preferred.

WanWizard:
I think you have the same problem for cross-border transmitters, it is the only way to receive DVB-T from two different countries. 

Which is very likely if you live close to the border, for example the entire east  of the Netherlands grows up with watching German TV, or in the south, watching Belgian TV. 

You can't merge those on a single cable too...

Huevos:
> 
> 
> I think you have the same problem for cross-border transmitters, it is the only way to receive DVB-T from two different countries.
> 
> Which is very likely if you live close to the border, for example the entire east of the Netherlands grows up with watching German TV, or in the south, watching Belgian TV.
> 
> You can't merge those on a single cable too...

And that isn't a problem with enigma. It is a problem with a foolish administration that put transmitters on the same frequency in overlapping areas.

If people want a hack like this they should build their own image, not vandalize the project for everybody else.

Huevos:
Hack after hack after hack. Hacks in enigma for Sundtek tuners. Hacks in blindscan plugin. Hacks for LNBs for 36/56 east. Hacks to ChannelSelection for 2 line display. More hacks in ChannelSelection to fix the previous hacks. And now this. All terrible shortcut implementations. When is it going to end? I thought OpenPLi prided itself on quality code. How many people benefit from all this hacking? One? All this does is add to the spaghetti and make the coder's work even more difficult.

Huevos:
> 
> 
> The situation is similar with us.
> CZ mux - 26 channel horizontally (antenna 1), SK mux 26 channel vertically (antenna 2). Ideally, use two tuner inputs and not have to buy an expensive programmable amplifier with channel conversion - since classic merging is not possible. Therefore, I find this idea useful, although for most users inland it will not use it (perhaps only with the exclusion of transmitters).
> 
> Otherwise, it is probably unnecessary to produce two-input tuners when only one input is to be used and LOOP is preferred.

It is very unlikely you would be able to receive both transmitters on the same frequency even if you had a big yagi. Maybe you would get one but not the other. Typical yagi would have 15 dBi or less gain. To get that both signals at usable levels the 2 transmitters would have to be exactly 90 degrees from each other to effectively null one out, which is unlikely. So if you were lucky you might manage to increase one transmitter by 6dB. You need a good 12-14dB for stable terrestrial reception so the second transmitter would need to be several dB down to start with in order to get the required signal.

If by some fluke you are able to receive a usable signal from both transmitters with a big yagi, just add a rotator, and you are done. No need to hack enigma.

BTW, don't forget this hack is designed for a deaf tuner. The fairy tale ability to be able to receive two different transmissions on the same frequency was just window dressing to get this ridiculous hack accepted.

EnoSat:
> > It is very unlikely you would be able to receive both transmitters on the same frequency even if you had a big yagi. Maybe you would get one but not the other. Typical yagi would have 15 dBi or less gain. To get that both signals at usable levels the 2 transmitters would have to be exactly 90 degrees from each other to effectively null one out, which is unlikely. So if you were lucky you might manage to increase one transmitter by 6dB. You need a good 12-14dB for stable terrestrial reception so the second transmitter would need to be several dB down to start with in order to get the required signal.
> 
> > If by some fluke you are able to receive a usable signal from both transmitters with a big yagi, just add a rotator, and you are done. No need to hack enigma.
> 
> > BTW, don't forget this hack is designed for a deaf tuner. The fairy tale ability to be able to receive two different transmissions on the same frequency was just window dressing to get this ridiculous hack accepted.

Planned added support for DVBT rotator in enigma ?

Dima73:
Let's postpone this question for the future.

PR #2761, opened 2020-10-17
LinerationSans-Regular font - added georgian chars

ims21:


ims21:
merge it only, if +28kB binary will not be problem

PR #2764, opened 2020-10-20
[skin] allow scaling of skin elements and fonts based on the current …

Huevos:
…screen resolution

This will allow screens from the default skin to be used at all resolutions.

"f" is used to represent skinFactor in formulas.

Example use:
<ePixmap position="0,0" size="974*f,148*f" pixmap="infobar/1stinfobar.png" alphatest="on" zPosition="-1" scale="1"/>

Note 1: in the example above the scale flag is used so images correctly fill their respective cell.

Note 2: In TemplatedMultiContent all formulas must evaluate to int.


Huevos:
The following screengrabs are examples of oe-a skin_default (720 resolution skin) running with this code at 1080 resolution (with "SImple 1080" as the currently active skin).

PR #2778, opened 2020-11-11
[font.cpp] stop enigma crash in case of font reading errors

Dima73:
Log:
[eListboxPythonConfigContent] second value of tuple is not a
tuple.
[eTextPara] renderString: No current_face->size!

dmesg

<6>[    1.368418] ehci-brcm f0480300.ehci: EHCI Host Controller
<6>[
1.373997] ehci-brcm f0480300.ehci: new USB bus registered, assigned bus
number 3

restart enigma2

This happens very rarely, but openATV also had a solution for this error

https://github.com/openatv/enigma2/commit/e8a14ed031c225353a77a168aa05beca339355df#diff-dedda219c4527313d2191fb406791b753dfd0f245b787f571ba2e71df5f4371a

### DIFF
--- a/lib/gdi/font.cpp
+++ b/lib/gdi/font.cpp
@@ -645,12 +645,19 @@ int eTextPara::renderString(const char *string, int rflags, int border)
 	singleLock s(ftlock);
 
 	if (!current_font)
+	{
+		eWarning("[eTextPara] renderString: No current_font!");
 		return -1;
-
+	}
 	if (!current_face)
-		eFatal("[eTextPara] renderString: No current_face!");
+	{
+		eWarning("[eTextPara] renderString: No current_face!");
+		return -1;
+	}
 	if (!current_face->size)
-		eFatal("[eTextPara] renderString: No current_face->size!");
+		eWarning("[eTextPara] renderString: No current_face->size!");
+		return -1;
+	}
 
 	if (cursor.y()==-1)
 	{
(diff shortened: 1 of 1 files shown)


WanWizard:
I've seen it a few times as well, and it's also reported in the forum.

What is the side-effect of no longer treating this as a Fatal?

Dima73:
No side effects.
I don't know why there is such an error, perhaps because of the skin in which there is a lot of different information, and different fonts are used.


Let's wait for the comments of other participants.

Perhaps the OPNATV option is more correct, I don't know ...

IanSav:
I am no expert in this area of the code but I wonder if allowing this code to continue could create issues in the code that calculates widget sizes based on font parameters.  When the code crashed these calculations could not proceed.  Now that we try to continue there may be consequential issues.

I am not objecting to this change.  We should just be careful to ensure that any crash logs that have these debug log entries present immediately prior to the crash need to be checked that the font issue is not the actual cause of the crash.

AbuBaniaz:
If we continue instead of crashing, any missing/miscalculation of font should result in missing/poor display. And you'd hope the person affected will post on the forum with the log showing the issue.

OpenATV have had their commit for four years. Surely any issues with the commit would have been picked up by now?

WanWizard:
Ok.

IanSav:
Given that no one can force the original problem to occur I don't agree that the implications of this change will always be harmless.  For example, what if a missing font size is used as a divisor in a calculation...  Now that the code carries on regardless this may trigger a divide by zero error crash!

If someone can force the original issue and then test the results of this change then there would be no need to guess the results of this change.  I am not asking for the change to be reverted but that attention be paid to *any* issues that occur soon after a potential font issue.

PR #2795, opened 2020-11-26
epgcache: clear opentv eit map

LraiZer:
clears residual data after each pid has been processed.

m_OPENTV_EIT_map.clear() replaces previously removed m_OPENTV_EIT_map.erase(it) with its missing sanity check


WanWizard:
Thanks!

PR #2796, opened 2020-11-26
Update all po files against current sources at develop

blzr:


WanWizard:
What did you do differently?

PR #2801, opened 2020-11-27
all po files updated from develop at 2020-11-27 17:06

pr2git:


blzr:
and why you insist on updating all po's once again?
I assume nothing's changeed since this:
https://github.com/OpenPLi/enigma2/commit/01f84dd8a9dad1200c679f0cab50c16cdaf02138
or?

pr2git:
I also update the french translation in it !!!!!!!!
I always performed like this and this was never a problem Harro just approuve the PR in the wrong order once again you post you pl translation after I already create a PR to update them all + FR so it should be your PR that should be redone not mine.
It is the last time today that I generate again a translation update.
If it is not approved french won't be fully translated.

blzr:
no reason to get mad ;-)
I'm just really curious why necessarily you need to submit translation update  t o g e t h e r  with all po files update (which is imho pointless since they've all been updated couple of hours ago and nothing has changed since?)?
why submitting just translations update is not an option?

pr2git:
blzr please I am managing the po files since years now.
You came from nowhere, please focus yourself on the nl.po update and let me update all the po files, like I am doing since years now. You already make me lose too much time.

pr2git:
Because most translators are not able to generate the updated files so I always generate them all so they just pick it up and work on it.
If you were active in translation on the OpenPLi forum you will have mention it.
But since you look like a troll I will no longer answer you. As you know: don't feed the Trolls.

blzr:
well, no problem for me at all, I'll focus on something more enjoyable than calming someone's bad feelings, since I really have no idea  what I did wrong and where  a c t  u a l problem is (?)  ;-)

PR #2826, opened 2020-12-27
Galician translation added to enigma2

qosobrin:
As per my conversation with @persianpros in https://github.com/oe-alliance/openmultibootmanager/pull/62 here you have the Galician translation of enigma2. It is based on the Spanish file but only to get the original strings because I found it to be very improvable.

I hope you will publish the promised guide on how to create and maintain a language in enigma2, specially for those like me that want to work upstream for the translations to move downstream to the different images.

Thank you very much. Regards.


persianpros:
Maybe gl.po isn't the right name but es_GL.po would be the correct one, don't worry I will check this and add a new PR with Galicia flag: https://en.wikipedia.org/wiki/Flag_of_Galicia

@norhap just to let you know we have another Spanish friend :+1:

blzr:
gl.po is perfectly correct for Galician language ;-)

qosobrin:
> gl.po is perfectly correct for Galician language ;-)

I absolutely agree with @blzr that gl is a correct name for the locale, since there are no geographical variations of the Galician language. But, if @persianpros insisted on using the long form, it should be gl_ES instead of es_GL In any case, I recommend using the short form that is completly valid.

persianpros:
I don't insist on anything :)

Just let me add python needed things too and check the whole thing that's all 👍

blzr:
@Persian Prince
FWIW, I've got all the Galician patches basically ready, including fakelocale...
but I definitely wouldn't want to get in your way if you're already working on it (from my experience some people happen to be quite sensitive about this, you know ;-)) so I'd like to ask just in case before I submit PRs
?

norhap:
> @norhap just to let you know we have another Spanish friend 👍

I am glad to know that a Spaniard is involved in a translation, it is good that a translation is made for Galician, Catalonia had it for a long time, congratulations @qosobrin

persianpros:
@blzr 

I'm not sensitive and also on vacation so I would be happy to see your PRs :)

blzr:
Lucky you! ;-)
I was supposed to be too, oh well...

blzr:
all PRs already merged

persianpros:
Great and thank you :+1: 

This PR could be closed now and for future updates @qosobrin could just update https://github.com/OpenPLi/enigma2/blob/develop/po/gl.po ;)

qosobrin:
Please, close this PR if you consider that everything needed is done. Thank you very much all of you for your wonderful work.

Best regards.

PR #2835, opened 2021-01-07
[frontend]  add 5V active antenna for hisilicon via proc

Dima73:


lib/dvb/frontend.cpp
```diff
@@ -2636,17 +2635,24 @@ RESULT eDVBFrontend::connectStateChange(const sigc::slot1<void,iDVBFrontend*> &s
...
 	switch (voltage)
 	{
 		case voltageOff:
 			m_data[CSW]=m_data[UCSW]=m_data[TONEBURST]=-1; // reset diseqc
 			vlt = SEC_VOLTAGE_OFF;
+			active_antenna_power = 0;
 			break;
 		case voltage13_5:
 			increased = true;
 			[[fallthrough]];
```
pieterg, line 2651: this fallthrough will cause active_antenna_power to end up being 1

lib/dvb/frontend.cpp
```diff
@@ -2659,8 +2665,22 @@ RESULT eDVBFrontend::setVoltage(int voltage)
 		default:
 			return -ENODEV;
 	}
+
```
pieterg, line 2668: whitespace?

lib/dvb/frontend.cpp
```diff
@@ -2659,8 +2665,22 @@ RESULT eDVBFrontend::setVoltage(int voltage)
...
 	if (m_simulate)
 		return 0;
+
+	if (active_antenna_power != -1 && (m_type == feTerrestrial || m_type == feCable))
+	{
+		char filename[256];
+		snprintf(filename, sizeof(filename), "/proc/stb/frontend/%d/active_antenna_power", m_slotid);
+		CFile proc_name(filename, "w");
```
pieterg, line 2676: most hardware does not require a proc entry, do you really need to try and open it on all platforms?

Dima73: Your suggestion?

lib/dvb/frontend.cpp
Dima73: I did this on purpose in order to separate the code for understanding.

lib/dvb/frontend.cpp
Dima73: https://github.com/OpenPLi/enigma2/pull/2835/commits/8fe25af6f169f6205bfc4e533da5539dec18ec2d

pieterg: the commit does not seem related?

lib/dvb/frontend.cpp
pieterg: 1. ask the manufacturer to use FE_SET_VOLTAGE
if that fails
2. use SystemInfo?

littlesat: When using SystemInfo it has to be swigged via python somehow... then it can also be done via an addNotifier on the config in python and only add the Notifier when the SystemInfo tells to do it via the proc... 

What you do is always a hack. The real solution is ask the manufacturer to change it indeed...

littlesat: I suggest we should at least add the systeminfo check...

Dima73:
Ok.

lib/dvb/frontend.cpp
eriksl: I really think setting supply power for a DVB-T antenna is fundamentally something different that using line voltage to do signalling towards a LNB. The first is something relatively static, the power should either be static off (if the antenna is passive). or it should be static on (active antenna) or it should be on/off at the same time as the tuner (dynamic power saving scheme). Thus the power should only be switched when opening and closing the tuner.

This is something else than DVB-S singnalling which is required for each channel change action. Even Multicable and JESS rely on them.

Therefore the code for setting both should not be mixed and should not be at the same place in enigma. Signalling is to be left where it is now, while power supply for DVB-T should be added to the code where the tuner is opened or closed.

I also follow the reasoning from Zgemma to not (mis-)use the DVB-S signalling iotctl for this. The only thing I find less ideal is that they choose to use /proc entry, I'd rather have seen a (separate!) ioctl.

persianpros:
https://github.com/OpenVisionE2/enigma2-openvision/commit/f17ee1ea399e9e96628bf2406a1f031e2efdbdae

pieterg:
they have the same fallthrough issue

littlesat (commented):
Isn't it better to switch on the voltage in openFrontend.. (when required) en switch it off in closeFrontend...

It shouldn't be switched on/off everytime you retune?

persianpros:
Dima's code is now in Open Vision but needs a MACHINE_FEATURE to be activated so it won't harm for now.

I asked our devs and testers to check it on Zgemma, Octagon and GigaBlue HiSilicon models.

littlesat:
Dima's code is in fact wrong.... the 5V should be arranged when you 'enable' the frontend end 'disable' the frontend and -NOT- at each single zap....  Doing it for each single zap is mandatory for DVB-S... -NOT- for DVB-T.....

Actually the ioctrl was already a work-a-round hack....

Why not immediately do it right?

In ADDITION first tell ZGemma to adapt the code it something I prefer...!!!!!

When some other community simply accept it then it is getting a next never ending story and manufacturers can do what they like...:(

persianpros:
We didn't accept it, it's a switch for devs and we will remove it soon.

Dima73:
Not only zgemma,all hisilicon model.

Dima73:
There will be a new version. 
https://github.com/OpenPLi/enigma2/pull/2842

PR #2852, opened 2021-02-08
[Tools.Directories] SCOPE_FONT: Check the "fonts" folder

Huevos:
Before this commit it was necessary to use a path when accessing a font from the current skin. It makes sense, if we are trying to locate a font, that the "fonts" folder is tested directly as this is the most likely place a font file will be located in the currently active skin.

e.g. to access...
<font name="Regular" filename="/usr/share/enigma2/<skin-name>/fonts/SFNSDisplay-Regular.otf" scale="100" />
or this...
<font name="Regular" filename="fonts/SFNSDisplay-Regular.otf" scale="100" />

it is now possible to do this...
<font name="Regular" filename="SFNSDisplay-Regular.otf" scale="100" />

It is wrong that it is possible to access a font located in "/usr/share/fonts" without using a path, but if accessing a font in "/usr/share/enigma2/<skin-name>/fonts" a path is required. This commit allows consistence between these locations.

Consistency needs to come first. We shouldn't have confusing traps like this in enigma.

Also remove some comments that if ever implemented would wreak havoc throughout the enigma community.

### DIFF
--- a/lib/python/Tools/Directories.py
+++ b/lib/python/Tools/Directories.py
@@ -127,11 +127,11 @@ def resolveFilename(scope, base="", path_prefix=None):
 		resolveList = [
 			os.path.join(defaultPaths[SCOPE_CONFIG][0], skin),
 			os.path.join(defaultPaths[SCOPE_CONFIG][0], "skin_common"),
-			defaultPaths[SCOPE_CONFIG][0],  # Can we deprecate top level of SCOPE_CONFIG directory to allow a clean up?
+			defaultPaths[SCOPE_CONFIG][0],
 			os.path.join(defaultPaths[SCOPE_SKIN][0], skin),
 			os.path.join(defaultPaths[SCOPE_SKIN][0], "skin_fallback_%d" % getDesktop(0).size().height()),
 			os.path.join(defaultPaths[SCOPE_SKIN][0], "skin_default"),
-			defaultPaths[SCOPE_SKIN][0]  # Can we deprecate top level of SCOPE_SKIN directory to allow a clean up?
+			defaultPaths[SCOPE_SKIN][0]
 		]
 		for item in resolveList:
 			file = os.path.join(item, base)
@@ -148,11 +148,11 @@ def resolveFilename(scope, base="", path_prefix=None):
 		resolveList = [
 			os.path.join(defaultPaths[SCOPE_CONFIG][0], "display", skin),
 			os.path.join(defaultPaths[SCOPE_CONFIG][0], "display", "skin_common"),
-			defaultPaths[SCOPE_CONFIG][0],  # Can we deprecate top level of SCOPE_CONFIG directory to allow a clean up?
+			defaultPaths[SCOPE_CONFIG][0],
 			os.path.join(defaultPaths[SCOPE_LCDSKIN][0], skin),
 			os.path.join(defaultPaths[SCOPE_LCDSKIN][0], "skin_fallback_%s" % getDesktop(1).size().height()),
 			os.path.join(defaultPaths[SCOPE_LCDSKIN][0], "skin_default"),
-			defaultPaths[SCOPE_LCDSKIN][0]  # Can we deprecate top level of SCOPE_LCDSKIN directory to allow a clean up?
+			defaultPaths[SCOPE_LCDSKIN][0]
 		]
 		for item in resolveList:
 			file = os.path.join(item, base)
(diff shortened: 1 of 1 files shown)


IanSav:
I disagree with this commit.  It add extra searches that will slow the system down for little to no gain.  All fonts are meant to be in /usr/share/fonts/.  Throwing fonts all over the place is going to bloat the images and make a real mess.  I am also interested to see what a mess this will make of the package manager when different skins try to manipulate fonts in the same directory.  It will likely end in tears.

If you want a font for all skins add it to the correct place (/usr/share/fonts) if you want a font for your skin then add it to your skin.

Huevos:
Of course, you would want to sabotage it. Not going to bother to discuss why you want inconsistent SCOPE_FONT based on where the font file is located.

WanWizard:
I agree with @IanSav, packages managers can NOT handle the same file being under version control in multiple IPKs. Fonts are either system fonts, in seperate font packages, installed in /usr/share/fonts, or skin specific fonts, as part of the skin.

p.s. nobody is stopping skin builders to split their skin into multiple packages. When building from source not a problem at all, including the correct dependencies. For skin builders distributing only ipk's, well, ...

RobvanderDoes:
No: skin fonts should be in the skin folder, not in a generic font folder.
That would mean that the contents of a skin should be populated over several system folders, which would make it difficult to oversee what comes from where.

If you want to see how fast such a search is, use the fast skin reload (meant for skinners) as is currently in use by ViX. In a split second the new or reloaded current skin in on GUI.

Huevos:
So you think...
`<font name="Regular" filename="SFNSDisplay-Regular.otf" scale="100" />`

Should search "/usr/share/fonts" but skip `/usr/share/enigma2/<current-skin>/fonts`?

This is as crazy as Ian's plan to delete "skin_default" in paths everywhere. Ended up with hundreds of changes being reverted.

WanWizard:
What uses that directory? I've never seen "/usr/share/enigma2/fonts" anywhere. 

If that is used by some skin author to share fonts between several skins, it should be a separate font package, that installs in "/usr/share/fonts", which is the linux standard? Or they should be included in the skin, in "/usr/share/enigma2/\<skin\>/fonts".

I have a feeling this is deemed needed because of a decision in the past of some skin designer?

The only skin I have here that uses its own fonts is a OpenPLi port of MetrixHD, which defines:
````
  <fonts>
    <font filename="/usr/share/fonts/lcd.ttf" name="LCD" scale="100" />
    <font filename="/usr/share/fonts/ae_AlMateen.ttf" name="Replacement" scale="100" replacement="1" />
    <font filename="/usr/share/fonts/tuxtxt.ttf" name="Console" scale="100" />
    <font filename="/usr/share/fonts/nmsbd.ttf" name="Arial" scale="100" />
    <font filename="/usr/share/enigma2/MetrixHD_MOD/fonts/segoeui.ttf" name="Regular" scale="100" />
    <font filename="/usr/share/enigma2/MetrixHD_MOD/fonts/segoeui.ttf" name="RegularLight" scale="100" />
    <font filename="/usr/share/enigma2/MetrixHD_MOD/fonts/segoeb.ttf" name="RegularB" scale="100" />
  </fonts>
````

Huevos:
No no no, `/usr/share/fonts` is the directory of the preinstalled fonts. I did not say `/usr/share/enigma2/fonts`. I said `/usr/share/enigma2/<current-skin>/fonts`

@WanWizard your skin is doing it wrong. Should be...
`<font filename="fonts/lcd.ttf" name="LCD" scale="100" />`

My commit means that could be changed to...
`<font filename="lcd.ttf" name="LCD" scale="100" />`

That would be exactly the same markup required to fetch the font from "/usr/share/fonts".

What is wrong is there is no consistency between one look up and the other.

No idea what the IPK comment is related to. This is just about SCOPE look ups.

Ian's argument is about number of look ups, but this too is false. In OpenViX a complete skin reload takes 400 milliseconds with and without this code (no enigma restart needed).

Huevos:
Also in the current code if I use `fonts/font.ttf` in the widget that can never drop back to fetch the font from `/usr/share/fonts` should that be necessary.

Anyway if you guys want to leave this lookup broken that is up to you.

WanWizard:
Well, you did, but corrected it after my post :smile:.  

The rest of the comments were related to the post before the typo correction, so no longer relevant.  I can't comment on the performance issue, but I can imagine that, with the limited amount of fonts in a skin, and only a lookup on load, the performance hit of a few additional stat calls is limited?

Since I don't agree with Ian that all fonts must be in /usr/share/fonts, but only fonts installed by separate font packages, I don't have any further objections. @littlesat, agreed?

And not my skin btw, and very old (but I happen to have it here so I could check).

Huevos:
No, what happened was this...
`/usr/share/enigma2/<current-skin>/fonts`

It wasn't quoted code so the web output hid the `<current-skin>` part. ;)

Huevos:
Anyway, it's up to you guys, I won't be offended if you don't agree with my logic.

Huevos:
> 
> 
> Well, you did, but corrected it after my post 😄.
> 
> The rest of the comments were related to the post before the typo correction, so no longer relevant. I can't comment on the performance issue, but I can imagine that, with the limited amount of fonts in a skin, and only a lookup on load, the performance hit of a few additional stat calls is limited?
> 
> Since I don't agree with Ian that all fonts must be in /usr/share/fonts, but only fonts installed by separate font packages, I don't have any further objections. @littlesat, agreed?
> 
> And not my skin btw, and very old (but I happen to have it here so I could check).

Yes just limited to a handful of fonts listed in the skin file.

My logic is the most likely place to find a font file is a folder called "fonts".

Forcing a user to use fonts/font.ttf in the widget breaks the fallback to the default directory.

Huevos:
I have tested the timing of this multiple times. Where fonts are located in `/usr/share/fonts` (worst case scenario) the extra lookups added 4 milliseconds loading a skin with 4 fonts. 

With this code:

<113711.9593> 19:44:09.8986 [Font] Adding font '/usr/share/fonts/OpenSans-Regular.ttf' -> 'Subs'.
<113711.9603> 19:44:09.8997 [Font] Adding font '/usr/share/fonts/OpenSans-Italic.ttf' -> 'Subsi'.
<113711.9612> 19:44:09.9006 [Font] Adding font '/usr/share/fonts/OpenSans-Semibold.ttf' -> 'Subsb'.
<113711.9622> 19:44:09.9015 [Font] Adding font '/usr/share/fonts/OpenSans-SemiboldItalic.ttf' -> 'Subsz'.

29 ms

Prior to this code:

<114589.6176> 19:58:47.5569 [Font] Adding font '/usr/share/fonts/OpenSans-Regular.ttf' -> 'Subs'.
<114589.6185> 19:58:47.5578 [Font] Adding font '/usr/share/fonts/OpenSans-Italic.ttf' -> 'Subsi'.
<114589.6193> 19:58:47.5587 [Font] Adding font '/usr/share/fonts/OpenSans-Semibold.ttf' -> 'Subsb'.
<114589.6201> 19:58:47.5595 [Font] Adding font '/usr/share/fonts/OpenSans-SemiboldItalic.ttf' -> 'Subsz'.

25 ms

Huevos:
Is this going to be accepted?

WanWizard:
I was awaiting comments from @littlesat, I have no objections.

PR #2857, opened 2021-02-14
[Satconfig] setup tuners FBCLink  remove all 1.2 diseqc mode

Dima73:
-When FBCRoot not 1.2 diseqc mode tuners FBCLink  not visible
"Additional cable of motorized LNB"

### DIFF
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -989,7 +989,7 @@ class NimManager:
 				nimList.remove(nim)
 		return nimList
 
-	def canDependOn(self, slotid, advanced_satposdepends=False):
+	def canDependOn(self, slotid, advanced_satposdepends=""):
 		type = self.getNimType(slotid)
 		type = type[:5] # DVB-S2X --> DVB-S, DVB-S2 --> DVB-S, DVB-T2 --> DVB-T, DVB-C2 --> DVB-C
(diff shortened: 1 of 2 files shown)


RobvanderDoes:
I hope this is not going to be merged: it would break my setup.
I use a motor with a (dual) Unicable LNB; this allows me to see/record e.g. BBC1/2/4

Dima73:
This patch only tuners FBC  link(virtual) not FBC root.
What are your settings?
You're using tuners FBC  link(virtual) as 1.2 diseqc mode?

RobvanderDoes:
Root = rotor Unicable scr1
Link = extra cable scr2

Dima73:
Root = rotor Unicable scr1 --> 1.2 diseqc mode
Link = extra cable scr2--> 1.2 diseqc mode?

RobvanderDoes:
If you're saying that with your tweak I can still use my setup, it's fine with me of course.

WanWizard:
It's all Python code, so easy enough to test on your particular setup?

RobvanderDoes:
Harro, you're absolutely right to say so, only atm I can't do that.
Nevertheless: this forced me to look a bit better at the PR and I think it won't harm my setup, and it indeed gets rid of an (under the circumstances) useless setting.

PR #2860, opened 2021-02-17
lt.po and fr.po update + add mk_MK.po also in Makefile.am

pr2git:
Probably that more changes are needed to fully support Macedonian since it is a new language in OpenPLi.

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: OBH enigma2\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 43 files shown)


persianpros:
@blzr 

Needs your attention :+1:

pr2git:
Why do you  think that this should be watched? 
This is the same po update as I always do. Why suddenly blzr should watch the PR that I create?

persianpros:
To help you with mk_MK.po and supporting it in related repos not watching your commit ;)

pr2git:
Thanks for the clarification, but I already internally draw the attention of the developpers on it. So don't worry.

PR #2861, opened 2021-02-19
Add Macedonian language

blzr:
also fix some errors in strftime date format in translations


persianpros:
@WanWizard would you please merge this PR and other related ones as now there's https://github.com/OpenPLi/enigma2/blob/develop/po/mk_MK.po inside po folder?

PR #2872, opened 2021-03-03
Update uk.po

badsystem:


WanWizard:
We don't accept PR's on the release branch,, as we can't backport them to develop. 

Please base your translations on the develop branch, so they will end up in the next release.

PR #2889, opened 2021-03-13
[opentv.cpp] Fix some spuriously timed EPG events (#659)

AbuBaniaz:
Fix for occasional spurious events appearing 0x20000 seconds ahead of their correct position in the OpenTV EPG at certain times of the day.

### DIFF
--- a/lib/dvb/opentv.cpp
+++ b/lib/dvb/opentv.cpp
@@ -157,7 +157,13 @@ OpenTvTitle::OpenTvTitle(const uint8_t * const buffer, uint16_t startMjd)
 		uint8_t descriptor_length = buffer[1];
 		uint8_t titleLength = descriptor_length > 7 ? descriptor_length-7 : 0;
 
-		startTimeBcd = (((startMjd - 40587) * 86400) + (UINT16(&buffer[2]) << 1));
+		uint32_t startSecond = (UINT16(&buffer[2]) << 1);
+
+		startTimeBcd = ((startMjd - 40587) * 86400) + startSecond;
+		
+		if (startSecond >= 86400)
+			startTimeBcd -= 0x20000;
+		
 		duration = UINT16(&buffer[4]) << 1;
 
 		//genre content


4l3x2k:
Wow.
That would be a great catch to overcome all of these "spinner after recording".

tvheadend showed that there is a race on adapterX ...

Next approach. I'am happy with that. Please get this in.

@AbuBaniaz 
Great catch!

WanWizard:
I can't imagine this having any effect on a spinner, nor causing one? This is just a small ajustment of the EPG event start time.

AbuBaniaz:
@4l3x2k 
Not my work, so can't take any credit.  I just merged the pull request on ViX and submitted a pull request to PLI

The issue was reported and investigated here (9 pages): 
https://www.world-of-satellite.com/showthread.php?64109-Weird-EPG-errors-EPG-Refresh-Recent-versions-of-OpenViX-e-g-5-4-008

4l3x2k:
> I can't imagine this having any effect on a spinner, nor causing one? This is just a small ajustment of the EPG event start time.

Which can run in parallel to any other action on your stb.
At least ... it would eliminate any races. So I welcome this change!

We alway have a run on viewing, catching epg and streaming.

4l3x2k:
Well...

... this is a real wired topic. The most common sense is that the libc library on mips behaves other than the one on arm. That may be the case, really.

BUT

many things go on parallel nowadays. So this change Is absolutely welcome despite any other reasons!

4l3x2k:
And is therefore merged already

WanWizard:
glibc in OE zeus is absolutely borked on MIPS, it gives all sorts of misery. We're thrinking  about an OE update to address this.

4l3x2k:
> glibc in OE zeus is absolutely borked on MIPS, it gives all sorts of misery. We're thrinking about an OE update to address this.

Yey. We should address this! Would be kind to know what is causing this! We're Heros with bitbake but could miss a pointer on a small change of underlying libraries.

I really tend to have this problem with zeus release of oe on mips. That is the only thing that has changed during enigma development.

Anyway ... this has been merged. We'll see how it goes.

PR #2896, opened 2021-03-19
resolveFilename: allow fallback to svg or png image

Taapat:
If not exit png image check also for svg image and vice versa.
It allows replace png images form screens with svg image in skin or svg image in skin with png from enigma.

### DIFF
--- a/lib/python/Tools/Directories.py
+++ b/lib/python/Tools/Directories.py
@@ -107,7 +107,7 @@ def resolveFilename(scope, base="", path_prefix=None):
 		baseList = [base]
 		if base.endswith(".png"):
 			baseList.append("%s%s" % (base[:-3], "svg"))
-		if base.endswith(".svg"):
+		elif base.endswith(".svg"):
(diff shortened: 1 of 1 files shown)


WanWizard:
Wouldn't it be more logical when you change an image type in a skin from png to svg, that you also change its references in the skin xml?

Taapat:
The logic is to first check the image that is specified in the skin, only if it does not exist then svg or png.
This is necessary for backward compatibility.
In skin_default or in our skin we could change images to svg and use on all resolutions, but therefore should not change skins or plugins as many of them are no longer maintained.

U.P.D.
And vice versa.
After that when in enigma we change the images to svg and also in the enigma code change the link to svg image, we can use the skin containing the image in png.

Taapat:
There are many places in the code where you can't specify an image in the skin, but where its load is included in the python code.
To use such images from the skin you can place the relevant images in the appropriate directory in the skin, but you can not specify in the skin to use them because they are hard pointed in the python code.

WanWizard:
Ok. Clear.

PR #2906, opened 2021-03-27
Move root python files to lib/python and rename mytest to Startup.py

SimonCapewell:
See also https://github.com/OpenPLi/openpli-oe-core/pull/938


persianpros:
We use "StartEnigma" which is better as "Startup" is very general, startup what ;)

https://github.com/OpenVisionE2/enigma2-openvision/blob/develop/StartEnigma.py

You may want to consider it.

Issue #2567, opened 2020-05-11
Parental Control on Subfolder in Favorites not working correctly


tigranbadalyan:
Hi

I have some problem in Pli 7.2, my receiver is Vu+ Uno 4K SE.

This kind of problem, I have IPTV bouquets in favorites with subdirectories for example
fav1
fav2
fav3
in the bouquet fav2 there are still subdirectories fav4 and fav5 and channels respectively in them.
If I use parental control on fav2 then the subdirectories fav4 and fav5 are duplicated in the list of parent favorites list 
fav1
fav2
fav3
fav4
fav5
but olso they are still in the fav2.
somehow it is possible to solve this problem so that fav4 and fav5 do not come out of the fav2
regards Tigran
in the screenshots there is "IPTV Edem TV" favorites with sub folders, if i put it to rarental control,
all the subfolders duplicated to root directory and olso they are still in "IPTV Edem TV" folder.
![1](https://user-images.githubusercontent.com/65174474/81569504-679b9c80-93b0-11ea-9293-3d862ffe37b6.jpg)
![2](https://user-images.githubusercontent.com/65174474/81569510-69656000-93b0-11ea-9695-256ed7f2fcad.jpg)
![3](https://user-images.githubusercontent.com/65174474/81569515-6bc7ba00-93b0-11ea-96ed-8199d0bbd5e5.jpg)

PR #2908, opened 2021-03-30
[ParentalControl] fix subfolder added in favourites list if hide serv…

Dima73:
…ices

more info:
Parental Control on Subfolder in Favorites not working
correctlyhttps://github.com/OpenPLi/enigma2/issues/2567


Dima73:
Attention!
I have not found a way to hide the alternative service

PR #2912, opened 2021-04-03
Allow automatically scaling images in LoadPixmap

Taapat:
Use the scaling factor to rescale svg images to suit the current resolution of the skin.
This allows later remove the skin_fallback_1080 folder and replace the images in this folder with svg images in the skin_default folder that can be scaled without loss of quality.
Variable SKIN_FACTOR there is temporary solution that needs to be changed later, when SkinFactor for enigma components will be fully implemented.


Huevos:
As shown here https://forums.openpli.org/topic/82463-add-support-for-svg-rendering/page-7#entry1334759 this does not work as expected by the author.

PR #2913, opened 2021-04-04
[InfoBarGenerics] add sanity check if extensions list is empty

Dima73:


betacentauri:
Isn't it better to add the check here?
https://github.com/OpenPLi/enigma2/pull/2913/commits/740b44da8be09e54b07b4dc693af3680ec4aba5c#diff-8eba37ae2d46c82544c09a221414a7dd4486d79e0f59efc749c92ca6680dcfe0R2210

Because in the lines above some entries in the extensionsList might be removed and the list might be empty.

Dima73:
> 
> 
> Isn't it better to add the check here?
> [740b44d#diff-8eba37ae2d46c82544c09a221414a7dd4486d79e0f59efc749c92ca6680dcfe0R2210](https://github.com/OpenPLi/enigma2/commit/740b44da8be09e54b07b4dc693af3680ec4aba5c#diff-8eba37ae2d46c82544c09a221414a7dd4486d79e0f59efc749c92ca6680dcfe0R2210)
> 
> Because in the lines above some entries in the extensionsList might be removed and the list might be empty.

extensionsList = self.extensionsList[:] this  list copy
https://forums.openpli.org/topic/68265-enigma2-plugins-development-in-vs-code-with-docker/?view=findpost&p=1334535

betacentauri:
Yes, I meant something like that. This is then not needed anymore.

betacentauri:
Ah, sorry! I thought it's in another pull request. Please wait, I'll commit the change

betacentauri:
Done:
https://github.com/OpenPLi/enigma2/commit/4edd1689478695c1a0a42a16dcbcf22c8f0b4e8e
I think the other part is not really needed, or?

Dima73:
part 2 is OK (not need first part)

PR #2914, opened 2021-04-06
Remove unused shutdown text

SimonCapewell:
QUIT_ERROR_RESTART is not referred to anywhere in the enigma2 code, and
the associated exit code of 5 is never supplied to TryQuitMainloop


WanWizard:
I see exit code 5 is also referred to here: https://github.com/OpenPLi/enigma2/blob/63c424549729aea3a88b748bb43825f55a71a133/tools/enigma2.sh.in#L59

No idea where an Enigma restart would be triggered after "new settings" are installed. @littlesat, any idea?

SimonCapewell:
Yes, I should've mentioned that I found that but ignored it as irrelevant. However, its purpose seems to make the exit code of 5 behave like a UI restart. There's still nothing actually within the Python code to cause an exit code of 5 to ever occur. I expect if we looked at the history of mytest.py, there would be some, now deleted, exception handling that returned 5. Shall I also remove this handler from the .sh file?

WanWizard:
Isn't this triggering it? https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Startup.py#L598

SimonCapewell:
Good find! However, it isn't calling onto the `TryQuitMainloop` or `QuitMainloopScreen` so, whilst the const value is actually still relevant, the "error in mytest" string isn't used.

WanWizard:
True, it was more a response to removing it from the sh file... 

I hope @littlesat can find the time to respond, he might shed some light on its history...

SimonCapewell:
There's also a quitMainLoop(5) here https://github.com/OpenPLi/enigma2/blob/4edd1689478695c1a0a42a16dcbcf22c8f0b4e8e/lib/python/python.cpp#L215

I've reduced the scope of this to just remove the unused shutdown screen message text and translations

PR #2915, opened 2021-04-08
Use StartEnigma name instead of Startup

persianpros:
- This is a better name as we're starting enigma not something else, Startup is very general and common as we may ask startup what?

- Add StartEnigma in debug lines so we understand StartEnigma is producing them.

- Correct related .py files which advertise this file.


persianpros:
@WanWizard 

After https://github.com/OpenPLi/enigma2/commit/6bcab95070404eca98a589d3da0024e1ec8729fc .po file changes are not needed so I took them out.

PR #2920, opened 2021-04-09
opentv: fix long event duration

LraiZer:
duration limit (65535) is not large enough,
20 hour events (72000 seconds) can be seen.

### DIFF
--- a/lib/dvb/opentv.cpp
+++ b/lib/dvb/opentv.cpp
@@ -224,7 +224,7 @@ uint16_t OpenTvTitle::getEventId(void) const
 	return eventId;
 }
 
-uint16_t OpenTvTitle::getDuration(void) const
+uint32_t OpenTvTitle::getDuration(void) const
 {
 	return duration;
 }
--- a/lib/dvb/opentv.h
+++ b/lib/dvb/opentv.h
@@ -77,7 +77,7 @@ public:
 	uint16_t getChannelId(void) const;
 	uint32_t getStartTime(void) const;
 	uint16_t getEventId(void) const;
-	uint16_t getDuration(void) const;
+	uint32_t getDuration(void) const;
 	void setChannelId(uint16_t channelid);
 	void setEventId(uint16_t eventId);
 };


WanWizard:
Thanks. 

Something vaguely related: since another recent OpenTV change, my code that stitches titles back together no longer works on all titles, any idea why (saves me diving into it :grin)?

LraiZer:
Do the events match the criteria for stitching, dots, spaces etc...
or
Do you have EIT and Freesat epg also enabled?

These readers have a higher priority source type id than OpenTV epg and can update the opentv epg with their own. The stiching code was not made as a function that is applied on all epg readers with a switch to enable/disable. It currently only applies to opentv events.

Recently opentv epg now uses the correct eventId from dvb si which also now matches the eventId sent in EIT and Freesat on 28.2e. The epgcache source type was also fixed, so the fixoverlapping code now also gets executed for opentv epg. If EIT or Freesat with the hierarchy to change opentv epg, finds update changes in the event data before opentv finds and updates them first, then their stitching would not have taken place for these events as they were not inserted by opentv.

If you want stitching done across all epg, you should make a function and apply at a later point of insertion used by all readers.?

willow5:
Hi, I have been directed here by betasentauri from the openatv forum. I have an issue with the openatv epg on 28.2° east whereby not all channels are being populated with the epg data. Can you help please?

willow5:
1_0_19_22D9_80D_2_11A0000_0_0_0_20210410125207

WanWizard:
> Do the events match the criteria for stitching, dots, spaces etc...
> or
> Do you have EIT and Freesat epg also enabled?

Yes, and only EIT, no Freesat, no now/next. But I have always had this, and the title issues have only appeared in the last few weeks. This is a typical example:
image
The first three are broken up (even though they have the three dots), the next three are ok (while the event text is identical).

LraiZer:
@willow5
There is nothing in the screenshot to indicate what the service references of the channels without epg are.

OpenTV needs to be enabled in the EPG settings and then you need to tune/ZAP to a channel called "IEPG data 1" for 3 minutes.

@WanWizard

If EIT epg is enabled then it will update changes for the next two programs with its now/next data for events scheduled on that tuned transponder. The prgrams that are still stitched are because you have not tuned to their trasponder and EIT has not updated their next two programs?


Do you have "Show EIT now/next in infobar" enabled in EPG settings?
Try turning that off so that the OpenTV cached data is used instead of the EIT?

willow5:
@LraiZer thank you, I will get the references for you. The main point here is that the channels with populated EPG data are self populated, I have not had to zap to "IEPG data 1" for 3 minutes. Is it possible to get the other channels to self populate and maintain this every 7 days like the other channels or do you need to zap to this channel every 7 days?

LraiZer:
@willow5 
The ones that have populated themselves have done so with Freesat epg. The others with no epg are not from the Freesat provider, they are from the Sky provider.

You need to zap to "IEPG data 1" at least daily to update the epg channels that have not populate themselves.
Most users install a plugin called "EPG Refresh" and set it to wake up and zap to this channel for 3 minutes while they are asleep.

willow5:
@LraiZer then it is clear for me but nobody in the openatv could answer my query so maybe you could add a note to explain this somewhere in the OpenEPG settings for others or include EPG refresh as part of the OpenATV build?

LraiZer:
Here is OpenPLi and for their users it is documented https://wiki.openpli.org/EPG#OpenTV

@willow5 You are now the master and clear on this going forward at OpenATV ;)

WanWizard:
> If EIT epg is enabled then it will update changes for the next two programs with its now/next data for events scheduled on that tuned transponder. The prgrams that are still stitched are because you have not tuned to their trasponder and EIT has not updated their next two programs?

I saw it again last night, on Channel 5, while I had been watching programmes on another sat, so I doubt it is EIT.

> Do you have "Show EIT now/next in infobar" enabled in EPG settings?
> Try turning that off so that the OpenTV cached data is used instead of the EIT?

No, disabled.

I'm beginning to wonder if they are perhaps using the utf-8 ellipsis character instead of three dots, that could explain it. Some debugging is needed.

LraiZer:
My first guess would be a bogus space after the tiltle dots...[space]

There are some on the 5 networks that don't meet the criteria because they have a colon and two dots:..

Title:..
...description

WanWizard:
Yes, those are on my todo list too... :grin:

WanWizard:
In the mean time I can confirm that is indeed EIT on the BBC channels (is that new?) is the culprit. 

I've changed my startup channel to 23,5, wiped all EPG, restarted the box, manually zapped to IEPG, and all titles are ok. As soon as I zap to BBC / ITV / C4 /C5 HD, I get the split title back. As I have only OpenTV and EIT active, but no EIT now/next, only EIT can be the cause?

WanWizard:
p.s. this could also be the reason @willow5 only sees a subset of EPG data, in that screenshot I see exactly the channels with which I have the split-title problem.

LraiZer:
Until recently for OpenTV the sourceId was wrong, and the eventId was random and not as in DVB SI table, these are now fixed.

The sourceId is now correctly identified for all on 28.2e.. EIT(PID:0x12 actual transponder, current/next), Freesat, OpenTV
The eventId is now identical for any given event for all on 28.2e across the different providers.. EIT, Freesat and OpenTV.

AFAIK.. What you are observing now is that the epg is actually working correctly as it was originaly coded to do.

EPG with the hierarchy to update lower ranked sourceId events will now do so on 28.2e for eg. Freesat will update OpenTV.
EIT can now correctly find a matching previously cached OpenTV eventId and will update that OpenTV event with its own data.

Freesat will update OpenTV events for all the next weeks programs
EIT will update OpenTV and Freesat events for the current/next programs

willow5:
Does this mean I no longer need the epgrefresh plugin?

LraiZer:
I don't use EPG Refresh, i just set the startup and standby service as IEPG data 1 via the OpenPLi GUI.

Turn box on in morning and put kettle on and by time you've made a 3 min brew, you can zap to a normal program ;)

Open bouquets and scroll to channel IEPG data 1 and press menu button to set this:

1_0_1_105D_7D4_2_11A0000_0_0_0_20210416211300-iepg

RobvanderDoes:
EPGRfresh is especially useful when you want more than just OpenTV data.

WanWizard:
Indeed. 

I use it to pull in EIT from providers I don't watch that much, take a long time to receive, or require zapping to different transponders to pull in all EPG for that provider. SKY UK is a positive exception, it uses a dedicated transponder with tonnes of bandwidth, and it sends all EPG on that transponder,

If I need to have a brew for the time it takes to pull in all EPG I need, I end up with tea poisioning... :grin:

LraiZer:
@WanWizard 
> Something vaguely related: since another recent OpenTV change, my code that stitches titles back together no longer works on all titles, any idea why (saves me diving into it :grin)?

I now use this patch on my homebuild develop image. It seems to fix your issue.
I've moved your code that stitches the OpenTV titles back together into a function and it now also stitches the EIT and Freesat epg.

[function_undoAbbreviation-20210420-2200.diff.txt](https://github.com/OpenPLi/enigma2/files/6346753/function_undoAbbreviation-20210420-2200.diff.txt)

WanWizard:
Brillant, thanks very much. Can you make a PR for it, so you'll get the attribution? 
That'll save me some commandline wizardry to get your name in...

LraiZer:
OK, will make a PR for it.
I'm sure you can add a GUI option to enable/disable it, if some ask for their split titles back : )

LraiZer:
@WanWizard 

Now that your old code is moved to a function and the patch is in, i'm unhappy about some of things that are happening.

I don't like the use of buildShortName(title) i added to get rid of the `<EM>...</EM>` in the current/next eit, as it doesn't just strip the brackets, it also truncates the title depending on where the brackets are in the title.

Also this is being done to every title, even when they are not abbreviated and that is just not right.

I will need to rework the code in the new function a little to fix a few things that should not be happening before peeps begin to complain ;)

WanWizard:
No problem, you're a lot more stilled than I am, I haven't written C in over 30 years... :smile: 

I did notice this evening that the titles on Channel 5 HD (now on +1) were still split. Not sure if that is historic data or if the join still has issues. Same on 5 USA, see

image

PR #2924, opened 2021-04-14
Add function loadImage

Taapat:
To use also for svg and jpg image load if such images are available.
And use this function in enigma cpp code instead of loadPNG.

### DIFF
--- a/lib/gdi/epng.h
+++ b/lib/gdi/epng.h
@@ -6,8 +6,9 @@
 SWIG_VOID(int) loadPNG(ePtr<gPixmap> &SWIG_OUTPUT, const char *filename, int accel = 0, int cached = 1);
 SWIG_VOID(int) loadJPG(ePtr<gPixmap> &SWIG_OUTPUT, const char *filename, int cached = 0);
 SWIG_VOID(int) loadJPG(ePtr<gPixmap> &SWIG_OUTPUT, const char *filename, ePtr<gPixmap> alpha, int cached = 0);
-SWIG_VOID(int) loadSVG(ePtr<gPixmap> &SWIG_OUTPUT, const char *filename, int cached = 0, int width = 0, int height = 0, float scale = 0);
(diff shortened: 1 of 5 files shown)


lib/gdi/epng.h
```diff
@@ -6,8 +6,9 @@
 SWIG_VOID(int) loadPNG(ePtr<gPixmap> &SWIG_OUTPUT, const char *filename, int accel = 0, int cached = 1);
 SWIG_VOID(int) loadJPG(ePtr<gPixmap> &SWIG_OUTPUT, const char *filename, int cached = 0);
 SWIG_VOID(int) loadJPG(ePtr<gPixmap> &SWIG_OUTPUT, const char *filename, ePtr<gPixmap> alpha, int cached = 0);
-SWIG_VOID(int) loadSVG(ePtr<gPixmap> &SWIG_OUTPUT, const char *filename, int cached = 0, int width = 0, int height = 0, float scale = 0);
+SWIG_VOID(int) loadSVG(ePtr<gPixmap> &SWIG_OUTPUT, const char *filename, int cached = 1, int width = 0, int height = 0, float scale = 0);
 
+int loadImage(ePtr<gPixmap> &result, const char *filename, int accel = 0, int width = 0, int height = 0);
```
SimonCapewell, line 11: Can we make this callable from python please!
`SWIG_VOID(int) loadImage(ePtr<gPixmap> &SWIG_OUTPUT, const char *filename, int accel = 0, int width = 0, int height = 0);`

Taapat:
In Python we already have LoadPixmap for this.
At first I wanted to move all the LoadPixmap logic here.
But then I concluded that I can't use getDesktop in gdi which I need for svg automatic scale.
If any exceptions are needed in LoadPixmap I decided that it doesn't make sense to change anything and leave LoadPixmap for python and create loadImage for c++.

PR #2926, opened 2021-04-16
Update po files using updateallpo-multiOS.sh

persianpros:


littlesat:
Cannot be merged (anymore) due to conflicts ?

persianpros:
Because you first merged the newer PR if you check PR numbers, I need to send this again later.

This a common mistake in PR merging because we should not merge by notifications we should take a look at PR numbers.

PR #2931, opened 2021-04-18
[Satconfig] fix wrong visual effect default value in setChoices

Dima73:


bbbuk:
I've left a comment on this [here](https://github.com/OpenPLi/enigma2/commit/7d9815802f299b5d1d875f3671e71b78fe373538) because in my limited testing, it did read the tuner setting but any future changes to that tuner (ie like scr channel changing) and a reboot resulted in it going back to 19.2 instead of 28.2 etc that I had it set as.

PR #2945, opened 2021-04-22
enigma2.pot file is needed for people that want to start 

pr2git:
a new language translation so we need to keep and update it in the repo.


persianpros:
I remember I told about this before and PLi rejected it but I don't remember why.

PR #2946, opened 2021-04-22
fr.po translation 2021-04-22

pr2git:


pr2git:
I have some doubts about this PR too many strings are deleted.

PR #2953, opened 2021-04-25
estring: undo abbreviation function

LraiZer:
* add colon ellipsis as a valid abbreviation
* better catch unusually punctuated sentences
* code optimization, use existing replace_all


WanWizard:
I think you also need to check for ":.." (a colon and two periods), which is quite common?

LraiZer:
Like:.. [this check here](https://github.com/OpenPLi/enigma2/pull/2953/files#diff-ca8aa4e29508426f0acedb6f7590e75521e7b2400b75919b6e0c3eb7062f785dR51)?

WanWizard:
Grumfff, time to open my eyes... :roll_eyes:

PR #2954, opened 2021-04-26
Update uk.po

badsystem:


WanWizard:
We don't accept PR's on release branches. 8.1-release is due soon, and will contain your develop translations.

PR #2958, opened 2021-04-27
[Satconfig] add warning text in second slot dual VU+ tuner

Dima73:
-Warning! The second slot of Vuplus DVB-S NIM(AVL6222) dual tuner does
not support Unicable/JESS type.

### DIFF
--- a/lib/python/Screens/Satconfig.py
+++ b/lib/python/Screens/Satconfig.py
@@ -397,7 +397,7 @@ class NimSetup(Screen, ConfigListScreen, ServiceStopScreen):
 			if currLnb.lof.value == "unicable":
 				warning_text = ""
 				if "Vuplus DVB-S NIM(AVL6222)" in self.nim.description and self.nim.internallyConnectableTo() is not None:
-					warning_text = _("Warning! The second slot of this dual tuner does not support this type. ")
+					warning_text = _("Warning: the second input of this dual tuner may not support SCR LNBs. ")
 				self.advancedUnicable = getConfigListEntry(self.indent % ("%s%s" % ("SCR (Unicable/JESS) ", _("type"))), currLnb.unicable, warning_text + _("Select the type of Single Cable Reception device you are using."))
 				self.list.append(self.advancedUnicable)
 				self.externallyPowered = getConfigListEntry(self.indent % _("Externally powered"), currLnb.powerinserter, _("Select whether your SCR device is externally powered."))
--- a/lib/python/Screens/Satconfig.py
+++ b/lib/python/Screens/Satconfig.py
@@ -395,7 +395,10 @@ class NimSetup(Screen, ConfigListScreen, ServiceStopScreen):
 				self.list.append(getConfigListEntry(self.indent % _("Threshold"), currLnb.threshold, _("Enter the frequency at which you LNB switches between low band and high band. For more information consult the spec sheet of your LNB.")))
 
 			if currLnb.lof.value == "unicable":
-				self.advancedUnicable = getConfigListEntry(self.indent % ("%s%s" % ("SCR (Unicable/JESS) ", _("type"))), currLnb.unicable, _("Select the type of Single Cable Reception device you are using."))
+				warning_text = ""
+				if "Vuplus DVB-S NIM(AVL6222)" in self.nim.description and self.nim.internallyConnectableTo() is not None:
+					warning_text = _("Warning! The second slot of this dual tuner does not support this type. ")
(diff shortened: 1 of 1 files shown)


littlesat:
Why give a warning and not block configuring? It should be avoided using if tuner x then do something else.... better consider going via systeminfo as feature...

Dima73:
1)I thought blocking would raise more questions than help. 
2)systeminfo = "Vuplus DVB-S NIM(AVL6222)" and /proc/stb/frontend/this slot number/rf_switch 
This is not possible and this is an exception.

eriksl:
I am not against applying this, but I don't understand exactly what's the problem. This AVL6222, is it a dual tuner with only one fysical input? And no way to loop them through? That would be strange, as that is the only way it could be working at all.

Dima73:
index dual tuner for two pysical input.
tuner 1(upper)  -->work all (Unicable/JESS, DiSEqC 1.0/1.1/1.2) and etc...
tuner 2 (lower ) -->Unicable/JESS not support -->hardware limit

eriksl:
Interesting. I am curious what is the problem, as both Unicable and JESS use a command scheme that is very close to DiSEqC. I can imagine you'd need to use an external loop through, though.

Dima73:
By the way, this tuner supports internal loop. 
https://forums.openpli.org/topic/83111-assigning-tuners-to-loop-through-from/?view=findpost&p=1338703

eriksl:
And DiSEqC works, SCR not?

eriksl:
```
...
NIM Socket 14:
	Type: DVB-S2X
	Name: Vuplus DVB-S NIM(45308X FBC)
	Frontend_Device: 14
NIM Socket 15:
	Type: DVB-S2X
	Name: Vuplus DVB-S NIM(45308X FBC)
	Frontend_Device: 15
NIM Socket 16:
	Type: DVB-S2
	Name: Vuplus DVB-S NIM(AVL6222)
	Frontend_Device: 16
	I2C_Device: 3
NIM Socket 17:
	Type: DVB-S2
	Name: Vuplus DVB-S NIM(AVL6222)
	Frontend_Device: 17
	I2C_Device: 3
```
Over here it works with JESS...

Dima73:
> 
> 
> And DiSEqC works, SCR not?

yes

Dima73:
I've seen similar restrictions on  DMM
https://github.com/opendreambox/enigma2/blob/master/usr/lib/enigma2/python/Components/NimManager.py#L1142

eriksl:
> FIXMEE no equal to for unicable .. because each tuner need a own SCR...

Their comment doesn't really help ;-). IIRC the tuner actually has two physical inputs and I connected them both. I can image it won't work with the internal loop through?

Dima73:
Unicable as SCR not work in second tuner when use physical cable .
https://www.vuplus-support.org/wbb4/index.php?thread/68777-umstellung-auf-unicable/&postID=857730#post857730

Moreover, I checked it myself.

eriksl:
Neither https://wiki.vuplus-support.org/index.php?title=Tunerkonfiguration_Vu%2B_Duo%C2%B2_mit_einem_DVB-S2_Dual_Tuner nor https://www.vuplus-support.org/wbb4/index.php?thread/68777-umstellung-auf-unicable/&postID=857730#post857730 say it is impossible. I only understand from it that you need to make an external loop through.

eriksl:
So I'd agree to add a warning saying that.

Dima73:
I spent a few hours after installing the Unicable LNB until I realized that this setting does not work in the lower tuner.

eriksl:
I think it doesn't work on an SCR LNB, but that it does work on a SRC switch (which I have).

So how about a warning, that it will probably not work on a SCR LNB, if you select SCR there?

Dima73:
Just need to change the text of the warning.
For example, probably won't work with SCR LNB/SRC switch.
How it should look right (is my English equal to a Google translator)?

Dima73:
unicableList = [("unicable_lnb", _("SCR (Unicable/JESS)") + " " + _("LNB")), ("unicable_matrix", _("SCR (Unicable/JESS)") + " " + _("Switch")), ("unicable_user", _("SCR (Unicable/JESS)") + " " + _("User defined"))]

Maybe...
Warning! The bottom input of this dual tuner may not support SCR LNB type.

eriksl:
I agree. If you change the request I'll apply it.

I propose the wording to be "Warning: the second input of this dual tuner may not support SCR LNBs" (which indeed is almost the same as you proposed).

eriksl (approved):

eriksl:
Done! Thanks dimitrij!

PR #2975, opened 2021-05-04
Movie player infobar pvr states enhancement

DimitarCC:
- moved pvrstate screen functionality to the Movieplayer infobar
- added possibility to use icons for the states


DimitarCC:
The png images probably can be changed to svg since they are more scalable and can fit in any sizes.

lib/python/Screens/InfoBar.py
```diff
@@ -189,14 +198,49 @@ def __init__(self, session, service, slist=None, lastservice=None, infobar=None)
...
 		self.infobar = infobar
 		self.lastservice = lastservice or session.nav.getCurrentlyPlayingServiceOrGroup()
+		self.onShow.append(self.__onShow)
+		self.currentSeekState = SEEK_STATE_PLAY
 		session.nav.playService(service)
 		self.cur_service = service
 		self.returning = False
 		self.onClose.append(self.__onClose)
+		self["statetext"] = Label(text="")
+		self["statetexticon"] = Label(text="")
```
nautilus7: Isn't `self["statetexticon"]` unused, or I miss something?

nautilus7:
Great idea! Using icons is much better than having some symbols for displaying pvr states.

lib/python/Screens/InfoBar.py
DimitarCC: Indeed....there is a typo and is used State text widget instead of statetexticon. I fixed this and now it is correct on line 225 self["statetexticon"].setText(statetext)

lib/python/Screens/InfoBar.py
```diff
@@ -190,6 +189,7 @@ def __init__(self, session, service, slist=None, lastservice=None, infobar=None)
 		self.infobar = infobar
 		self.lastservice = lastservice or session.nav.getCurrentlyPlayingServiceOrGroup()
 		session.nav.playService(service)
+		InfoBarPVRState.__init__(self)
```
DimitarCC, line 193: InfoBarPVRState must be initialized here not earlier since otherwise Pixmap widgets throw errors!

nautilus7: @DimitarCC, so if you leave lines 151-192 as they were, you get errors in pixmaps? Why?

DimitarCC: Well i am not sure what is the cause....however when it tries to do setPixmap complains that pixmap instance is None. I have no idea why.

DimitarCC:
Well i changed things again. Now the skinners have to add 2 new widgets in PVRState screen or TimeshiftState screen so to be able to display icons for the states. Also the state icons are defined in a MultiPixmap and have to be added in specified order.
Since i kept the PVRState screen if skinner want to show the icons or texts in the Infobar they will have to position and overlay the PVRState screen over the infobar. That is the only option if we want to keep backward compatibility.

lib/python/Screens/InfoBar.py
```diff
@@ -543,4 +543,4 @@ def displayPlayedName(self, ref, index, n):
 		Notifications.AddPopup(text=_("%s/%s: %s") % (index, n, self.ref2HumanName(ref)), type=MessageBox.TYPE_INFO, timeout=5)
 
 	def ref2HumanName(self, ref):
-		return enigma.eServiceCenter.getInstance().info(ref).getName(ref)
+		return enigma.eServiceCenter.getInstance().info(ref).getName(ref)
```
nautilus7, line 547: don't delete the new line here

lib/python/Screens/InfoBarGenerics.py
```diff
@@ -3629,4 +3648,4 @@ def HDMIIn(self):
 			elif isStandardInfoBar(self):
 				self.session.nav.playService(slist.servicelist.getCurrent())
 			else:
-				self.session.nav.playService(self.cur_service)
+				self.session.nav.playService(self.cur_service)
```
nautilus7, line 3667: don't delete the new line here

lib/python/Screens/InfoBar.py
DimitarCC: Yes i re-added it with other fixes. seems i forgot to merge the recent changes first.

lib/python/Screens/InfoBarGenerics.py
DimitarCC: Yes same as above. I re-added it ;)

lib/python/Screens/InfoBarGenerics.py
```diff
@@ -1748,25 +1749,59 @@ def seekNextMark(self):
 
 class InfoBarPVRState:
 	def __init__(self, screen=PVRState, force_show=False):
+		self.seekstate = self.SEEK_STATE_PLAY
```
Huevos, line 1752: This PR is a reinvention of the wheel with new variable names, image names, etc, which is not a great idea for code sharing between distros.

https://github.com/OpenViX/enigma2/blob/Dev/lib/python/Screens/InfoBarGenerics.py#L2514-L2515
https://github.com/OpenViX/skins/blob/master/src/skins/1080/Simple_Ten_Eighty/skin_templates.xml#L27-L28

DimitarCC: Well i can not talk about other distros. Nor i am intending to provide solution capable of sharing between them.
If that is approved we can polish it. About the images there is no change in names because you can specify the images in the stateicon widget freely so you can use any names you want.

Huevos: That is what we do. We share code and skins between distros. i.e. PLi has contributors from both PLi team and the teams of other distros. When people reinvent the wheel it breaks the possibility to share code and makes it less likely for other teams to push back interesting code to PLi.

DimitarCC: The question then is why this code is not in OpenPLi if you share interesting stuff between distros? You could propose it before but it appears not happened untill now. Otherwise i understand your point.

Huevos: I push most of my code back to PLi. It makes code sharing easier in the future. 

In this case you should at least use the established variable names and image names otherwise it makes things very difficult for people that write skins for sharing between distros.

littlesat:
Should I merge this one?

Huevos:
No, because it uses completely different variable names to other images so it breaks skin sharing.

littlesat:
Did you offer and we accept a merge request that covers the features?

WanWizard:
Can we make a decision on either fixing this (i.e. add backward compatibility) or revert it, as it breaks every plugin that uses the PVR Infobar, like MediaPortal. I don't want this code in 8.1-release like this.

WanWizard:
We've decided to revert this one.

Although it is clearly an improvement, it is implemented in such way that it breaks backward compatibility too much.

Feel free to offer a refactored solution that doesn't break plugins using the PVR infobar...

DimitarCC:
@WanWizard Just for info what is breaking? And did you determine what is the cause?

WanWizard:
The most important one seems to be MediaPortal, we received quite a few complaints about that.

I haven't looked into it, my Python skills aren't all that, but I've just installed it on last weeks develop, and it crashes on:
````
  File "/usr/lib/enigma2/python/Navigation.py", line 75, in dispatchEvent
  File "/usr/lib/enigma2/python/Components/ServiceEventTracker.py", line 82, in event
  File "/usr/lib/enigma2/python/Screens/InfoBarGenerics.py", line 1485, in __serviceStarted
  File "/usr/lib/enigma2/python/Screens/InfoBarGenerics.py", line 1528, in setSeekState
  File "/usr/lib/enigma2/python/Screens/InfoBarGenerics.py", line 1784, in __playStateChanged
KeyError: 'statespeed'
[ePyObject] (CallObject(<bound method Navigation.dispatchEvent of <Navigation.Navigation instance at 0xb02b9508>>,(0,)) failed)

````

Huevos:
And should be using the same variable names as all the other images.

PR #2985, opened 2021-05-16
el.po update 2021.05.19

MCelliotG:
el.po update 2021.05.19

### DIFF
--- a/po/el.po
+++ b/po/el.po
@@ -93,7 +93,7 @@ msgid " (lower than any auto)"
 msgstr " (χαμηλότερα όλων των αυτόματων)"
 
 msgid " Warning: the selected tuner should not use SCR Unicable type for LNBs because each tuner need a own SCR number."
-msgstr ""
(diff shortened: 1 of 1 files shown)


po/el.po
```diff
@@ -6020,7 +6020,7 @@ msgstr ""
 "Όταν είστε έτοιμοι, πιέστε ΟΚ για να συνεχίσετε."
 
 msgid "Please configure the fallback tuner setup?"
-msgstr "Ρυθμίστε το εφεδρικό tuner."
+msgstr "Ρυθμίστε το εφεδρικό tuner;"
```
nautilus7, line 6023: @MCelliotG despite the question mark is on par with the original text, it is just wrong (both the English and the Greek wording).

WanWizard:
I'll fix that manually.

po/el.po
MCelliotG: You are absolutely right! What happened is that the auto translation in poedit does not understand the differences between languages in punctuation. So that was "fixed" automatically and I missed it. Of course there should not be any quotation mark, as there wasn't until now.

MCelliotG:
> 
> 
> I'll fix that manually.

Thank you! These are really stupid mistakes that happen by human error between thousands of lines. It's a good thing such errors are discovered eventually, so that all translations will benefit.

PR #2999, opened 2021-06-02
[epgcache.cpp] Fix UTF-8 double encoding bug

Huevos:
This bug was causing some EPG and AutoTimer searches to fail due to an encoding mismatch.

Patch author: LraiZer.


WanWizard:
Thanks, I've very short on available time atm.

PR #3004, opened 2021-06-05
all po update release-8.1 + fr.po translation

pr2git:


WanWizard:
Not needed, it isn't released yet, so it can still be rebased with develop.

PR #3005, opened 2021-06-06
update ru.po

Dima73:


WanWizard:
Your update crossed the po update, can you recreate it?

PR #3007, opened 2021-06-06
[PositionerSetup] add option "Checking ONID/TSID"

Dima73:
-"Checking ONID/TSID" use only manual.
Enter valid onid/tsid and run check.
-add option "Stop live TV service"
-update last sat rotor position
-small cosmetics code

### DIFF
--- a/lib/python/Plugins/SystemPlugins/PositionerSetup/plugin.py
+++ b/lib/python/Plugins/SystemPlugins/PositionerSetup/plugin.py
@@ -16,7 +16,7 @@ from Components.ActionMap import NumberActionMap, ActionMap
 from Components.NimManager import nimmanager
 from Components.MenuList import MenuList
 from Components.ScrollLabel import ScrollLabel
-from Components.config import config, ConfigSatlist, ConfigNothing, ConfigSelection, ConfigSubsection, ConfigInteger, ConfigFloat, KEY_LEFT, KEY_RIGHT, KEY_0, getConfigListEntry
+from Components.config import config, ConfigSatlist, ConfigNothing, ConfigSelection, ConfigSubsection, ConfigInteger, ConfigFloat, KEY_LEFT, KEY_RIGHT, KEY_0, getConfigListEntry, NoSave
 from Components.TuneTest import Tuner
 from Components.Pixmap import Pixmap
 from Tools.Transponder import ConvertToHumanReadable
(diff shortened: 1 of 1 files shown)


lib/python/Plugins/SystemPlugins/PositionerSetup/plugin.py
```diff
@@ -778,17 +786,48 @@ def tune(self, transponder):
...
+			self.onid = tsidonid[0]
+			self.tsid = tsidonid[1]
+			if self.frontend and self.isLocked() and not self.isMoving and hasattr(self, "raw_channel") and self.raw_channel:
+				self.checkingTsidOnid = True
+				self.raw_channel.receivedTsidOnid.get().append(self.gotTsidOnid)
+				self.raw_channel.requestTsidOnid()
+
+	def gotTsidOnid(self, tsid, onid):
+		if tsid == self.tsid and onid == self.onid:
+			msg = "\c0000??00" + _("This valid ONID/TSID")
```
nautilus7: Why do you keep hardcoding colors in the code?? Do you understand that skins cannot change these colors? This is unacceptable!

Dima73: `This is unacceptable!`
Why?
Green color ok.
Red color not ok.


What else does the user just need?
I think in 10 seconds of displaying the message, the user will tolerate these colors.

betacentauri: Eg some suffer red green weakness and can’t see the difference.
But I don’t think nautilus7 means this.

nautilus7: @Dima73 the purpose of skins is to define the colors THEY want (among other things). You do not allow them to do so with your coding....

A few years ago, I tried to clear all hard coded colors, but I see them coming back again. And you did it one more time recently with another commit (that I can't find right now).

nautilus7: If you think that adding colors gives some value to your code, then this is how I propose to do it https://github.com/OpenPLi/enigma2/commit/17f344a92cb05354e26179227b6ee49703d3c7e9

Dima73: Let's apply this patch.
Will you create a new one with custom colors later? 
Ok?

nautilus7: I am not in a mood of fixing other people's incomplete/scrappy code....

Btw, what does this patch do? Your description is not detailed at all.

PR #3018, opened 2021-06-19
[OSDPositionSetup] use standard save instead config.save()

Dima73:
If several windows are open at the same time, then only the settings of
the last window are saved.


littlesat:
Don’t you save the config file 4 times instead of one time? So a simple one time save just saves all? Or are there additional tricks in config to avoid this?

Dima73:
https://github.com/OpenPLi/enigma2/pull/3020

PR #3019, opened 2021-06-20
Fix the mess (Emergency)

persianpros:
@WanWizard 

This commit should be reverted and applied as a patch against the repo: https://github.com/OpenPLi/enigma2/commit/ad3d13eebef882b01494ce7d6e89628ff64b8e60

It's only the patch file.

P.S. https://github.com/OpenPLi/enigma2/pull/3016 is safe to merge aslo, it's just a PEP8 cleanup which we do automatically.


WanWizard:
I want to stay out of the PEP8 discussion if you don't mind, I have enough on my plate. 
I personally am all for code standards (no matter which one, but pick one and stick to it :grin: ).

So, this corrects the revert mess 100%?

persianpros:
@WanWizard 

Yes it's the fix for e2 repo, PEP8 is accepted among all teams even PLi (see the history of those files). It's just a monthly commit to keep it up-to-date with PEP8.

PR #3021, opened 2021-06-29
Support "itemHeight" in String Lists

Dima73:
original code: madie
original commit like: OpenVisionE2/enigma2-openvision@dca1e31

### DIFF
--- a/lib/gui/elistboxcontent.cpp
+++ b/lib/gui/elistboxcontent.cpp
@@ -275,6 +275,13 @@ void eListboxPythonStringContent::setList(ePyObject list)
 		m_listbox->entryReset(false);
 }
 
+void eListboxPythonStringContent::setItemHeight(int height)
+{
+	m_itemheight = height;
+	if (m_listbox)
+		m_listbox->setItemHeight(height);
+}
+
 PyObject *eListboxPythonStringContent::getCurrentSelection()
 {
 	if (!(m_list && cursorValid()))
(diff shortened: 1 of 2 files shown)


IanSav:
Is it possible to extend the functionality offered by this pull request and also provide a method that allows flags to be passed into the eListboxPythonStringContent object?  I would love to be able to use flags like:
- #define RT_HALIGN_LEFT 1
- #define RT_HALIGN_RIGHT 2
- #define RT_HALIGN_CENTER 4
- #define RT_HALIGN_BLOCK 8
- #define RT_VALIGN_TOP 0
- #define RT_VALIGN_CENTER 16
- #define RT_VALIGN_BOTTOM 32

to allow code more control over how text is displayed in the control.  The method could be called setFlags() and take a value that is the appropriate logical or combination of the bits represented by the above values.

persianpros:
@Dima73 

Always try to cherry-pick to keep the original commit like: https://github.com/OpenVisionE2/enigma2-openvision/commit/dca1e31cbc2feb23671e5c7dcda11ea0792b6030

OV is similar to PLi so cherry-picking is possible ;)

Dima73:
>  The method could be called setFlags() and take a value that is the appropriate logical or combination of the bits represented by the above values.

To do this, you need to change a lot in the code.
I hardly understand this.

68foxboris:
My image does not start, check someone else.
`
18:43:03.0923   File "/usr/lib/enigma2/python/Tools/LoadPixmap.py", line 10, in LoadPixmap
18:43:03.0928 TypeError: in method 'loadPNG', argument 2 of type 'char const *'
`

PR #3026, opened 2021-07-13
all po update 2021.07.13 + fr.po latest

pr2git:


Dima73:
Why need remove it and etc?
https://github.com/OpenPLi/enigma2/blob/9c459b4b5e8ca02af4978070b5c8a632e5e30b1f/lib/python/Plugins/SystemPlugins/NetworkWizard/networkwizard.xml#L86

pr2git:
I will check again because normally this should only update the .po files and don't touch to anything else so I close this one.

PR #3031, opened 2021-07-20
updated encoding.conf

ims21:


WanWizard:
8.0 will not be built anymore, I'll merge it into 8.1 later.

PR #3037, opened 2021-07-29
[Satconfig] when disabled tuner root(FBC socket 1) -->automatically d…

Dima73:
…isable all virtual tuners for this FBC DVB-C tuner

more info:

https://forums.openpli.org/topic/84734-vu-duo-4k-wrong-tuner-handling-when-adding-a-usb-tuner


WanWizard:
I've checked with the Gigablue DVB-C FBC tuner, it also has one input and one loop-through.

PR #3045, opened 2021-07-31
[EventView] Use "info" button to close screen

nautilus7:
While the "EventView" screen is displayed by pressing the "info" button on the remote control while watching a channel, you cannot use the same button to close it. This is not convenient and rather unusual, since most screens can be closed with the same button they are opened with (e.g. menus, infobar, etc).

This change will reassign the "info" button to close (cancel) operation for the "EventView" screen. Prior to this, the "info button and the "menu" button were used for the same purpose: To open a menu with specific plugins (if any of these plugins are actually installed in the system - I don't have any, so I don't know which they are) when the "EventView" screen in on.

The "EventViewActions" actions that are modified here, are not used anywhere else in enigma2, so I don't expect any side effects.

### DIFF
--- a/data/keymap.xml
+++ b/data/keymap.xml
@@ -647,7 +647,7 @@
 		<key id="KEY_RIGHT" mapto="nextEvent" flags="mr" />
 		<key id="KEY_RED" mapto="openSimilarList" flags="m" />
 		<key id="KEY_GREEN" mapto="timerAdd" flags="m" />
-		<key id="KEY_INFO" mapto="contextMenu" flags="m" />
+		<key id="KEY_INFO" mapto="cancel" flags="m" />
 		<key id="KEY_MENU" mapto="contextMenu" flags="m" />
 		<key id="KEY_PROGRAM" mapto="timerAdd" flags="m" />
 		<key id="KEY_EDIT" mapto="timerAdd" flags="m" />


Huevos:
Here is a list of a few plugins that do use `WHERE_EVENTINFO`.

- advancedmovieselection
- autotimer
- birthdayreminder
- epgrefresh
- epgsearch
- imdb
- merlinepg
- ofdb
- partnerbox
- pluginhider
- pluginsort
- seriesplugin
- EPGTranslator
- tmdb
- TVSpielfilm

Dima73:
It is a bad idea.
Many users are used to this functionality.

Ev0-BH:
Not sure why anyone would be pressing the same button to enable something as they would to disable it tbh.

Surely it makes much more sense to just press exit or back to go back to where you were before.

nautilus7:
@Huevos thanks for list. I don't use any of these...

@Dima73 @Ev0-BH most TVs do that. They display the program info by pressing the "info" button, and they close it as well with same key. The same happens in enigma2 with the infobar. You display the infobar by pressing the "ok" button and you close it with "ok" as well. Not sure what's the problem... Why do you have to move your finger to another button???

I don't really care if users are used to something... In my opinion, having 2 keys do the same thing is wrong design anyway...

Ev0-BH:
I don't really care if users are used to something... In my opinion, having 2 keys do the same thing is wrong design anyway...

Not a good way to look at things imo.

RobvanderDoes:
"OK" on InfoBar opens SecondInfoBar.

Ev0-BH:
> I don't really care if users are used to something... In my opinion, having 2 keys do the same thing is wrong design anyway...

Not a good way to look at things imo.

nautilus7:
> "OK" on InfoBar opens SecondInfoBar.

Oh come on!!! And "ok" again close it!

RobvanderDoes:
Yes, meaning the second "OK" opens a completely different functionality.

nautilus7:
> > I don't really care if users are used to something... In my opinion, having 2 keys do the same thing is wrong design anyway...
> 
> Not a good way to look at things imo.

My point is, that if something does not feel right, it should be changed, no matter if users are used to it.


If the majority of you like the way the buttons are configured right now, that is "menu" opens a menu and "info" opens the same menu (!!!), I am fine with that!

nautilus7:
> Yes, meaning the second "OK" opens a completely different functionality.

No it's not different! It's infobar, second infobar (i.e. an infobar with more information), and then close!

WanWizard:
There are two different discussions going on here:
* close with the same button as open
* multiple buttons that do the same thing

Please don't mix them up.

If I look at TV's, they tend to use toggle buttons for functions starting from watching TV. This allows you to quickly pop up a screen and remove it again so you don't loose extra "screen time". I don't have any particular problem with this (consistency is good), even though I am used to info-info to get to the event menu. But if this happens, it should happen for all functions, not just for info, so it should be a stuctured effort, not just a change for a single button.

As for multiple buttons for the same function, I tend to agree with that, but also here, there are a lot of screens where this is the case, just think about the number of times the red button is used as close or exit.

Finally, I agree that for (end)users, this can be a massive change, so I'd rather not implement it in the middle of a release cycle, but keep it for OpenPLi 9 which is more logical for such a major change.

nautilus7:
@WanWizard thanks for your comprehensive opinion and time spent writing it.

pr2git:
Personally I am in favor of this change. Here is my personnal situation I purchase brand new Sony TV and the VU+ is know and I can control the VU+ from the Sony RCU.  Problem is that Sony doesn't send the back button to the STB it seems to be reserved for  Sony TV.
So when I entered the EPG (= Info button) I cannot leave it.  Pressing several time the EPG/INFO button I stuck in the last screen.
So would be nice that when reaching the very last screen, pressing EPG/INFO button again simply leave it and return to TV screen.
If possible this behavior should became the default one, to avoid to get stuck in the very last screen.

Ev0-BH:
> 
> 
> Personally I am in favor of this change. Here is my personnal situation I purchase brand new Sony TV and the VU+ is know and I can control the VU+ from the Sony RCU. Problem is that Sony doesn't send the back button to the STB it seems to be reserved for Sony TV.
> So when I entered the EPG (= Info button) I cannot leave it. Pressing several time the EPG/INFO button I stuck in the last screen.
> So would be nice that when reaching the very last screen, pressing EPG/INFO button again simply leave it and return to TV screen.
> If possible this behavior should became the default one, to avoid to get stuck in the very last screen.

So just because your tv doesn't support the back button, everyone should be forced to change they way they use there stb ?

Maybe it would be better to add it as an option in the epg settings, so the user can decide if they want to enable the same button to close the function ? that way nobody has a change forced upon them, but those that would like this function can enable it.

pr2git:
> > Personally I am in favor of this change. Here is my personnal situation I purchase brand new Sony TV and the VU+ is know and I can control the VU+ from the Sony RCU. Problem is that Sony doesn't send the back button to the STB it seems to be reserved for Sony TV.
> > So when I entered the EPG (= Info button) I cannot leave it. Pressing several time the EPG/INFO button I stuck in the last screen.
> > So would be nice that when reaching the very last screen, pressing EPG/INFO button again simply leave it and return to TV screen.
> > If possible this behavior should became the default one, to avoid to get stuck in the very last screen.
> 
> So just because your tv doesn't support the back button, everyone should be forced to change they way they use there stb ?
> 
> Maybe it would be better to add it as an option in the epg settings, so the user can decide if they want to enable the same button to close the function ? that way nobody has a change forced upon them, but those that would like this function can enable it.

This won't change anything for people, just that they will no longer be stucked on the latest possible screen.

Anyway I just test the exit button on the Sony RCU so now I can leave the EPG screen with it too.

littlesat:
We should make a compromise here… when there is no plug-in then exit, when there is a plug-in like IMDb show it (and when there is just one plug-in show it immediately). Need extra python code and rename button task to epg plug-in or so.

nautilus7:
I am not sure I like this diversity. I think it will be confusing. Keep in mind that there is also "menu" button that does the same. Is it going to change as well to match the functionality of the "info" button?

So, better keep them like they are today, since people are used to them.

littlesat:
It at least keeps functionality as it was… and when there was no plug-in it becomes exit… where in the paar nothing happened… and when there is only one plug-in behind it you do not get the choicebox anymore (why should you see it?)

PR #3052, opened 2021-08-04
Update uk.po

badsystem:


WanWizard:
Thanks, but 8.0 isn't being build anymore.

PR #3055, opened 2021-08-06
lt.po thanks adga + fr.po + all macedonia

pr2git:


po/el.po
```diff
@@ -5642,6 +5639,9 @@ msgstr "Κανονική λειτουργία"
 msgid "North"
 msgstr "Βόρεια"
 
+msgid "North Macedonia (The Republic of)"
+msgstr "Δημοκρατία των Σκοπίων"
+
```
nautilus7, line 5644: I don't know where you got this, but they should be called "Βόρεια Μακεδονία" in Greek, since 2020.

pr2git:
This PR create this:
https://github.com/OpenPLi/enigma2/pull/3054

pr2git:
The problem is that the generation of the po file doesn't tag it as to be reviewed (and I don't know why).
I saw this PR so I change the translation in French but other translators won't see it clearly.

nautilus7:
that's because the new string was added by hand in the place of the old one in each po file. That was done in https://github.com/OpenPLi/enigma2/commit/b13e51bdfab5f441b7bd25cb6d22dd4114a63b52

WanWizard:
Sorry about that, I now get what you meant. I need a holiday... :tired_face:

nautilus7:
We all need.

pr2git:
I have create a PR that clean up all the translation for North Macedonia:

https://github.com/OpenPLi/enigma2/pull/3056

po/el.po
WanWizard: Committed translations for all supported languages. Issue should be fixed now.

PR #3056, opened 2021-08-06
Fix wrong North Macedonia translation

pr2git:


pr2git:
Solved by WanWizard so no longer needed.

PR #3058, opened 2021-08-09
[Satconfig/ServiceStopScreen] reopen tuner after save

Dima73:
*[ServiceStopScreen] fix stop alternative service, add slot_number, add
oldAlternativeref
*[Satconfig] when tuner DVB-T/C or ATCS mode and disable tuner and the
previous playing service was of the same type close/reopen frontend for
clear all setting


AbuBaniaz:
I think it is a good idea stopping services when a tuner has been disabled and saved instead of waiting  for a restart or changing tuner/active service.

PR #3067, opened 2021-08-16
[ClockToText] Use named arguments in strings

nautilus7:
This eliminates the following warning:

```
'msgid' format string with unnamed arguments cannot be properly localized:
The translator cannot reorder the arguments.
Please consider using a format string with named arguments,
and a mapping instead of a tuple for the arguments.
```

Existing translations DO NOT need any modification by hand, because ALL of them use the original string (untranslated or just copied and pasted).

I don't really see the point in offering these two strings for translation anyway, but since they're already there, I would prefer not seeing the warning when running the updateallpo.sh script.

### DIFF
--- a/lib/python/Components/Converter/ClockToText.py
+++ b/lib/python/Components/Converter/ClockToText.py
@@ -95,10 +95,10 @@ class ClockToText(Converter, object):
 
 		if self.type == self.WITH_SECONDS:
 			# TRANSLATORS: full time representation hour:minute:seconds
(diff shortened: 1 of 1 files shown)


nautilus7:
We get several warnings like this when we run the updateallpo.sh script. Are we interested in eliminating these? If we do, I can provide patches for most of them, while making sure existing translations are kept/fixed when necessary.

WanWizard:
Sounds like a plan, yes. 

What is the point of making them translatable if you can't reorder (altough for a time it is indeed a bit pointless).

PR #3088, opened 2021-08-25
[RecordTimer] create directory  /media/hdd/movie if it doesn't exist

Dima73:
Attention!
Let's discuss this proposal first.

### DIFF
--- a/lib/python/RecordTimer.py
+++ b/lib/python/RecordTimer.py
@@ -64,15 +64,20 @@ def findSafeRecordPath(dirname):
 	from Components import Harddisk
 	dirname = os.path.realpath(dirname)
 	mountpoint = Harddisk.findMountPoint(dirname)
-	if mountpoint in ('/', '/media'):
-		print '[RecordTimer] media is not mounted:', dirname
+	if mountpoint in ("/", "/media"):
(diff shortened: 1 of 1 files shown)


WanWizard:
What is the reasoning for this, as this has been removed a few years ago.

I have my NAS mounted on /media/hdd, and record in the root of that mount, I don't want nor need a "movie" directory.

Dima73:
> 
> 
> What is the reasoning for this, as this has been removed a few years ago.
> 
> I have my NAS mounted on /media/hdd, and record in the root of that mount, I don't want nor need a "movie" directory.

I recently installed a new disk and the movie folder had to be created manually.
Even if you do not use this folder, how does it get in the way?

WanWizard:
If the directory is there, it will be used automatically. So that means a directory I don't use is created, AND I need to manually change the default location for recordings to /media/hdd, otherwise they get into the wrong place.

Another option: create it only if /media/hdd is a mointpoint for a local block device, AND it is empty ?

Dima73:
> 
> 

> 
> Another option: create it only if /media/hdd is a mointpoint for a local block device, AND it is empty ?

I don't know how to check it.

PR #3091, opened 2021-08-25
[HdmiCecSetupScreen] Clean-up and optimization

nautilus7:
* Put all imports on top and sort them alphabetically.
* Allow for individual skinning, while keeping existing fallback to "Setup" skin.
* Use parent class functions for saving and closing. Less code plus the user is now asked to exit without saving changes.
* Use parent class mechanisms for updating the config list and description. Again, less code written.
* Renamed green button text from "OK" to "Save" to be consistent with other setup screens.
* Other small cleanups and enhancements.

### DIFF
--- a/lib/python/Plugins/SystemPlugins/HdmiCEC/plugin.py
+++ b/lib/python/Plugins/SystemPlugins/HdmiCEC/plugin.py
@@ -1,122 +1,94 @@
-from Screens.Screen import Screen
+import Components.HdmiCec
+from Components.ActionMap import ActionMap
 from Components.ConfigList import ConfigListScreen
-from Components.config import config, getConfigListEntry
+from Components.config import config, ConfigBoolean, ConfigSelection, getConfigListEntry
 from Components.Label import Label
 from Components.Sources.StaticText import StaticText
+from Plugins.Plugin import PluginDescriptor
+from Screens.LocationBox import LocationBox
+from Screens.Screen import Screen
+from os import path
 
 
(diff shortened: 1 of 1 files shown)


nautilus7:
Everything seem to work as expected regarding my changes, but @littlesat @IanSav @Huevos @WanWizard @athoik can have a look just to be sure...

Another change I am thinking of is to convert this plugin to a screen. hdmi-cec is considered standard functionality today, so there's no point to keep it as a plugin. Should I offer a patch for this as well?

WanWizard:
Sounds like a plan, yes!

Huevos:
@nautilus7, have a look at OpenViX version. Much simpler. 60 lines of code not 125.

https://github.com/OpenViX/enigma2/blob/Dev/lib/python/Plugins/SystemPlugins/HdmiCEC/plugin.py

nautilus7:
@Huevos thanks for the heads up! I can't do it this way in openpli, because your "Setup" class is on steroids to read xml files from plugins, plus it allows conditional creation of config entries. There are about 50 commits from @IanSav and you in this direction during the past year, which openpli unfortunately lacks. These changes in the "Setup" class can help a lot to simplify the code of many setup-like screens around enigma2. I don't know why they are not in openpli too.

nautilus7:
> 60 lines of code not 125.

I dropped from 160 to 125 and could go to 110 if converted to screen instead of a plugin, without changing anything else in enigma2. But yes, further reduction in this and many other similar classes is possible with your distro.

PR #3096, opened 2021-08-27
EPG description - replacing newlines (optionaly), separator in fullescription (optionaly)

ims21:


ims21:
It could be made with 2 lines in cpp + several lines in python, but - unfortunately - eConfigManager does not work out of main loop.

PR #3101, opened 2021-08-30
[RecordTimer/ServiceList] add  timer when need zap in history  list

Dima73:


WanWizard:
Why is `adjust` tested and passed as a number, while the prototype defines it as a boolean with value `True` ?

Dima73:
When timer start and need zap to service 'adjust' set 1 and this means that the ServiceList must save this channel. 
I checked, everything works.

WanWizard:
I understand it works due to Python's loose typing, but it still looks odd that the variable is defined as a boolean, but never used as a boolean. Should it then not be `adjust = 0` ?

Dima73:
`Should it then not be adjust = 0 ?`
Not.
Default adjust is True same as 1 when check  boolean.
Here's a similar implementation example where everything works great:
https://github.com/OpenPLi/enigma2/commit/adf6ac4a560b290076b1676e1b1f432113f0e88f#diff-2e243a8b8596fb268da9df3b3eefdf0472809bbd3d04dcf5d5a4fa34abfd452eR113

WanWizard:
I get that it works.

I just think is bad programming which can lead to errors. It is either a boolean or an integer, but not both.

Dima73:
This is a feature of the Python code s that you can and should use for your own purposes, as in this case.

nautilus7:
> This is a feature of the Python code s that you can and should use for your own purposes, as in this case.

You use this variable like an boolean and you test its value like it's a boolean, you even set its default value to 'true', but you still insist that it's good practice to set it equal to 1. Just because Python allows it, doesn't mean it's should be done. @WanWizard made it pretty clear...

Dima73:
No and no again.
Look carefully where in the code this variable is used. 
	def playService(self, ref, checkParentalControl=True, forceRestart=False, adjust=True):
		session = None
		startPlayingServiceOrGroup = None
		count = isinstance(adjust, list) and len(adjust) or 0
		if count > 1 and adjust[0] is 0:
			session = adjust[1]
			if count == 3:
				startPlayingServiceOrGroup = adjust[2]
			adjust = adjust[0]
This variable can be used in different ways, for example:
adjust ="bla-bla"
or 
adjust =["test",]
if  adjust --> answer True
adjust =""
or
adjust =[]
if  adjust --> answer False

I don't want to break many plugins that use def playService and add a new variable to def playService.

nautilus7:
```
	def setCurrent(self, ref, adjust=True):
		force_save = False
		if self.l.setCurrent(ref):
			return None
			if adjust is 1:
				force_save = True
			else:
```

Are you kidding us??????

Dima73:
> Are you kidding us??????

Explain what you are talking about?

nautilus7:
When you set "adjust=True" as the default value in the method definition, I expect this variable to be _used_ as a boolean... so when 4 lines later you test if "adjust is 1" is a bit confusing!!! Why is it so difficult to get it?

We all know that python is a weakly typed language and you can change the type of the variables, which allows you to do clever things, but in this case there is _no reason_ to do that! You can just test if "adjust is True" which is consistent with what you "told" us a few lines before when you defined this variable as a boolean.

I don't know how to explain it in more detail. These stuff are for beginners... Let's see what other coders will say on this to try to convince you.

I already spent much time writing theses, for a matter that is not so important. As you said this style of coding is perfectly fine in Python, but it's not accepted by me, especially on an open source project that other people will read my code... Unfortunately, some people, thankfully quite a few, are trying to improve the enigma2 code in every aspect, while others don't care at all about code quality....

WanWizard:
I am a n00b when it comes to Python, but have been programming for over 40 years, the last 20-odd years in PHP, which is also a loosly typed language.

These things eventually go horribly wrong, either because the language evolves (PHP now supports strict typing, and old code written like this now fails all over the place), of because someone sees the integer values, misses the fact it is actually used as a boolean switch, starts introducing a zero value, with the result that zero and False may constitute two different states, and all hell breaks loose...

littlesat:
Nautilus is right…. I know python make this possible…. But it is not done…

PR #3112, opened 2021-09-07
Update uk.po

badsystem:


po/uk.po
```diff
@@ -10713,7 +10711,7 @@ msgstr "Ви не обрали канал для запису."
 
 #, python-format
 msgid "You have to wait %(min)d minutes, %(sec)d seconds!"
-msgstr ""
+msgstr "Вам потрібно почекати %(хв)d хвилин, %(сек)d секунд!"
```
nautilus7, line 10714: @badsystem the "%(min)d" and "%(sec)d" are **variables**. The difference from "%d" is that they are named, so you can re-order them as you like (i.e. put them in any order) in your translation.

Please do not translate them, keep them as in the original string!!!

nautilus7 (changes requested):

PR #3115, opened 2021-09-08
Update About.py

badsystem:


WanWizard:
Why is this needed? These strings are non-translatable?

nautilus7:
Yep, "MHz" and "°C" are units. They should not be translated!

badsystem:
"GB" - "ГБ"
"MHz" - "МГц"
these values in the Ukrainian, Russian language have a translation

nautilus7:
Well, as a physicist (I have a BSc in Physics), I will tell you that not using the English/[International units](https://en.wikipedia.org/wiki/International_System_of_Units) for physical quantities is not the best thing to do. But if in your language you use a unit in a translated form, then I would say I am ok.

So, how do you say "Giga Bytes" in Ukrainian for example? I mean do you ACTUALLY have Ukrainian words for Giga Bytes, or you just substitute the "GB" with the corresponding Cyrillic letters "ГБ" and you call this "translation"? How do you say Giga Bytes, Celsius, Mega Hertz, etc in Ukrainian?

For example in Greek with translate these:
meter (m) (this is a Greek word anyway) --> μέτρο (μ)
square meter (m²) --> τετραγωνικό μέτρο (τμ)
cubic meter (m³) --> κυβικό μέτρο (κμ)
hour (h) --> ώρα (no abbreviation in Greek)
second (s) --> δευτερόλεπτο (no abbreviation in Greek)
kilogram (kg) (this is a Greek word anyway) --> χιλιόγραμμο (χλγ)

The above pretty much cover the base quantities. For all others we use the International units (names and symbols) exclusively.

As a conclusion, if you would like to translate the units, just offer for translation only them, not the complete string. For example, do it like this: "%s %s + _("MHz") + (%s)". That's for these specific strings, that have no other words to translate.

If @WanWizard agrees, of course...

WanWizard:
It is weird, but there are other languages that do the same. In French for example, a "Byte" is an "Octet", so they talk about "200 Go" instead of "Gb"...

badsystem:
I agree that the translation "% s% s + _ (" MHz ") + (% s)" will be more correct. Delete this Pull requests and I will make a new translation. Thanks!

nautilus7:
Well, in Greek for "bit" we have the word "δυφίο", but 9 out of 10 people will not understand what you're talking about if you use that word. The phrase "δυαδικό ψηφίο" which means "binary digit" is what we use to explain what a "bit" is and also used as a translation more often. On the other hand, we do not have a word for "byte" at all... Go figure.

For temperature, we translate "Celsius" to "Κέλσιος", who is a person after all, so that's to be expected... The problem is that if we use the initial "Κ" from "Κέλσιος" as a unit and write "20°K" instead of "20°C", there will be a misunderstanding with degrees "Kelvin" which is "Κέλβιν" in Greek (starts with "K" as well).

So, that's why I believe using the international units is the best way to go.
Of course, in scientific articles, essays, school books, etc. written in Greek, using anything else than the international units (with the exception of the base quantities I mentioned before) is just wrong, no discussion about that.

badsystem:
Then I do not see the logic in translating such data
#, python-format
msgid "%3.01f dB"
msgstr "%3.01f дБ"

#, python-format
msgid "%d GB"
msgstr "%d ГБ"

#, python-format
msgid "%d MB"
msgstr "%d МБ"

msgid "0"
msgstr "0"

msgid "1"
msgstr "1"

msgid "1.0"
msgstr "1.0"

msgid "1.1"
msgstr "1.1"

msgid "1.2"
msgstr "1.2"

nautilus7:
I am in the process of removing some of them.

The diseqC version numbers "1.0", "1.1", "1.2" are already removed from the translations. The po files are not updated yet (I will do it tonight), that's why you see them.

The byte strings, (kB, MB, GB) will be cleaned up (organized). I can remove all byte strings if @WanWizard is ok with that!

The plain numbers you see (0, 1, 2, 3, etc) come from the xml files, not the python files, so it's not trivial to remove them. Will have to live with these.

badsystem:
At the expense of translation of units "MHz", "MB", "kB" to solve to you, but in Cyrillic it is accepted to write values in a native language.

persianpros:
In Persian we say:

MB -> مگابایت
GB -> گیگابایت
KB -> کیلوبایت

nautilus7:
Ok, thanks for your input.

So, I won't remove any quantity units from the translation (I haven't done so, anyway).

WanWizard:
Can I close this PR then?

nautilus7:
Yep!

PR #3140, opened 2021-09-18
lt.po + fr.po

pr2git:


po/lt.po
```diff
@@ -1108,8 +1125,9 @@ msgstr "Kūrimas"
 msgid "Auto"
 msgstr "Automatinis"
 
+#, fuzzy
 msgid "Auto DiSEqC"
-msgstr "Automatinis DiSEqC"
+msgstr "Automatinis Diseqc"
```
nautilus7, line 1130: Why this change? Even in the LT the translation there is no consistency for this term...

po/lt.po
```diff
@@ -1519,7 +1537,7 @@ msgid "Cannot determine"
 msgstr "Nustatyti nepavyko"
 
 msgid "Cannot find any signal ..., aborting !"
-msgstr "Jokio signalo nerasta ..., nutraukiama !"
+msgstr "Jokio signalo nerasta …,  nutraukiama !"
```
nautilus7, line 1540: Same here, the ellipsis is written as 3 full stops (...) in other cases in this translation.

po/lt.po
```diff
@@ -4235,8 +4255,8 @@ msgid ""
 "If you are happy with the result, press OK."
 msgstr ""
 "Jeigu jūsų televizorius turi šviesumo arba kontrasto išplėtimą, išjunkite jį. Jeigu yra kažkas pavadinta \"dinaminis \", nustatykite jį, kaip standartinį. Nustatykite jums tinkamiausią apšviestumo lygį. Sumažinkite TV kontrastą tiek, kiek tai įmanoma.\n"
-"Tada sumažinkite šviesumą tiek, kiek tai įmanoma, bet tik tiek, kad du pačius žemiausius pilkus atspalvius jūsų akis dar skirtų.\n"
-"Dabar nebesirūpinkite šviesiais atspalviais. Jie bus nustatyti sekančiame žingsnyje.\n"
+"Tada sumažinkite šviesumą tiek, kiek tai įmanoma, bet tik tiek, kad du pačius žemiausius pilkus atspalvius jūsų akis dar skirtų. \n"
+"Dabar nebesirūpinkite šviesiais atspalviais . Jie bus nustatyti sekančiame žingsnyje.\n"
```
nautilus7, line 4259: Is this space before the full stop on purpose?

nautilus7, line 4259: Is this space before the full stop on purpose?

po/lt.po
```diff
@@ -7290,7 +7310,7 @@ msgid "Select channel to record from"
 msgstr "Pasirinkite kanalą įrašymui"
 
 msgid "Select char for replacing newlines in EPG."
-msgstr "Pasirinkite simbolį, skirtą pakeisti naujas linijas EPG."
+msgstr "Pasirinkite simbolį, skirtą pakeisti naujas linijas epg."
```
nautilus7, line 7313: Same here, why not in uppercase?

po/lt.po
```diff
@@ -7328,8 +7348,9 @@ msgstr "Pasirinkite, kaip norite, kad imtuvas išsaugotų tinkamą laiką iš DV
 msgid "Select image"
 msgstr "Pasirinkite atvaizdą"
 
+#, fuzzy
 msgid "Select in which resolution pictures are displayed with picture viewer or movie player."
-msgstr "Konfigūruokite, kokia rezoliucija rodomos nuotraukos nuotraukų naršyklėje ir grotuve."
+msgstr "konfigūruokite, kokia rezoliucija rodomos nuotraukos nuotraukų naršyklėje ir grotuve"
```
nautilus7, line 7353: Same here, this reverts the fixed version.

nautilus7, line 7353: Same here, this reverts the fixed version.

nautilus7 (commented):
In the LT translation there are a lot of strings that I fixed in my previous commits and are now reverted. I highlighted some of them, but there are much more. I don't think these are done on purpose. I think the lt.po file used for the latest update by the translator was not up-to-date with the latest lt.po in github, so it accidentally reverts some of the latest fixes.

I don't think this should be committed.

@pr2git next time I suggest you make a different PR for the fr.po which is done by you (thus it is checked) and a different one for other .po files posted in the forum.

pr2git:
That's normal you are working the wrong way.
You need to fix stuff in the source code and that's all, you should not change anything in the existing translation. Just let translators do the job.
The problem is that you change all the po files while the normal way is to change the source and then generate the newest lt.po without changing anything yourself. Then the changed strings in the source code will be taggued as fuzzy and translators will saw them and apply the changes.

There is always a difference between po files posted on the forum and the one in github. So what I do is download the file place it into the po file then generate the updated version and commit. Many translators just pickup the file when they saw that it is changed in github and start working on it.
But since you made changes in the lt.po between the latest published po and the translation proposed your change are lost.
If you don't touch the lt.po yourself the string will simply be marked as fuzzy and next time the translator will fix them.

So please stop changing existing translation for languages that you don't manage yourself and the problem will be solved by itself. Change in source will be taggued as fuzzy and translators will review them. Without this fuzzy tag we don't know what you have changed so we cannot review them as translator.

So you get frustrated because your change are lost and translators get frustrated because there work are rejected.
So if you use the po with the standard procedure everything will smoothly be properly translated. Changing source strings and gettings fuzzy strings in return during the next generation is the normal process po update process and you try to work-around it but it is wrong.

nautilus7:
I understand what you say.

For me it much easier and less time consuming (I estimate 3-4 times more time spent) if I don't update the po files. It's not my preferred way to work, anyway. I believe the strings should be fixed in the code and that's it.

Some years ago, I started fixing some strings that were badly written, just the way you describe, but there were some arguments from other members/contributors that existing translations will be lost (some marked as fuzzy, thus not used unless fixed, others will be lost completely if changes are broader). And the decision was made at that time to update the po files manually as well.

I don't know whether anything has been changed in the direction you describe since then, but no one told me not to update the .po files like I do...

@WanWizard what's your opinion on this?

pr2git:
Yes but years ago the po file in the github of OpenPLi were nearly never updated there was no translation update. So I can imagine that people are complaining in case of changes.
Now the translation process is described in the wiki, the po file are regularly updated and there is a follow-up.
If users complains on translation we can easily send them to the wiki page where everything is explained.
So the situation is totally different.
Personally I dislike the "sed like" changes in translation because you need more a less space to translate and english sentence. For example french sentence are most of time longer than the english one, so when translating I am sometimes faced to the need to reduce some words or use some abbreviation to have the text fitting:  Example: Save in english is translated by Sauvegarde (or Enregistrer) in French and this does't fit on most of the button so I use Svgd instead of Sauvegarde for the buttons.
So I'd like to have control and to have any changed string taggued as fuzzy for review.
Moreover now that OpenPLi is using released version it is less problematic to lose some translation into the develop branch. This lets time for translators to work. So a released image won't have problem with existing translation.
So the situation is totally different than years ago.

nautilus7:
No problem by me. As I told you, I like to move things forward and fix things I don't like, even at the cost of loosing some translations or breaking compatibility in other cases.

nautilus7:
@pr2git

I'll rebase the lt.po to the latest changes and make a new PR, so no harm done. It's easy enough for me, but not for a translator, who doesn't know coding and git. I understand it.

On the other side, your fr.po seems fine. Do you want to make a new PR with just that, or close this one and let me do both?

WanWizard:
What is the status on this?

nautilus7:
@WanWizard you can close this one and merge #3151 instead.

pr2git:
No problem. Thanks nautilus7 for the job. :-)

PR #3141, opened 2021-09-18
[po] Update ar (Arabic) file translation

fairbird:


po/ar.po
```diff
@@ -512,33 +450,34 @@ msgstr "4:3 شاشة مصغرة"
...
 msgstr "4:3 شاشة عريضة"
 
+#, fuzzy
 msgid "4K"
-msgstr ""
+msgstr "4K"
 
-#, fuzzy
 msgid "5"
 msgstr "25"
```
nautilus7, line 458: You missed that!

fairbird: My translation depend on en.po (I think No need it)
https://github.com/OpenPLi/enigma2/blob/develop/po/en.po#L446

nautilus7: no idea

WanWizard:
What is the status of this PR?

po/ar.po
pr2git: There is a wiki page on how to contribute to translation:
https://wiki.openpli.org/Information_for_translators

So you means that you start from the en.po to create your ar.po?
This is not really the easiest way I think but since the pot file is not in the repository of OpenPLi you cannot start from 
it.
Why didn't you start from the current existing ar.po?

Did you test your own translation on in your OpenPLi?

About the fuzzy strings it is always better to review them and tag then has non fuzzy when done. A fuzzy string is ignored when generating the mo file anyway so the original string is then displayed.
So the fuzzy string pointed by nautilus won't arm (but you should fix it on your next translation PR, so indeed 5 will be displayed and not 25).

WanWizard:
Closed due to lack of response.

PR #3156, opened 2021-09-24
[gpixmap.h] gRGB::operator const std::string () const conversion incorrect

nautilus7:
gRGB::operator const std::string () const should now correctly
decode numbers to hex strings.

I have not been able to test this fix in context, only with the
method body in a test harness. An alternative fix is:

escapecolor[i] = "0123456789abcdef"[val & 0xf];

or similar, but that may be slightly less efficient.

This fix should also help address, or even fix, a bug where subtitle
colours go black.

(cherry picked from commit 37503289183fcc7b27338f7f998d68246d1c42d6)

### DIFF
--- a/lib/gdi/gpixmap.h
+++ b/lib/gdi/gpixmap.h
@@ -107,7 +107,9 @@ struct gRGB
 		escapecolor.resize(10);
 		for (int i = 9; i >= 2; i--)
 		{
-			escapecolor[i] = 0x40 | (val & 0xf);
+			int hexbits = val & 0xf;
+			escapecolor[i] = hexbits < 10	? '0' + hexbits
+							: 'a' - 10 + hexbits;
 			val >>= 4;
 		}
 		return escapecolor;


nautilus7:
@WanWizard @Huevos openvision has this change while openpli doesn't. I haven't tested, but I think this will do the work.

WanWizard:
Ok, so the root cause was the first PR was dependent on another one we didn't have?

I'll merge and build a test image.

nautilus7:
Yep, something like this... Looking at the [gpixmap.h history](https://github.com/OpenVisionE2/enigma2-openvision/commits/develop/lib/gdi/gpixmap.h) in openvision, they merged the [first commit](https://github.com/OpenVisionE2/enigma2-openvision/commit/c872d3995cd191ce4b35d1e6b47951ffcab23fd9#diff-d7a4034f376dbe282c26b16a1edbbb8a73661e821abcf79994a09575cb4af7d7) (that caused your problem) and [this one](https://github.com/OpenVisionE2/enigma2-openvision/commit/37503289183fcc7b27338f7f998d68246d1c42d6#diff-d7a4034f376dbe282c26b16a1edbbb8a73661e821abcf79994a09575cb4af7d7) (that fixes it - or like you said one is dependent on the other) from prl at Aug 17, 2019 (2 years back), but they do not included the changes in the python files (the new color formatting)... The result was broken colors in python (some skin widgets that required the hex2colorstr), so they [reverted only one of them](f39dfd878b056f2f3f352b044db17d98556a339b) (our first one) a month later.

Then 1,5 year later they re-added the reverted commit but [properly this time](https://github.com/OpenVisionE2/enigma2-openvision/commit/b449483ac72d63e3d47e8d52b244585b30ed1f9f#diff-d7a4034f376dbe282c26b16a1edbbb8a73661e821abcf79994a09575cb4af7d7) with the python color strings corrected. This was done on Feb 22, 2021. I only saw this commit and made a PR for us, but not the other one (the not reverted in openvision) from 2019.

Huevos:
All those commits come from Beyonwiz by PRL.

https://bitbucket.org/beyonwiz/easy-ui-4/commits/4d0bbf11b3057a2324144bfbe1af3ff034ce9b55

Except dropping the use Hex2ColorStr.py...

nautilus7:
Yes, I mentioned that prl is the author, although I have no idea that he develops for beyonwiz and where their code is.

I can't be aware of all enigma2 teams and code around the world, but thanks for the heads up.

Huevos:
> 
> 
>  I have no idea ... where their code is.
> 

Australia

nautilus7:
I meant where is hosted...

Bitbucket, github, etc

WanWizard:
This has fixed the issue, thanks !

nautilus7:
Thanks to you for noticing the issue soon enough so it was quite easy to determine what caused it.

PR #3159, opened 2021-09-26
fix leak as openATV

Dima73:
-CWE-561 remove dead code
-fix leak bmp_load
-fix leak / close open socked if connect failed


persianpros:
Why don't you give credits to the author?

Dima73:
thanks @jbleyel

PR #3192, opened 2021-11-08
Update ServiceInfo.py

badsystem:


WanWizard:
Are `Mhz` and `Ksymb/s` not internationally used abbriviations?

badsystem:
In many parts of the image, these values are translated

PR #3198, opened 2021-11-13
[ChannelSelection] fix ValueError: invalid literal for int()

Dima73:


s3n0:
LOL, why `try-except` algorithm ? This is dangerous to use because it contains a many (all) exceptions ! And that's dangerous. At the same time, the source code becomes non-debugging, because then you will never know what the error is there, if a completely different error exception appeared there again. Sorry, but this is a stupid Python solution ;-).

As I wrote on the [OpenPLi discussion forum](https://forums.openpli.org/topic/86614-crash-in-channel-list-satellites-green-button/#entry1402110) - **the question is why the length of equal 4 digits is checked in the condition ?** Why not just a length greater than 4 digits ?

I don't know why. I did not study this one python-class :) and its use :). Maybe it has a special meaning.

In any case... if, for example, there was a condition of only **length > 4** instead of **lentgth >= 4** , the problem would be solved. Alternatively, if the range **string[0:2]** is pulled from the string instead of **string[:-4]**, that is, without trimming the last 4 digits (which does not exist and an error exception occurs).

**However, the question still remains as to why there should be only 4 digits.** I've never seen only 4 digits in userbouquet files (I mean in the case of DVB-S standard, in the service-reference-code, in the "namespace" section). I live in Europe. Maybe in other countries different "namespaces" are used in Enigma2 ? :-) Really, I don't know. Sorry.

ims21:
Agree - try/except is way to hell ... if you want get last 4 digits, change op[:-4] to op[-4:] , thats all...

Dima73:
Ok.

PR #3211, opened 2021-12-02
[InfoBarGenerics] add service name to subservices

Dima73:
thanks @Huevos

Item will now read...
servicename, event time, event
title

Having the event time in second position means...
1) no addition
delimiter is needed, keeping things as short as possible,
2) event time
won't get forced of the screen by extra long event titles.


littlesat:
Please use “%s” % structure instead of “ “ + style

and are you sure it is not too much (please add screenshot)

Huevos:
Extra long title will go off the screen like they always did.

Previously the time went off the screen so was not readable.

Now time is in second position so always is readable.

And my font is a lot bigger than the PLiHDNight font, so more will show in your skin.

Previously no channel name showed and users say the feature was unusable like that.

PR #3216, opened 2021-12-09
Update uk.po

badsystem:


WanWizard:
uk.po:7139:62: syntax error

WanWizard:
@littlesat why was this merged, I already commented that it contained errors?

PR #3229, opened 2021-12-22
FlashImage - fixed typo

ims21:


nautilus7:
Maybe @WanWizard prefers to do an amend commit to keep the git history clean. Anyway, what ever you like.

WanWizard:
Sorry for the typo.

You can't do an amend of a commit that is already pushed public, that would rewrite history and create problems for people that have already pulled it.

nautilus7:
True, but git experts can overcome them.

WanWizard:
I'm not expecting everyone is... :grin:

PR #3234, opened 2021-12-25
ServiceTime - StartTime - returns time for directory too

ims21:


littlesat:
Why not solve in the general function that retrieves the time.It gives me the feeling that it is solved I. The incorrect location.

PR #3236, opened 2021-12-29
add lcdpicon.py needed by enigma2-display

luisteraar:
Needed by enigma2-display in the feature


littlesat:
Please revert it as it is duplicated code.... when you want to use a different lcd skin (eg from openatv) then change the skin to use the already existing picon renderer. This code is duplicated code and therefore muda.

PR #3242, opened 2021-12-31
[Picon] add new skin attrib "usePicLoad" for use ePicLoad

Dima73:
Need for some box(e.g. gigablue uhd ue/quad 4k) in lcd display.
- example:
<widget source="session.CurrentService" render="Picon"
isFrontDisplayPicon="1" usePicLoad= "1" zPosition="13"
position="224,394" size="242,142" transparent="1" alphatest="blend">

<convert type="ServiceName">Reference</convert>
  </widget>

### DIFF
--- a/lib/python/Components/Renderer/Picon.py
+++ b/lib/python/Components/Renderer/Picon.py
@@ -167,8 +167,13 @@ class Picon(Renderer):
 		if self.instance:
 			if self.showPicon or config.usage.show_picon_in_display.value:
 				pngname = ""
-				if what[0] != self.CHANGED_CLEAR:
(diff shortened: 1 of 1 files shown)


littlesat:
Why do we need the function updatePicon (here)

Dima73:
Don't know, original code https://github.com/openatv/enigma2/blob/7.0/lib/python/Components/Renderer/LcdPicon.py

Taapat:
I have a feeling that the LcdPicon code is from old images that do not have the function setPixmapFromFile, or in which the image scale does not work.

nautilus7:
These commits should have been merged as a squash commit.

PR #3254, opened 2022-01-06
Long button press emulation

Dima73:
- added an ability to emulate long button presses with two normal
(short presses). One button on the remote must be defined for the long
press emulation purpose. Then a normal (short) press on the emulation
button followed by a normal (short) press on a second button emulates a
long press on the second button;
- new setting “Long press emulation with button XXX” on button setup
screen to define the button for that purpose.

-thanks openATV @team

### DIFF
--- a/data/setup.xml
+++ b/data/setup.xml
@@ -45,6 +45,7 @@
 		<item level="2" text="Alternative URL for images listed on FlashImage" description="Enter an URL where alternative images could be listed for FlashImage" requires="canFlashWithOfgwrite">config.usage.alternative_imagefeed</item>
 		<item level="1" text="Time synchronization method" description="Select how you want your receiver to keep the correct time, either from the DVB transponder, or from the internet using NTP. Or use Auto, which will favour using NTP, but will fall back to transponder time when no internet connection is available.">config.ntp.timesync</item>
 		<item level="1" text="NTP Hostname" conditional="config.ntp.timesync.value != 'dvb'" description="The hostname or the IP address of the NTP server to synchronise the time with.">config.ntp.server</item>
+		<item level="0" text="Long press emulation with button" description="Press this button to emulate a long press on a button pressed afterwards (this also overrides any other actions assigned to emulation button).">config.usage.long_press_emulation_key</item>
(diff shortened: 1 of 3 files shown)


WanWizard:
What is the use case for this change?

Dima73:
Press fixed key(set in the settings), e.g. TV, and press any key, e.g. blue --> this emulates is long press blue.
Many remotes do very poorly for long press.

littlesat:
I understand they define a kind of shift button to substitute long press because some remotes do no respond good to long press...

I also think OpenATV needs it as they define short an long press on any button... so all buttons in their image are release or long press (hold a specific time) that also 'extend' their respond time a couple of hundred of miliseconds when a button is pressed.

I strongly recommend not doing this with as reason long key does not work....

I think the background idea adding a shift button to 'extend' hotkeys could be a good idea.... but I think when we add this we should 'offer' a standard key for this to keep it simple...

I think this suggested patch is a terrible hack... they also changed cpp code which is for a shift function totally not needed... double click can also have other timing issues what is challenging but I think you can perfectly solve the same in pure python UI code... You just put a screen after the shift with a simple timeout that closes the screen, gives a shift picon on the screen, and then 'call' the right function at the second key press when it occurred on the make of it within a specific time.

And they still named hotkeys button setup, while I think hotkey is a better name for it as it is no keymap.xml editor... ;)

ims21:
Sorry, I must agree with littlesat... 

Btw - some as "shift" button we have implemented long time in PLi by miracle.
It can be used instead it ... user can define own /etc/enigma2/keytranslation.xml file, where he can write, which button will be used as "shift" and then write there remaping for buttons for using after pressed 'shift'

Dima73:
> It can be used instead it ... user can define own /etc/enigma2/keytranslation.xml file, where he can write, which button will be used as "shift" and then write there remaping for buttons for using after pressed 'shift'

Can an example ?

ims21:
see [here](https://github.com/OpenPLi/enigma2/commit/65df3da8cb5d50934ef36bcf8c974b8b437c3708)

PR #3261, opened 2022-01-09
[frontend] update snr/db gigablue plug & play tuners

Dima73:


Dima73:
[[frontend] correct snr/db GB plug & play tuners](https://github.com/OpenPLi/enigma2/pull/3285)

PR #3262, opened 2022-01-09
[Picon] if use usePicLoad and old service same as source.text stop fi…

Dima73:
…nd picon


Dima73:
usePicLoad method works great for solo4k display.

PR #3264, opened 2022-01-09
Videomode - fixed helptexts

ims21:


nautilus7:
You have accidentally changed the file mode from 644 to 755 as well.

nautilus7:
The getCurrentDescription() method on line 197 can be removed completely. It's identical to the parent method from ConfigListScreen.

ims21:
Fixed. I created it remotely and must pass it over ntfs disk, and did not removed redundant lines...

PR #3267, opened 2022-01-09
Add led settings in display setup for gbue4k and gbquad4k

Taapat:
At the request of Dimitrij.
This allows in display setup turn the receiver led on or off in various modes.


littlesat:
A bit another Gb exception....
Why can't they do the same as others :( see how much extra code is required for it in e2....

Taapat:
I agree.
I don't know how many other manufacturers use the existing solution in enigma.
The existing solution may be an exception for another manufacturer ;).

My opinion is that without such settings the led works properly.
I only offer this because I was asked to do so. Some users want to turn off the led in standby or in other modes.

PR #3283, opened 2022-01-16
Properly call maxsize from sys module

nautilus7:
After commit 2df14421705de1465ea50b1881b2c189d4462d9a the call to maxsize was done improperly, so commits 2711e0ff3139c8728e90391b37ac0636484cb422 0dc1bb8316f69edc5e9763796e8a80f4344c4a21 b33fc21e3f2670765907972c12701ccc1fd6705c and 1b176b6545d087d351e3491d3af582ae4bbc38e3 were done to fix the problem.

This commit fixes the initial problem, by basically reverting the flawed commits.


nautilus7:
@Hains mystery solved

Hains:
Yes thank.
I wasn't happy with it either.

PR #3286, opened 2022-01-19
[Satfinder] add more config if not sat list

Dima73:
File
"/usr/lib/enigma2/python/Plugins/SystemPlugins/Satfinder/plugin.py",
line 164, in createSetup
ValueError: invalid literal for int() with base
10: ''

### DIFF
--- a/lib/python/Plugins/SystemPlugins/Satfinder/plugin.py
+++ b/lib/python/Plugins/SystemPlugins/Satfinder/plugin.py
@@ -325,7 +325,7 @@ class Satfinder(ScanSetup, ServiceScan):
 				continue
 			if n.config_mode in ("loopthrough", "satposdepends", "nothing"):
 				continue
-			if n.isCompatible("DVB-S") and n.config_mode == "advanced" and len(nimmanager.getSatListForNim(n.slot)) < 1:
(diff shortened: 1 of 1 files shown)


nautilus7:
> File "/usr/lib/enigma2/python/Plugins/SystemPlugins/Satfinder/plugin.py", line 164, in createSetup ValueError: invalid literal for int() with base 10: ''

This message has nothing to do with your changes....

Dima73:
Install new image without restoring settings(not configured tuners) and open this plugin

nautilus7:
I don't want to do anything. I am just saying that your error message is NOT related with your changes in this PR.

ims21:
> I don't want to do anything. I am just saying that your error message is NOT related with your changes in this PR.

indirectly yes  ... there is problem, that when there is some as None, cannot be used int().
But it should be fixed on line with "int"

Dima73:
> > I don't want to do anything. I am just saying that your error message is NOT related with your changes in this PR.
> 
> indirectly yes ... there is problem, that when there is some as None, cannot be used int(). But it should be fixed on line with "int"

Not.
Install new image,default sat tuner config "Simple/diseqc(A/B)/nothing connected".Empty sat list for this tuner.
So...
When not sat list for this tuner,  let's skip it.

if n.isCompatible("DVB-S") and n.config_mode in ("simple", "equal", "advanced") and len(nimmanager.getSatListForNim(n.slot)) <1:
				continue
This all.

ims21:
Ok

PR #3311, opened 2022-01-22
MovieList.py: Write bytes to file

nautilus7:
We open the file in binary mode, so write binary data to it

### DIFF
--- a/lib/python/Components/MovieList.py
+++ b/lib/python/Components/MovieList.py
@@ -129,7 +129,7 @@ def resetMoviePlayState(cutsFileName, ref=None):
 				cutlist.append(data)
 		f.close()
 		f = open(cutsFileName, 'wb')
-		f.write(''.join(cutlist))
+		f.write(b''.join(cutlist))
 		f.close()
 	except:
 		pass


ims21:
It seems, it could be problem with crashing e2 when is working with cuts...

nautilus7:
Is it crashing with this change or without it?

The .cuts file is definitely a binary file, so maybe it needs some changes.

nautilus7:
Do you actually get a crash? I can't make it crash with or without my commit and I know why...

This line
```
f.write(''.join(cutlist))
```
is exactly the same as this
```
f.write(b''.join(cutlist))
```

because you basically join the cutslist with an empty string. It doesn't matter if it is binary or text.
It only makes my python linter happy and it is syntactically correct, since we work with bytes here.

nautilus7:
@ims21 I just read the error report in the forum.... This PR has nothing to do with it, as it's for the python3 branch.

Regards

ims21:
My note was about it, that it could be solution for occassional crashing when is worked with cutlist editor. For python2 it should be for writing cutlist to .cuts as binary same.

PR #3313, opened 2022-01-23
MovieList - write cutlist to binary file as binary.

ims21:
It should remove occassionaly crashing in cutlist editor.

Thx for the inspiration by nautilus7.

### DIFF
--- a/lib/python/Components/MovieList.py
+++ b/lib/python/Components/MovieList.py
@@ -128,7 +128,7 @@ def resetMoviePlayState(cutsFileName, ref=None):
 				cutlist.append(data)
 		f.close()
 		f = open(cutsFileName, 'wb')
-		f.write(''.join(cutlist))
+		f.write(b''.join(cutlist))
 		f.close()
 	except:
 		pass


nautilus7:
It shouldn't do anything in python2... Bytes is the default for strings, unlike python3 which works with text strings by default.

ims21:
ok. There must be different problem, as I saw it now ...

nautilus7:
Yep, just to be clear, see this:

```
Python 2.7.18 (v2.7.18:8d21aa21f2, Apr 20 2020, 13:25:05) [MSC v.1500 64 bit (AMD64)] on win32

>>> c = b""
>>> isinstance(c, str)
True
>>> isinstance(c, bytes)
True

>>> c = ""
>>> isinstance(c, str)
True
>>> isinstance(c, bytes)
True
```

In python 2 str and bytes are the same. But unicode string is different.
In python 3, unicode became str (text string) and str/bytes became bytes.

In python3, things are more clear. That means we have text stings which are unicode/utf-8 and byte strings.

PR #3319, opened 2022-01-23
Allow unicode strings in user bouquet filenames

nautilus7:
Until now, only ascii characters were allowed in user bouquet file names and non ascii were stripped away, for example "userbouquet.mybouquet.tv." Since unicode is the default for python3 and we are in the 2022, this should be changed. In addition this could lead to weird naming when a user creates a bouquet using only non ascii characters as they are all filtered out by the old code. For example a bouquet name of "ελληνικά" would result in a corresponding user bouquet file name of "userbouquet..tv".

This PR will allow for all valid utf-8 characters to be used as bouquet file names, except from some special or blacklisted characters (in windows, etc.). The old code prevented these special characters as well.

The same restrictions and freedom is applied in names of picons, meaning picons can now contain utf-8 characters. In other words the picon name can match the service name (although the code might need some polish for this to work).

The whole file name check is moved to a new function in the "Directories.py" module. This function is taken from the "sanitize-filename" module in https://pypi.org/project/sanitize-filename/ with the addition of a couple of blacklisted characters (parenthesis and space) which were also blacklisted in the old code.

The "sanitizeFilename" function makes some more tests (like the maximum number of characters allowed in windows for a file name), which may seem overkill for our use case, but the point is to have reusable code that can applied to other files like recordings, etc.

### DIFF
--- a/lib/python/Components/Renderer/Picon.py
+++ b/lib/python/Components/Renderer/Picon.py
@@ -1,11 +1,10 @@
 from __future__ import print_function
 import os
 import re
-import unicodedata
 from Components.Renderer.Renderer import Renderer
 from enigma import ePixmap, ePicLoad, eServiceCenter, eServiceReference, iServiceInformation
 from Tools.Alternatives import GetWithAlternative
-from Tools.Directories import pathExists, SCOPE_SKIN_IMAGE, SCOPE_CURRENT_SKIN, resolveFilename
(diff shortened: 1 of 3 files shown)


nautilus7:
@athoik any suggestions or modifications? I did minimum testing and works with weird bouquet names as expected. Existing picons load as well.

athoik:
I think it's a great step forward!

Hains:
:+1:

PR #3323, opened 2022-01-25
[SystemInfo] add "HasHiSi" check when is /proc/hisi entry

Dima73:


WanWizard:
What is the relation with "Viper"?

Dima73:
> What is the relation with "Viper"?

I looked at the code here https://github.com/OpenViX/enigma2/blob/Dev-python3-compatible/lib/python/Components/SystemInfo.py#L16, there is an exception some viper box.

WanWizard:
Ah, Ok.

I assume it is for the Amiko Viper 4K series, for which we don't make an image. 

Besides that, it looks like a hack for some "imagebackup" feature of VIX, if the commit text is correct, so I think it is safe to remove it.

PR #3328, opened 2022-01-27
[FlashImage] fix multi search folder

Dima73:
When use two folder "downloaded_images"(zip image) and
"imagebackups"(zip
image) visible only one folder and one zip file.
Can't use 'media' variable twice in a loop.

### DIFF
--- a/lib/python/Screens/FlashImage.py
+++ b/lib/python/Screens/FlashImage.py
@@ -67,14 +67,16 @@ class SelectImage(Screen):
 
 	def getImagesList(self):
 
-		def getImages(path, files, folder=None):
+		def getImages(path, files):
 			for file in [x for x in files if os.path.splitext(x)[1] == ".zip" and model in x]:
 				try:
-					folder = "Image backups" if folder == "imagebackups" else "Downloaded Images"
 					if checkimagefiles([x.split(os.sep)[-1] for x in zipfile.ZipFile(file).namelist()]):
-						if folder not in self.imagesList:
-							self.imagesList[folder] = {}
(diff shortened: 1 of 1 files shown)


lib/python/Screens/FlashImage.py
```diff
@@ -67,14 +67,16 @@ def __init__(self, session, *args):
...
+		def getImages(path, files):
 			for file in [x for x in files if os.path.splitext(x)[1] == ".zip" and model in x]:
 				try:
-					folder = "Image backups" if folder == "imagebackups" else "Downloaded Images"
 					if checkimagefiles([x.split(os.sep)[-1] for x in zipfile.ZipFile(file).namelist()]):
-						if folder not in self.imagesList:
-							self.imagesList[folder] = {}
-						self.imagesList[folder][file] = {'link': file, 'name': file.split(os.sep)[-1]}
+						imagetyp = _("Downloaded Images")
+						if 'backup' in file.split(os.sep)[-1]:
```
WanWizard, line 75: Isn't this a bit too generic? This would include the autobackup directory.

Dima73: PLi-AutoBackup.tar.gz not *.zip

Dima73:
https://github.com/openatv/enigma2/commit/87fa67f9d97fa27f383fb2cbcdcf4f6dc97091bf

lib/python/Screens/FlashImage.py
WanWizard: I have more in that directory than only that file, stuff that I want to restore too.

Dima73:
Fullbackup plugin, and now BackupSuite plugins and make zip archives with the word backup in the title.
This patch is for that.

PR #3329, opened 2022-01-27
[ChoiceBox] left/right add updateSummary

Dima73:


Taapat:
In updateSummary is already used displayDescription: https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Screens/ChoiceBox.py#L213
If you add updateSummary here then you need to remove displayDescription to not to duplicate it call twice.
But in that case this function updateDescription does not do what its name says.
Maybe it would be more correct use updateSummary in those places where it necessary instead of updateDescription: https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Screens/ChoiceBox.py#L147-L150

PR #3335, opened 2022-02-01
Fix descriptions not showing -- Urgent

nautilus7:
This PR fixes broken descriptions in ConfigListScreen without changing the SetupSummary class and making it incompatible with most plugins out there.

Many screens that inherit from ConfigListScreen, used to just define `self["description"] = Label("")` and have descriptions properly displayed and updated for every configList entry. This is done by a hook in the SetupSummary class (which is called inside the createSummary method in ConfigListScreen). By moving this hook from SetupSummary to the Setup class with commit 93e1f6784f225b2e85eb9a0619df123f2460b30b in plugins like Videomode the descriptions were disappeared. Then commit https://github.com/OpenPLi/enigma2/commit/643d12de290732844295a6b6ab2aa34378e4e710 was done to bring them back, by redefining the same methods already defined from parent classes. Apart from the obvious problem (code replication), this was not done (it's not possible) for every other plugin out there!!!

In this PR, both of the above commits are reverted and a proper fix for the HdmiCECSetup screen is applied. The special thing about descriptions is this screen is that we want some extra text to be displayed along with the description is each configList entry). 

This bug is almost 3 months old, but I just noticed it yesterday. But report here: https://github.com/OpenPLi/enigma2/commit/93e1f6784f225b2e85eb9a0619df123f2460b30b#commitcomment-65416190

Note: The current code in the the setup and config/confgilist/confgilistscreen classes is like spaghetti. Competitive images have made significant updates the past couple of years in this code, but these updates haven't find their way to openpli. I plan to bring these updates here, after the python3 transition is made. So, for now do now make any significant changes to this code. Thanks.

### DIFF
--- a/lib/python/Plugins/SystemPlugins/HdmiCEC/plugin.py
+++ b/lib/python/Plugins/SystemPlugins/HdmiCEC/plugin.py
@@ -34,7 +34,8 @@ class HdmiCECSetupScreen(Screen, ConfigListScreen):
 
 		self.list = []
 		self.logpath_entry = None
-		ConfigListScreen.__init__(self, self.list, session=self.session, on_change = self.changedEntry)
+		ConfigListScreen.__init__(self, self.list, session=self.session, on_change=self.createSetup)
+		self["config"].onSelectionChanged.append(self.updateDescription)
 		self.createSetup()
 		self.updateAddress()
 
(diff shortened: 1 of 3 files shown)


ims21:
And could you tell me, where is reverted patche, which all this started ? It was huevos "vix" part "implemented" to PLi and corrupting worked CEC texts... 

Not all boxes will go to py3...

nautilus7:
What are you talking about? Does this look like a python3 problem or a python3 pull request?????

Huevos broke PARTIALLY the HdmiCECSetup descriptions...

YOU BROKE DESCRIPTIONS COMPLETELY IN MANY MANY PLUGINS!!!! AND YOU KNOW THAT FOR 3 MONTHS NOW (SINCE YOU MADE THE FIX IN VIDEOMODE PLUGIN) BUT YOU' VE DONE NOTHING FOR THAT!!! THAT IS VERY IRRESPONSIBLE BEHAVIOUR OF YOU!!!

Give me a break now.

WanWizard:
Gents, please keep it civilized.

Taapat:
@nautilus7 try this changes: https://github.com/OpenPLi/enigma2/pull/3336

nautilus7:
I replied in your PR, I will test later when I get back home.

nautilus7:
#3336 restores broken descriptions as well, and places the description hook inside the ConfigList which is a more appropriate place.

PR #3336, opened 2022-02-01
Set description in ConfigListScreen

Taapat:
This update description on config selection change in ConfigListScreen.
There are many screeen where there is only a defined description but they rely on SetupSummary to display it.
Now the description display in the parent screen has been removed from SetupSummary and this no longer works.
Therefore introduces a description display in ConfigListScreen.
See also: https://github.com/OpenPLi/enigma2/pull/3335

### DIFF
--- a/lib/python/Components/ConfigList.py
+++ b/lib/python/Components/ConfigList.py
@@ -206,6 +206,8 @@ class ConfigListScreen:
 		elif "VKeyIcon" in self:
 			self["VirtualKB"].setEnabled(False)
 			self["VKeyIcon"].boolean = False
+		if "description" in self:
+			self["description"].text = self.getCurrentDescription()
 
 	def KeyText(self):
 		self["config"].hideHelp()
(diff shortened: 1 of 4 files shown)


nautilus7:
Is the createSummary method required for the description to show up? If yes, then your changes are NOT compatible with existing plugins.

Taapat:
In SetupSummary description is set only for the parent screen.
Therefore, I wrote that SetupSummary is not the right place for this, and it is better to move it elsewhere.

nautilus7:
Yes, I see. It seems fine, so I am ok with that. I am ok with wherever you put description in, as long as it works and it is compatible with older code like it seems it is now. Thanks.

nautilus7 (approved):

nautilus7:
Everything works ok, so I closed my PR. Awaiting the present PR to be merged as it remains an urgent fix. Descriptions are broken for 3 months now.

PR #3345, opened 2022-02-05
Fix and optimize iterables for dictionaries

nautilus7:


nautilus7:
It should be ok, but check them to be 100% sure.

Hains (approved):

PR #3362, opened 2022-02-08
Rework "PluginDescriptor"

nautilus7:
This PR initially reverts commits 8cc034a7be1a9b94dbb18eb50ca8cd0a30474e0c and e22ba45f17298388d23e9bc71343364a70ff63d3 in the python3 branch which do NOT make any sense... at all! The problem trying to fix with the 2 reverted commits was:
```
File "/usr/lib/enigma2/python/Components/PluginComponent.py", line 106, in readPluginList
    self.addPlugin(p)
  File "/usr/lib/enigma2/python/Components/PluginComponent.py", line 31, in addPlugin
    plugin(reason=0)
TypeError: 'PluginDescriptor' object is not callable
```

In python3, in order to run our plugins like this:
```
for p in self.getPlugins(PluginDescriptor.WHERE_MENU):
	res += p(menuid)
```

they must be callable, that is implement the special `__call__` method. The PluginDescriptor class, which is what `p` is in the above example, did not have a `__call__` method. Instead, it defined a `__call__` attribute that was set equal to the `fnc` argument, which is essentially what we want to run. This worked in python2, but not in python3.

So, this PR, fixes the mess and clears the confusion created by having a `__call__` attribute in the PluginDescriptor class (that crap exists in the code for decades!) by properly defining the `__call__` method, so plugins can be called.

Do you see any room from improvement in this PR?

In openatv for example they have put some basic "safety" measures in their [respective class](https://github.com/openatv/enigma2/blob/7.0/lib/python/Plugins/Plugin.py). They return an empty [ ] in case a plugin defines a `fnc` that is not callable. Also they define a `__getattribute__` method for old code still using the `__call__` attribute expecting to get the plugin's `fnc`. Is there any point to this? I have updated all such code in enigma2.

What about putting some checks in the PluginDescriptor class for the different "WHERE..." cases? Depending on what is the "where" argument, the "fnc" argument differs significantly! It can be a function to a list or a dictionary! I believe it would be nice having some basic checks of what a plugins throws us as a "fnc". 

Opinions?

### DIFF
--- a/lib/python/Components/Network.py
+++ b/lib/python/Components/Network.py
@@ -609,7 +609,7 @@ class Network:
 	def msgPlugins(self):
 		if self.config_ready is not None:
 			for p in plugins.getPlugins(PluginDescriptor.WHERE_NETWORKCONFIG_READ):
-				p.__call__(reason=self.config_ready)
+				p(reason=self.config_ready)
 
 	def hotplug(self, event):
 		interface = event['INTERFACE']
--- a/lib/python/Components/PluginComponent.py
+++ b/lib/python/Components/PluginComponent.py
@@ -27,7 +27,7 @@ class PluginComponent:
 			for x in plugin.where:
 				insort(self.plugins.setdefault(x, []), plugin)
 				if x == PluginDescriptor.WHERE_AUTOSTART:
-					plugin.__call__(reason=0)
+					plugin(reason=0)
 		else:
 			self.restartRequired = True
 
@@ -133,7 +133,7 @@ class PluginComponent:
 	def getPluginsForMenu(self, menuid):
 		res = []
 		for p in self.getPlugins(PluginDescriptor.WHERE_MENU):
-			res += p.__call__(menuid)
+			res += p(menuid)
 		return res
 
 	def clearPluginList(self):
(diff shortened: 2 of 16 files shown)


WanWizard:
This exceeds my python skill level... :smile:

athoik (approved):
I think it's good that __call__ now checks if fnc is callable, that will help reduce code in plugins (no need to check if __call__ is callable).

Using p.fnc if fine as well, since we are passing a fnc on constructor, better to use fnc than __call__. 

So let's merge it, but please note, if there is any plugin that still uses p.__call__ it will fail.

nautilus7:
Yes, that's true, they will crash. 

I have detected one such case in the [plugins repo](https://github.com/OpenPLi/enigma2-plugins/blob/python3/pluginhider/src/PluginHiderSetup.py#L91), but we can fix it. The real problem is with third party plugins though. That's why Captain wrote this [failback](https://github.com/openatv/enigma2/blob/7.0/lib/python/Plugins/Plugin.py#L115). Better add it as well, don't you think?

Hains (approved):

WanWizard (approved):

PR #3366, opened 2022-02-13
Directories: add in resolve list only directories that exist

Taapat:
This should increase search speed.
Usually config directories are added at the beginning of the resolve list, which in most cases do not exist for the standard user.
Using resolveFilename these non-existent paths are checked all the time.
This adds in the resolve list only existing directories so as not to check the many non-existent paths on each resolveFilename call.

### DIFF
--- a/lib/python/Tools/Directories.py
+++ b/lib/python/Tools/Directories.py
@@ -74,11 +74,7 @@ defaultPaths = {
 
 
 def addInList(*paths):
-	resolve_list = []
-	for path in paths:
-		if os.path.isdir(path):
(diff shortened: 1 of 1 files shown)


nautilus7:
resolveFilename() for gui skin, lcd skin and fonts scope isn't almost exclusively used when loading enigma2 and not afterwards? 

Does it offer a real benefit?

nautilus7:
I answer my question... They are also used for loading skin icons in epglist, servicelist, movielist, timerlist, etc.


I don't see a problem with this PR.

PR #3368, opened 2022-02-14
[DIrectories.py] remove scopes

Huevos:
Remove SCOPE_PLUGIN_ABSOLUTE and SCOPE_PLUGIN_RELATIVE.

During testing it took 35 seconds to test for 25 files. This is due to using the very inefficient inspect.stack method.

As nothing is using these scopes remove them now before they are used.

### DIFF
--- a/lib/python/Tools/Directories.py
+++ b/lib/python/Tools/Directories.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 import errno
-import inspect
 import os
 
 from enigma import eEnv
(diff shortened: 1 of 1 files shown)


nautilus7:
Just to be clear...

The commit that @Huevos referenced did NOT introduced these inspect.stack() calls. They exist in the code for a very long time now in the develop branch in [L138](https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Tools/Directories.py#L138) and [L221](https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Tools/Directories.py#L221).

I do not understand WHY they are a problem NOW.

Also, if this PR is accepted, then make the similar changes (Huevos or whoever merges this) in the develop branch.

Huevos:
If this is PR is accepted I am happy to remove the same code from develop branch.

Huevos:
> I do not understand WHY they are a problem NOW.
> 

Re: NOW, the other changes reminded me of this, that is all. Obviously this was introduced without proper testing.

`return ','.join([resolveFilename(SCOPE_CURRENT_PLUGIN_ABSOLUTE, "images/busy%d.png" % x) for x in range(1,25) if path.exists(resolveFilename(SCOPE_CURRENT_PLUGIN_ABSOLUTE, "images/busy%d.png" % x))])`

That line takes 35 seconds to run. For me that is not acceptable.

Taapat:
These scopes are introduced here: https://github.com/OpenPLi/enigma2/commit/d75b2d67613f93c05eecafdb227e39977bf451c7
If it takes 35 seconds, then it seems to me that it has not been tested and not used anywhere.

nautilus7:
Why not just delete the inspect.stack() call and leave the scopes as they are? This is what slows things down.

Huevos:
All the scope does is return the path of the source plugin. Surely, there are much more efficient ways to do that directly from the plugin, e.g. `path.dirname(path.realpath(__file__))`

It looks like in a year and a half I am the only person who has ever tried this.

This code is a solution looking for a problem. If it had been written for a real problem in the first place it would already be in popular use, but it is not.

PR #3376, opened 2022-02-24
Add Fast Channel Change support

Huevos:
This is a fully working, stable and tested patch.

Originally from Vu+.

First merged into an OpenPLi based images by OpenVision

Later updated by OpenBlackHole team to get it working properly (previously was slow).

Integrated for use in OpenPLi by Huevos.

Tested in OpenBlackHole, OpenViX and OpenPLi.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -322,6 +322,8 @@ lib/python/Plugins/SystemPlugins/DefaultServicesScanner/Makefile
 lib/python/Plugins/SystemPlugins/DefaultServicesScanner/meta/Makefile
 lib/python/Plugins/SystemPlugins/DiseqcTester/Makefile
 lib/python/Plugins/SystemPlugins/DiseqcTester/meta/Makefile
+lib/python/Plugins/SystemPlugins/FastChannelChange/Makefile
+lib/python/Plugins/SystemPlugins/FastChannelChange/meta/Makefile
 lib/python/Plugins/SystemPlugins/Hotplug/Makefile
 lib/python/Plugins/SystemPlugins/Hotplug/meta/Makefile
 lib/python/Plugins/SystemPlugins/Makefile
--- a/lib/dvb/Makefile.inc
+++ b/lib/dvb/Makefile.inc
@@ -35,7 +35,9 @@ dvb_libenigma_dvb_a_SOURCES = \
 	dvb/pmtparse.cpp \
 	dvb/encoder.cpp \
 	dvb/fbc.cpp \
-	dvb/atsc.cpp
+	dvb/atsc.cpp \
+	dvb/fcc.cpp \
+	dvb/fccdecoder.cpp
 
 dvbincludedir = $(pkgincludedir)/lib/dvb
 dvbinclude_HEADERS = \
@@ -79,4 +81,6 @@ dvbinclude_HEADERS = \
 	dvb/pmtparse.h \
 	dvb/encoder.h \
 	dvb/fbc.h \
-	dvb/atsc.h
+	dvb/atsc.h \
+	dvb/fcc.h \
+	dvb/fccdecoder.h
--- a/lib/dvb/decoder.cpp
+++ b/lib/dvb/decoder.cpp
@@ -15,6 +15,8 @@
 #include <sys/stat.h>
 #include <errno.h>
 
+#include <lib/dvb/fccdecoder.h>
+
 #ifndef VIDEO_SOURCE_HDMI
 #define VIDEO_SOURCE_HDMI 2
 #endif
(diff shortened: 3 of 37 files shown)


RobvanderDoes:
Is it possible for users to disable this? I ask, becauss the Dutch & Belgian smartcards only allow a maximum of 4 channels to be decrypted, and FCC uses 3 of them, without taking into account how many recordings are active. 
That was the reason for PLi not to implement FCC at an earlier stage.

Ev0-BH:
> Is it possible for users to disable this? I ask, becauss the Dutch & Belgian smartcards only allow a maximum of 4 channels to be decrypted, and FCC uses 3 of them, without taking into account how many recordings are active. That was the reason for PLi not to implement FCC at an earlier stage.

Yes it can be disabled completely (default setting), but it can be configured to use anything from 2 - 7 tuners (plus the one you are currently using).

Huevos:
It is disabled by default.

atvcaptain:
vu duo 4k , many black screens and only audio, not perfect with solo4K works

Ev0-BH:
> vu duo 4k , many black screens and only audio, not perfect with solo4K works

No issue with OpenBh 5.0 and FCC on Duo4K or Duo4K SE

Huevos:
> vu duo 4k , many black screens and only audio, not perfect with solo4K works



> vu duo 4k , many black screens and only audio, not perfect

Running OpenViX on the Ultimo4k. First version had black screens but current version is working well so far.

WanWizard:
@littlesat @eriksl Comments please? I know in the past you had a strong opinion about this?

WanWizard:
@Huevos Is this VU+ specific, or does this work on any box with an FBC tuner (or multiple tuners)? Sat and cable?

Ev0-BH:
> @Huevos Is this VU+ specific, or does this work on any box with an FBC tuner (or multiple tuners)? Sat and cable?

It's Vu+ specific for now (fbc only) and sat only (afaik, I don't have cable feed anymore)

Huevos:
> @Huevos Is this VU+ specific, or does this work on any box with an FBC tuner (or multiple tuners)? Sat and cable?

So FBC is Vu+ and Gigablue. Gigablue has required elements missing from the file system (and possibly drivers), e.g.` g_max_fcc = len(glob.glob('/dev/fcc?'))` and `/proc/stb/frontend/fbc/fcc`.

Huevos:
> @littlesat @eriksl Comments please? I know in the past you had a strong opinion about this?

By the way I had a strong negative opinion about this in the past but when you actually see it running it is pretty impressive.

WanWizard:
Like @RobvanderDoes already mentioned, it is only interesting for people watching FTA on the sat, or have a smartcard and softcam without limitations. That group will become smaller and smaller, with everyone being forced to use CI modules. And FBC cable already zaps pretty quickly. 

So we've never seen the point, given the fact it's a massive hack for only a handful of boxes. Even more now that VU+ has disappeared from the face of the earth and Gigablue isn't doing much better (BCM has stopped making SoC's for the OEM market)...

dpeddi:
I would suggest you to to ask if users are interested in it. I know in the past there were a couple of thread in your forum asking for the feature.

Dima73:
Let's apply the patch and see if it works.

RobvanderDoes:
No doubt it'll work.
The real question is if you want to allow a hack like this one (for a limited number of boxes for a limited number of users).

littlesat:
It is not only that it should work or not but the most thing I do not like is that the current service list is copied from python room to cpp room to 'predict' services. Why not just get next precious service by swigging special python functions. Maybe something that can optimize later. Let's put it on the scrum board.
It is a wish for some users and something we missed and also the first merge request that gives a clean maintainable and structured feeling. So I suggest we should go for this attempt and when bugs are found we are helped... and hopefully this will be improved with the suggestion I give.... it would also be nice to add a feature here that it is only valid for fta channels or so...
I advise to merge it this time... this is finally something that has added value (functionality) for at least some users who want it... instead of structural improvement suggestion that creates a lot of work as improvement attempts most time also break something leading to no real add value...

littlesat:
And indeed only vu+ did implement fcc in their drivers and they could improve/fix/optimize some stuff in their drivers for it.

persianpros:
OV compiles FCC support only when "fcc" is present in MACHINE_FEATURES so only Vu+ 4K

littlesat:
That is indeed an option, but you need conditional compile… so another patch… the question is what is most optimal?

I better see eg when encryption is cashed or the cached caid or so is within a blacklist a service is not cached to ensure your card or ci+ module not bothered too much. The code is not ready for this and only really working for fta or when you you c-lines or so that could be ‘overloaded’… 

Most easiest way is at least to have an additional feature that all encrypted channels (via cached caid) are not ‘prezapped’…

I also (maybe I missed it) stop services in prezap when you do not zap for x time…

littlesat:
It now has a conflict, while I wanted to verify and merged it today…

WanWizard:
This better works very well, as with this merge is it now a potential blocker for 8.2-release.

PR #3388, opened 2022-03-01
[Satconfig] sanity check when open tuner setup

Dima73:
File "/usr/lib/enigma2/python/Screens/Satconfig.py", line 809, in
okbuttonClick
AttributeError: 'bool' object has no attribute 'slot'


Taapat:
Why need a check _and current[3]_ at the end if before is check _hasattr(current[3], "slot")_?

Dima73:
> Why need a check _and current[3]_ at the end if before is check _hasattr(current[3], "slot"_

nim = current and len(current) > 2 and hasattr(current[3], "slot")
This bool True or False

Taapat:
Ok sorry, I got it.

PR #3389, opened 2022-03-02
[frontend] add snr/db rs6060 new tuner Zgemma

Dima73:


nautilus7:
How do you know what values/numbers to use in the division?

PR #3404, opened 2022-03-26
Do not close root FBC frontend

Dima73:
-root FBC frontend always applies voltage to the LNB


WanWizard:
Doesn't this have any side-effects?

Dima73:
I do not think. 
On the contrary, it might help here:
https://forums.openpli.org/topic/87814-vu-uno-4kse-unicable-two-satellites-recording-issues/?view=findpost&p=1423714

PR #3432, opened 2022-05-05
Add support for enabling and disabling CI devices (#2283)

catalinii:
*By default every device is enabled
*Allows the user to disable the CI device from the menu (to be used by other software)
*At startup a disabled device will not initialize but will appear in the menu as disabled and the user can choose to enable it

*Note: this is not tested and is a cherry-pick of openatv patches that were tested. At this point I am not sure about the Ci.py changes.

### DIFF
--- a/lib/dvb_ci/dvbci.cpp
+++ b/lib/dvb_ci/dvbci.cpp
@@ -1027,6 +1027,14 @@ PyObject *eDVBCIInterfaces::readCICaIds(int slotid)
 	return 0;
 }
 
+int eDVBCIInterfaces::setCIEnabled(int slotid, bool enabled)
+{
+	eDVBCISlot *slot = getSlot(slotid);
+	if (slot)
+		return slot->setEnabled(enabled);
+	return -1;
+}
+
 int eDVBCIInterfaces::setCIClockRate(int slotid, int rate)
 {
 	singleLock s(m_slot_lock);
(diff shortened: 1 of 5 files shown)


WanWizard:
**>>> Please do not merge until after 8.2-release <<<**

Can you explain what the added benefit of this is? What "other software" needs access to the CI device?

catalinii:
minisatip master has support for CI on systems using openpli. This allows using minisatip access directly to the HW and enigma acting as a satip client.

Anyway the point is that the user has the choice of who owns the HW.

Dima73:
[catalinii](https://github.com/catalinii)
need add patch #include <lib/base/esimpleconfig.h>

Dima73:
[WanWizard](https://github.com/WanWizard)
please apply this patch

WanWizard:
No, not until the 8.2-release branch is frozen, I don't fancy a merge and rebase hell...

catalinii:
I still need to test it as Ci.py is probably not correct.

68foxboris:
| ../../git/lib/dvb_ci/dvbci.cpp:17:10: fatal error: lib/base/esimpleconfig.h: No such file or directory
|    17 | #include <lib/base/esimpleconfig.h> // access to startup config
|       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
| compilation terminated.

Dima73:
new patch not use esimpleconfig

Dima73:
[catalinii](https://github.com/catalinii)

openPLi image not #ifdef __sh__

WanWizard:
> I still need to test it as Ci.py is probably not correct.

What is the status?

catalinii:
I am still working on it, but basically the menu does not refresh correctly

Dima73:
I think this should be applied, then I will correct the necessary.

Dima73:
> I am still working on it, but basically the menu does not refresh correctly

I will fix it later after applying this patch.

Dima73:
Maybe

	def updateState(self, slot):
		self.list = []
		self.slot = 0
		for slot in range(SystemInfo["CommonInterface"]):
			state = eDVBCI_UI.getInstance().getState(slot)
			if state != -1:
				self.slot += 1
				self.appendEntries(slot, state)
		lst = self["entries"]
		lst.list = self.list
		lst.l.setList(self.list)

catalinii:
@Dima73 I finally managed to test this and get the desired result. Feel free to review.

Thanks

PR #3443, opened 2022-05-12
Set power led on on record in standby

Taapat:
GB does not support led blink, so the red led is on during standby recording and does not indicate that recording is in progress.
Therefore, during recording, the power led is also turned on to indicate that recording is in progress just like in normal mode.
On request by Dima73.


WanWizard:
I don't see anything that makes this GB specific. What is the impact of this on other hardware?

Dima73:
This code is only for GB used.

Taapat:
Yes it looks like this is only used by GB.
But even if that's not the case, I don't see anything terrible if the power LED turns off during recording in standby, signaling that something is going on in the receiver.

PR #3463, opened 2022-06-01
AVSwitch - fixed translation in readChoices

ims21:


ims21:
prepared, will be tested yet

PR #3473, opened 2022-06-18
[About] not visible  fp_version is zero

Dima73:


WanWizard:
If I look at getFPVersion(), it returns None if all lookups fail, or the value of the succeeded lookup.

In which cases is 0 not a valid value?

Dima73:
Front processor this is for very old DM/VU+ receivers.

WanWizard:
I get that, but it doesn't answer the question?

Which box returns 0 instead of None, and why is that a problem?

Dima73:
0 is default answer, but 99% of receivers simply don't support front processor updates.
What's the point of showing it?

WanWizard:
If I look at https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Tools/StbHardware.py#L7, the default is clearly `None`, and not `0`. 

You can only get `0` if one of the nodes exist and returns it.

Dima73:
Yes, this node is there, but it is a vestige(rudiment) if the value is 0.

WanWizard:
Odd, it is a dummy you say? Which box is that?

Dima73:
gbquad4k login: root
root@gbquad4k:~# cat /proc/stb/fp/version
0

same vu+solo4k

Dima73:
lunix3-4k login: root
root@lunix3-4k:~# cat /proc/stb/fp/version
cat: can't open '/proc/stb/fp/version': No such file or directory
root@lunix3-4k:~# cat /dev/dbox/fp0
^C
root@lunix3-4k:~# pythin
-sh: pythin: command not found
root@lunix3-4k:~# python
Python 2.7.18 (default, Aug  4 2021, 02:29:18)
[GCC 9.2.0] on linux2
>>> from fcntl import ioctl
>>> fp = open("/dev/dbox/fp0")
>>> ioctl(fp.fileno(), 0)
0
>>>

WanWizard:
Great, more Dags crap... :angry:

PR #3495, opened 2022-07-20
[frontend] update default maxdb value in calculateSignalPercentage

Dima73:
-tnanks openViX team

### DIFF
--- a/lib/dvb/frontend.cpp
+++ b/lib/dvb/frontend.cpp
@@ -902,20 +902,27 @@ static inline uint32_t fe_udiv(uint32_t a, uint32_t b)
 }
 
 int eDVBFrontend::calculateSignalPercentage(int signalqualitydb)
+/*
+		When using new API, signalqualitydb is Carrier-to-Noise-Ratio (CNR) - i.e. SNR at a specific point in the receiver chain
+		Minimum CNR for lock is ~20 dB, not 0 dB as previously assumed
+		Maximum expected CNR is an estimate of typical value of CNR at which limiting could occur in the absence of interference etc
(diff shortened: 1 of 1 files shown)


littlesat:
It is not clear to me why this is needed. The new max levels look far too high. When you make it this way then better introduce a min value instead of increasing the max high value and do not depend on min = 2/3 of max or so….

Dima73:
I noticed a long time ago that the SNR is often 100%, in relation to an incomplete DB, when using a new API.
This applies to many new receivers, especially on HiSilicon.
I think these changes will give a more accurate SNR ratio.
Moreover, this code is used only when there is a DB, but there is no SNR.

That is, the manufacturer can always correct this in the drivers, as he sees fit.

littlesat:
Maybe we should simply show dB only when the api gives dB and no percentage… now we are implementing two ‘estimations’…. SNR in percentage makes actually no sense at all as the dimension is usually always expressed in dB.
The 2/3 factor I do not understand at all….

Dima73:
https://github.com/OpenPLi/enigma2/blob/develop/lib/dvb/frontend.cpp#L1457

Dima73:
/* provide an estimated percentage when drivers lack this info */

littlesat:
This function calculateSignalPercentage...

Mayby better go for the type and directly return a value without going for a max db etc...

Dima73:
The authors(openViX team) has a detailed explanation of the changes, I trust them.
https://github.com/OpenPLi/enigma2/pull/3495/commits/ec6d66fe8fa25530abe1fced2ccbfb357a3c05ae#diff-c9bdd86f4f4fa89ad924eb5b5e5394962c92619d035229fa90fac09b6e2adb38R906

WanWizard:
What is the status of this PR?

Dima73:
Let's put it off for now.

PR #3496, opened 2022-07-22
Add VMSG_TYPE2 for arm box

Dima73:
-need in vtuners  IOCTL
-more info

https://forums.openpli.org/topic/90211-problem-with-dvb-t-usb-rtl2832-on-openpli-82/

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -93,6 +93,10 @@ if test `echo "$BOXTYPE" | cut -b 1-2` == "vu"; then
 	AC_DEFINE(FORCE_ADVANCED_REMOTE, 1,[define to fixup the input device identification when the remote control is actually an 'advanced' remote (with play/forward/rewind keys)])
 fi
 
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -93,6 +93,10 @@ if test `echo "$BOXTYPE" | cut -b 1-2` == "vu"; then
 	AC_DEFINE(FORCE_ADVANCED_REMOTE, 1,[define to fixup the input device identification when the remote control is actually an 'advanced' remote (with play/forward/rewind keys)])
 fi
 
+if test `uname -m | cut -b 1-3` == "arm"; then
```
WanWizard, line 96: Are you not testing the ARCH of the build host, instead of that of the target platform?

Dima73: Yes.
That's my fault.
Need to check  BOXTYPE.

PR #3500, opened 2022-07-23
Added missing Resolution and F4 buttons for uclan's new distributed RC

ims21:
- it could be added to release too


orator89:
Really thanks ;)

PR #3512, opened 2022-08-20
SystemInfo["HasPhysicalLoopthrough"] add AVL6211

Dima73:
-et7500/ et8500 drivers bug
Has_Outputs  != Has_Ouput
NIM Socket 1:
Type: DVB-S2
Name: AVL6211
Frontend_Device: 1
Has_Ouput: yes
I2C_Device: 1


persianpros:
Seems a little ugly, I think you could do it with "and".

WanWizard:
I can live with this, "in" is used elsewhere too, when is a good time to switch between "in" and "and"?

persianpros:
Maybe:

`SystemInfo["HasPhysicalLoopthrough"] = ["Vuplus DVB-S NIM(AVL2108)", "GIGA DVB-S2 NIM (Internal)", "AVL6211"] if model in ("et7500", "et8500") else ["Vuplus DVB-S NIM(AVL2108)", "GIGA DVB-S2 NIM (Internal)"]`

WanWizard:
Repeating the same info is more ugly imho...

PR #3513, opened 2022-08-21
[Standby] add delay 1.5 sec deep standby and CEC

Dima73:


WanWizard:
This is not needed, sorry my own stupidity wasted your time.

PR #3521, opened 2022-09-03
[config.py] Fix: TypeError: '<' not supported between instances of 'str' and 'int'.

fairbird:


Hains:
Question via PM forum.
PR on hold.

fairbird:
I have answered you on PM also !!

PR #3548, opened 2022-10-15
[Directories] add isPluginInstalled

Dima73:


littlesat:
This is the wrong way to do it... So I would not recommend to merge it...
It has a lot of loops etc... This can be done in a much smoother way. And why scan for [".pyc", ".py"]? In the folder there should be a plugin.py*... 

And why is exec(pluginconfig) needed? sounds it does do more than just check if a plugin exists?

Why do you actually need it as you already can get plugin lists in a more easy way...

I think you need it for a reason where the episode list is not read yet.

I suggest the loop stuf etc can be removed totally and we can simply check in one step if the plugin exist especially when we make the type (system plugin, extension, so the sub folder) mandatory we can simply check if a plugin.pyc file (the py should ALWAYS be compiled) exists in that subfolder for the type and the plugin name directory. 

So keep it simple please,,,

Dima73:
Ok.Todo.

PR #3555, opened 2022-10-31
lt.po + all po update 2022-10-31

pr2git:


WanWizard:
Can't merge, conflicts with an earlier PR for bg.po. Can you regenerate?

PR #3563, opened 2022-11-15
idvb.h: add lamedb service flag information

Huevos:


athoik:
Nice! 👍

littlesat:
The comment in the new flag is nog correct... it has nothing to do with the last scanned bouquet list. New services are flagged and these flags can be shown in the service lists.

PR #3571, opened 2022-11-30
[eServiceEvent] update fixUTF8

Dima73:
- add utf8 correction for getShortDescription and getExtendedDescription
- remove fixUTF8 from convertDVBUTF8
 -add setting to enable the this correction

thanks @jbleyel


WanWizard:
In the category "dumb questons": in which case would you choose not to correct something invalid? In other words, why is the config option needed?

Dima73:
This applies to downloading epg from the internet.
You can trace the error.

Next step [Repair utf8](https://github.com/openatv/enigma2/commit/68e21f64622253c524819428dad3d55a6e508004)
When I get confirmation from users that it works.

WanWizard:
Still find it odd that it is configurable, it looks to me like a lot of debug and/or test code.

If something invalid, and it can be fixed, fix it without bothering the user?

Perhaps it is better to wait and see how this ends up, instead of copying ATV's test code?

Dima73:
Ok.Wait.

PR #3572, opened 2022-12-06
Repair utf8

Dima73:
[eServiceEvent]
* add repairUTF8 and enable by default to prevent writing of incorrect UTF8 data for short and extended description in epg cache

thanks @jbleyel


littlesat:
Why should this fix be configurable? (just to add another config or are we not sure about the fix?)

 

I would consider to get this enabled with debug at high level (e.g. 5). Reduces a lot of code and it makes the fix a real fix as intended!

WanWizard:
I asked the same earlier.

PR #3574, opened 2022-12-08
[Satfinder] add compatibility with AutoBouquetsMaker if avable

Dima73:
thanks @Huevos

### DIFF
--- a/lib/python/Plugins/SystemPlugins/Satfinder/plugin.py
+++ b/lib/python/Plugins/SystemPlugins/Satfinder/plugin.py
@@ -1,18 +1,34 @@
 from enigma import eDVBResourceManager, eDVBFrontendParametersSatellite, eDVBFrontendParametersTerrestrial, eTimer
-
-from Screens.ScanSetup import ScanSetup, buildTerTransponder
-from Screens.ServiceScan import ServiceScan
-from Screens.MessageBox import MessageBox
-from Plugins.Plugin import PluginDescriptor
-
-from Components.Sources.FrontendStatus import FrontendStatus
 from Components.ActionMap import ActionMap
-from Components.NimManager import nimmanager, getConfigSatlist
 from Components.config import config, ConfigSelection, getConfigListEntry
+from Components.NimManager import nimmanager, getConfigSatlist
+from Components.Sources.FrontendStatus import FrontendStatus
 from Components.SystemInfo import SystemInfo
 from Components.TuneTest import Tuner
+from Plugins.Plugin import PluginDescriptor
+from Screens.MessageBox import MessageBox
+from Screens.ScanSetup import ScanSetup, buildTerTransponder
+from Screens.Screen import Screen # for services found class
+from Screens.ServiceScan import ServiceScan
 from Tools.Transponder import getChannelNumber, channel2frequency
 from Tools.BoundFunction import boundFunction
+
+try: # for reading the current transport stream (SatfinderExtra)
+	from Plugins.SystemPlugins.AutoBouquetsMaker.scanner import dvbreader
+	dvbreader_available = True
+except ImportError:
+	print("[Satfinder] import dvbreader not available")
+	dvbreader_available = False
+
+if dvbreader_available:
+	from skin import parameters
+	from Components.Sources.StaticText import StaticText
+	from Components.ScrollLabel import ScrollLabel
+	from Components.Label import Label
+	from Tools.Hex2strColor import Hex2strColor
+	import time
+	import datetime
+	import _thread as thread
 
 
 class Satfinder(ScanSetup, ServiceScan):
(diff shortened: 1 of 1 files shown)


WanWizard:
I'm not so happy with this link from core code to an external plugin. What does this do? Why does it need "compatibility"?

/cc @littlesat @Huevos

Dima73:
Add more info(FTA/Radio/TV/sid and etc) after scan for each service.

WanWizard:
If that is really an improvement, why not add the functionality to Enigma?

I can understand VIX does it this way, their focus is the UK market and virtually all their users use ABM, but for us that is not the case, and I don't want people having to install ABM just to get this additional functionality.

Dima73:
Ok.Close(if so, the code is available).

Huevos:
> I'm not so happy with this link from core code to an external plugin. What does this do? Why does it need "compatibility"?
> 

That is exactly what OpenPLi does with its channel importer. A main enigma screen imports channel data via the web interface from a plugin.

> If that is really an improvement, why not add the functionality to Enigma?

It is a huge improvement. It allows you to see the TSID, ONID and OrbPos, (and service list) read direct from the SI tables on the transponder (not satellites.xml) which verifies that you are pointing at the correct transponder. e.g. could be used for alignment. On the screen you see the details updated in real time as soon as the data is read from the transport stream.

Some of the code already exists in enigma, e.g. TsidOnidTest used in AutoDiSEcQ. But that code is not stable in this use. I tried before this was ever written. Writing new code in c++ would be a lot of work for even an experience coder and what is the point when a working solution already exists? If people want functionality in enigma they have to install the dependencies. That is how it has always been. We can't build everything into the base image. We just download extra functionality that we want from the feeds.

This goes back to the new improved OpenPLi channel importer that relies on an external plugin, written after there already was a fully functional (non-PLi) version that didn't rely on any external plugin. In my opinion there just aren't the resources in the Enigma community to sustain this infinite desire to recreate the wheel because our way is better. This rivalry gets us nowhere.

> I can understand VIX does it this way, their focus is the UK market and virtually all their users use ABM, but for us that is not the case, and I don't want people having to install ABM just to get this additional functionality.

Not just for UK. ABM has over 60 providers, but it is true ABM is shipped by default in the OpenViX base image.

This code does not modify Satfinder class at all. It is an extra class that inherits Satfinder class and only if the dvbreader is present. If not present just the Satfinder class is used. This happens at startup. If you look at the patch only lines 16-32, and the lines after 592 belong to the patch. The rest appears to be a lot of cosmetic reformatting.

Dima73:
Big thanks @Huevos
You explained everything correctly.
Yes, this is a good improvement for E2.
There hasn't been anything new lately.
I'm reopening the PR.

littlesat:
I do not think this is an improvement and at least not the way how it should be done. The way how the 'plugin' is 'hoked' to the sat finder is not-done (even when another plugin it is doing approx the same I think this is also not the right method). 

There are also more thinks included in this merge request as just added functionality for ABM. Such as an indent that is changed and some config strings. This should be done e.g. in different merge requests.

Also the ABM plugin itself has a bad structure. E.g. I think the cpp code to 'scan' read the service list from the transponders should be included into the enigma2 cpp core code and instead of write -> read -> modify -> write -> reload in userbouquets/lamedb files I think it should directly be witten inside the database already available within enigma2. That is why I think it could not be in enigma2 in a standard way as the plugin itself needs relative lot of optimizing (we discussed this before and I know it was still a good job based on what the developers known at the time it was created).

Maybe it is a better idea to put all the functionality in the plugin itself... so the new screen SatfinderExtra for example. And then arrange a way that the 'default' satfinder does not appear in the menu anymore instead of putting the screens which belongs to the plugin actually add in the standard enigma2 code.

In short at least the classes should be included in the plugin code itself.

Maybe another method that could be done is remove the satfinder and replace it by the ABM plugin by somehow remove the plugin satfinder.

Huevos:
If the code were in enigma core how would it be a plugin that can be shared?

littlesat:
Easy answer… We have general code in enigma2 to grab data from transponders that more plugins can use…

Huevos:
Yes, but you don't, although you have talked about it for years. And anyway building part of a plugin into your core doesn't help sharing with distros that don't have that core.

littlesat:
I discussed this indeed many times for a longer time. But the developers from the ABM plugin do have the time to adapt it to do it in a more structured way. And my time is limited as well especially related what happened to me approx. 2 years ago.

So it is what it is... 

This is simply not the way how it should be done and therefore I think we should not merge this in the way...

For this patch I recommend to make a function stealing trick in the plugin itself, so it is not mandatory to adapt the current enigma2 code. Also 'ugly'... -or- find a way to replace the complete satfinder thing in the menu.

Huevos:
Littlesat, I don't know what you mean by this. ABM uses one *.so file that is compatible with all images. Everything else is Python.

littlesat:
What I mean is add the code from the so to the enigma2 binary so we have that part or interface that can read data from transponders standard in enigma2 so it can be used for multiple purposes.
In addition instead of reparsing the service lists and lamedb in python, manipulate them and then reload settings is also not the way to go. Better manipulate the database by add/modify services and lamedb in enigma2 itself and at the end do a simple save. Eg cablescan/other scans are doing this.

Huevos:
Why do you need it in your base? It is just an *.so file. Why complicate things?

What happens in Python is nothing to do with the *.so file that this commit wants to use.

PLi is not shy to use *.so files in other plugins.

PR #3595, opened 2023-01-06
[UsageConfig] auto audio - add more value

Dima73:


littlesat:
What providers do use this? Why when how is it needed?

Dima73:
> What providers do use this? Why when how is it needed?

This is in the openATV / openVIX  images, I think they knew what to add.

littlesat:
I’m afraad it does not have Amy effect at al…

WanWizard:
If I read [EN 300 468 - V1.3.1 - Digital Video Broadcasting (DVB) - ETSI ](https://www.etsi.org/deliver/etsi_en/300400_300499/300468/01.03.01_40/en_300468v010301o.pdf) then it clearly states that it follows ISO-639 for language codes, which means they are all 3-letter codes.

Is this perhaps en attempt to deal with incorrectly formatted audio in multi-track IPTV or media files?

ims21:
Nothing news in [V1.17.1](https://www.etsi.org/deliver/etsi_en/300400_300499/300468/01.17.01_20/en_300468v011701a.pdf) ... 24bits (3chars) only ...

littlesat:
Please revert has no added value and even for iptv it cannot have such strings.

WanWizard:
Done.

WanWizard:
p.s. I checked OpenVIX, those codes have been in there since at least 2016 (didn't go further back).

Given it is an OpenPLi fork, chances are we had it too, but we removed it a long time ago.

Dima73:
https://github.com/OpenViX/enigma2/blob/Developer/lib/python/Components/UsageConfig.py#L1038
https://github.com/openatv/enigma2/blob/7.1/lib/python/Components/UsageConfig.py#L1544

WanWizard:
Check the commit history, it's been in there for dogs years...

WanWizard:
It was added by Andy Blackburn in 2013: https://github.com/openatv/enigma2/commit/bb220e81f0a552e16fcdf35b154eb3e81bd9d2c9 without explanation.

littlesat:
And this is not the first time someone tried to get it in while it does not help at all. It does not Harm as you never get these values… long time ago the autolanguage was initiale my investigation and code.

PR #3598, opened 2023-01-08
[pmt] not select audio track narration by default if auto select off

Dima73:
-do not select audio track narration by default if there are other tracks

### DIFF
--- a/lib/dvb/pmt.cpp
+++ b/lib/dvb/pmt.cpp
@@ -478,6 +478,7 @@ int eDVBServicePMTHandler::getProgramInfo(program &program)
 				break;
 			}
 		}
+		int narration = -1;
 		for (i = 0; i < program.audioStreams.size(); i++)
 		{
 			if (program.audioStreams[i].pid == cached_apid_ac3
@@ -505,6 +506,22 @@ int eDVBServicePMTHandler::getProgramInfo(program &program)
 			if (!program.audioStreams[i].language_code.empty())
 			{
 				int x = 1;
+				if (autoaudio_languages.empty() && narration != -2)
+				{
+					std::string audioCurrentLanguage = program.audioStreams[i].language_code;
+					if ((audioCurrentLanguage.find("NAR") != std::string::npos) || (audioCurrentLanguage.find("qad") != std::string::npos))
+						narration = i;
+					else
+					{
+						if (i == 0)
+							narration = -2;
+						else if (narration != -1 && i > (unsigned)narration)
+						{
+							program.defaultAudioStream = i;
+							narration = -1;
+						}
+					}
+				}
 				for (std::vector<std::string>::iterator it = autoaudio_languages.begin();x <= autoaudio_level && it != autoaudio_languages.end();x++,it++)
 				{
 					bool languageFound = false;


littlesat:
I do not understand when/why this is needed? What does it solve?

littlesat:
NAR and qad should not be hardcoded values here and sorry as we do not have an example or problem description I do not understand the need….
I understand it are values for standaard/original language so I Twinkly we should Challenge the list of original language stuff we have in python.

Dima73:
> I do not understand the need….

When the narration audio is the first to be listed in the SI table.
You need it for people that are too lazy to set their preferred audio.

AbuBaniaz:
Official receivers and most 3rd party receivers do not automatically select the Narrative audio. Only in enigma2 is this being done.

Vast majority of users do not require narrative audio.

This has nothing to do with laziness. Those are just obstructive posts that someone makes to prevent suggestions I make. The same person actually set the default language in Vix and had then turned round and implied by not preferring NAR, we are not in favour of accessibility. If these objective comments are genuine, you would have thought that even subtitles should be defaulted on. 

Even the wording in the audio menu is misleading, set audio per service implies you have to set per service whereas it was said that it allows to override preferred audio for specific services.

The whole point of this request is that it is one less optional option for people to set. Vast majority of users will be single language users. Not that a survey has been done, I'm guessing.

littlesat:
I think we should not do this. Hard coded NAR for lazy users so put a setting as a kind of default and only do it for NAR is what we should not do... this looks exceptional and hacky sorry!

Dima73:
OK.
It's not a problem for me.
I was asked to find a solution, I did it.

Huevos:
> Official receivers and most 3rd party receivers do not automatically select the Narrative audio. Only in enigma2 is this being done.
> 
> Vast majority of users do not require narrative audio.
> 
> This has nothing to do with laziness. Those are just obstructive posts that someone makes to prevent suggestions I make. The same person actually set the default language in Vix and had then turned round and implied by not preferring NAR, we are not in favour of accessibility. If these objective comments are genuine, you would have thought that even subtitles should be defaulted on.
> 
> Even the wording in the audio menu is misleading, set audio per service implies you have to set per service whereas it was said that it allows to override preferred audio for specific services.
> 
> The whole point of this request is that it is one less optional option for people to set. Vast majority of users will be single language users. Not that a survey has been done, I'm guessing.

Nothing was done with NAR in OpenViX. No nasty cpp hacks.

wedebe:
> > Official receivers and most 3rd party receivers do not automatically select the Narrative audio. Only in enigma2 is this being done.
> > Vast majority of users do not require narrative audio.
> > This has nothing to do with laziness. Those are just obstructive posts that someone makes to prevent suggestions I make. The same person actually set the default language in Vix and had then turned round and implied by not preferring NAR, we are not in favour of accessibility. If these objective comments are genuine, you would have thought that even subtitles should be defaulted on.
> > Even the wording in the audio menu is misleading, set audio per service implies you have to set per service whereas it was said that it allows to override preferred audio for specific services.
> > The whole point of this request is that it is one less optional option for people to set. Vast majority of users will be single language users. Not that a survey has been done, I'm guessing.
> 
> Nothing was done with NAR in OpenViX. No nasty cpp hacks.

Does this mean that in OpenViX that NAR isn't selected by default when it occurs as the first available audio track?

Huevos:
In OpenViX Eng is the default 1st language in python.

PR #3599, opened 2023-01-10
[Satfinder] ServiceScan is not a parent of satfinder

Huevos:
Wrongly introduced here 10 years ago: 
https://github.com/OpenPLi/enigma2/commit/9805b85ac24309592bf795e49e88527f560ab5ce


littlesat:
Are you 100% sure it is not needed as it somehow 'borrows' functions from that class!

Huevos:
Like what? It was put there by accident because it is used later in self.session.open(ServiceScan).

Huevos:
I removed it here: https://github.com/OpenViX/enigma2/commit/84ac177b152ddf1792a7fd112fa6af7776d3d990 and no side effects. Absolutely certain it was there by accident.

PR #3601, opened 2023-01-14
[ServiceInfo] add "Codec & lang" in IPTV 4097/5001/5002

Dima73:


littlesat:
Now subList does not get an value as you remover that line. Was this intended?

Dima73:
> Now subList does not get an value as you remover that line. Was this intended?

This value 'subList = self.getSubtitleList()' not used anywhere.

lib/python/Screens/ServiceInfo.py
```diff
@@ -167,14 +167,18 @@ def ShowServiceInformation(self):
...
-				if self.audio:
-					self.numberofTracks = hasattr(self.audio, "getNumberOfTracks") and self.audio.getNumberOfTracks() or 0
-					if self.numberofTracks:
-						currentTrack = self.audio.getCurrentTrack()
-						for i in range(0, self.numberofTracks):
-							audioDesc = self.audio.getTrackInfo(i).getDescription()
-							audioLang = self.audio.getTrackInfo(i).getLanguage()
+				self.IPTV = True
+				self["key_blue"] = self["blue"] = Label("")
+				self["key_yellow"] = self["yellow"] = Label("")
```
Huevos, line 172: Should not overwrite the source object.

Should be:
`self["key_blue"].text = self["blue"].text = ""`

Dima73: Please test this self["key_blue"].text = self["blue"].text = "" and see result.

Huevos: Has been working in OpenViX since forever.

https://github.com/OpenViX/enigma2/blob/Developer/lib/python/Screens/ServiceInfo.py#L169

PR #3603, opened 2023-01-14
Get rid of getConfigListEntry()

Hains:
No need to check the number of arguments when they are fixed in the function call.
Assertion is disabled in OpenPLi oe-core as well.

Remain all brackets to create a tuple with the arguments. Except Autodiseqc.py since it has one argument.

### DIFF
--- a/lib/python/Components/config.py
+++ b/lib/python/Components/config.py
@@ -1915,7 +1915,7 @@ configfile = ConfigFile()
 configfile.load()
 
 
--def getConfigListEntry(*args):
+def getConfigListEntry(*args):
 	assert len(args) > 0, "getConfigListEntry needs a minimum of one argument (descr, configElement)"
 	return args
 
--- a/lib/python/Components/config.py
+++ b/lib/python/Components/config.py
@@ -1915,6 +1915,11 @@ configfile = ConfigFile()
 configfile.load()
 
 
+-def getConfigListEntry(*args):
+	assert len(args) > 0, "getConfigListEntry needs a minimum of one argument (descr, configElement)"
+	return args
+
+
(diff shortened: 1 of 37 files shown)


Hains:
PR for review

Taapat:
Please leave the getConfigListEntry function in config for compatibility.
I think many plugins use this function.

I agree remove getConfigListEntry in enigma code, but the function should be left.

Taapat:
In addition, why leave these many appends for list?
Create list immediately or use extend.
See how I did it in my plugin: https://github.com/Taapat/enigma2-plugin-youtube/commit/5fcea3cd920599beb8c9366fbf894f61790b8ac1

Taapat:
And please fix invalid syntax: https://github.com/OpenPLi/enigma2/actions/runs/3919504702/jobs/6700550930#step:8:3501
https://github.com/OpenPLi/enigma2/actions/runs/3919504702/jobs/6700550930#step:8:3462

Hains:
> In addition, why leave these many appends for list? Create list immediately or use extend. See how I did it in my plugin: [Taapat/enigma2-plugin-youtube@5fcea3c](https://github.com/Taapat/enigma2-plugin-youtube/commit/5fcea3cd920599beb8c9366fbf894f61790b8ac1)

+1
But I think better to do that in another commit.

Taapat:
> +1
> But I think better to do that in another commit.

I agree.

Huevos:
Do not delete the method from config.py or you are going to break every plugin.

Hains:
https://github.com/OpenPLi/enigma2/commit/dc5413b1fd841960672871138bc6f83066932903

getConfigListEntry() not deleted.

PR #3608, opened 2023-02-01
- added icons for menus base on their entryID

DimitarCC:
A simple solution to be possible icons to be added to menues.
Currently if there is no image/icon present it take the parent menu icon or a default icon.
Supports svg and png for icons.

### DIFF
--- a/lib/python/Components/PluginComponent.py
+++ b/lib/python/Components/PluginComponent.py
@@ -137,6 +137,13 @@ class PluginComponent:
 			res += p(menuid)
 		return res
 
+	def getDescriptionForMenuEntryID(self, menuid, entryid ):
+		for p in self.getPlugins(PluginDescriptor.WHERE_MENU):
+			if p(menuid) and isinstance(p(menuid), (list,tuple)):
+				if p(menuid)[0][2] == entryid:
+					return p.description
+		return None
+
 	def clearPluginList(self):
 		self.pluginList = []
 		self.plugins = {}
--- a/lib/python/Screens/Menu.py
+++ b/lib/python/Screens/Menu.py
@@ -12,7 +12,7 @@ from Components.SystemInfo import SystemInfo
 from Components.Pixmap import Pixmap
 from Tools.BoundFunction import boundFunction
 from Plugins.Plugin import PluginDescriptor
-from Tools.Directories import resolveFilename,  fileExists, SCOPE_SKIN, SCOPE_CURRENT_SKIN
+from Tools.Directories import resolveFilename, SCOPE_SKINS, SCOPE_GUISKIN
 from Tools.LoadPixmap import LoadPixmap
 from enigma import eTimer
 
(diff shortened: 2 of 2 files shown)


Taapat:
You don't need to check if a png or svg image exists.
You only need to specify one of them.
resolveFilename return svg if available, if not, png if available: https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Tools/Directories.py#L113

SCOPE_CURRENT_SKIN is specified on Deprecated scopes: https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Tools/Directories.py#L40
It would be better to use SCOPE_GUISKIN instead

DimitarCC:
@Taapat Thanks. I will review the code and update this PR in next few hours.

DimitarCC:
I changed the code but that PR dont want to update. I dont know what is wrong :(

jbleyel:
Can you please add the author credits if you copy code from other distros?
Thank you.

DimitarCC:
Well yes sure i can. But just to note that the idea is gotten from another distro...But the actual implementation is not the same since in that distro it was working a bit different way and it was buggy. So it is not just a copy.

But when we are speaking about that where should we put those credits? under the specific function? What about if the code is not the same?

littlesat:
Why is this needed? The menu picons could always be added to the menu depending on the skin used.

DimitarCC:
Yes but you add them in a separate panel in the screen (as far as i am ware of) and is not exactly in the menu listbox.
At least for me is convinient when i can see the icon right a way. like this:
image

And in the skin is just defined like this:
`
<widget source="menu" render="Listbox" position="72,255" size="564,545" scrollbarMode="showNever" enableWrapAround="1" transparent="1" selectionPixmap="selections/sel_564x60.png">
			<convert type="TemplatedMultiContent">
						{"template": [ MultiContentEntryText(pos = (97,5),size = (435,45),flags = RT_HALIGN_LEFT|RT_VALIGN_CENTER,text = 0),
									   MultiContentEntryPixmapAlphaBlend(pos = (32,5),size = (48,48), png = 5)],
						"fonts": [gFont("Regular",34)],
						"itemHeight": 60
						}
					</convert>
		 </widget>
`
For me seems more easy ;)

jbleyel:
What you mean with buggy?
It’s perfectly working in ATV.

DimitarCC:
I don't know how it works in OpenATV(i didn't even try OpenATV). I got the idea from DreamOS and there is buggy.
And what i mean with "buggy" is that it is choosing random icon (based on the last selection which can be not exactly the parent item) for items without existing icon.

jbleyel:
Then you should take a look to the menu.py from ATV and don’t care about DreamOS closed source half baked implementations.

DimitarCC:
And why i should look there?  We are in OpenPLi and even if OpenATV has a similar code what's the point of taking it from there if my solution works?
I think if OpenPLi guys didn't take it over from OpenATV or another distro it is for a reason (maybe).
And they can decide to accept my PR or not. In general i dont need to look in other distros implementation so to offer my one ;)

ims21:
> Why is this needed? The menu picons could always be added to the menu depending on the skin used.

+1 !!!

DimitarCC:
Nobody is forced to use it!!! People have different wishes and preferences. If you dont define it in the template of the menu (skin) it can work like it is now.
But people that want this currently have no choise.

DimitarCC:
> Then you should take a look to the menu.py from ATV and don’t care about DreamOS closed source half baked implementations.

and to clear that.... i have just lokked in OpenATV.... that approach..
`def okbuttonClick(self):
                global lastKey
		self.resetNumberKey()
		current = self["menu"].getCurrent()
		if current and current[WIDGET_MODULE]:
			lastKey = current[WIDGET_KEY]
			current[WIDGET_MODULE]()`
in my oppinion is not error free because when you click ok you take the last menu that you clicked OK on but that could be different from the parent menu. Especially if you have missing icon in the root of the menu.
If you have icons for all entries is not a problem but if you dont have one for some entry it may take wrong icon for it.
Besides that implementation in OpenATV is completly the same like in DreamOS ;)

littlesat:
DreamOS is closed source they do not want to things over by us and I still do not see the added value. Instead maybe a standard renderers might be considered skins can use even they might possible to add a parent icon when some does not exist.

DimitarCC:
That part is in their opensource part of the code. Not everything is closed source. Only enigma2 core is closed there.
So it can be taken over.
Regarding is there value or not...Its subjective....
And how you will put standard renderer inside the menu items?
I understand your point but you are implying putting the icons outside of the list items which is not what was my target.

littlesat:
It might be open source but as far I know it is not under open license. It is use for dreambox hardware only.

DimitarCC:
But well the code i propose is anyway not the same is only based on the idea of it.
Since is also used in OpenATV we can consider it as open.

DimitarCC:
P.S. guys i shared the PR because i thought it will be nice to have. If is such a big issue...just close it and thats it. No need for such a conversation. ;)

littlesat:
Maybach better to discus it on the forum before implementing it. Then you have an idea of it is wanted before apending energy in the implementation and Maybach al the comments even lead to better code/implementations

DimitarCC:
Maybe... But i dont get much answers to my questions in the forum. Espetialy technical ones... So i doubt there will be much of a change.
Well it seems that lead us nowhere so i close this PR :)

PR #3611, opened 2023-02-08
[NimManager] force save unicable scfrequency value

Dima73:
more info:
https://forums.openpli.org/topic/93703-sf8008-bug-in-unicable-settings/

### DIFF
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -1394,7 +1394,11 @@ def InitNimManager(nimmgr, update_slots=[]):
 
 				def scrListChanged(productparameters, srcfrequencylist, configEntry):
 					section.format = ConfigSelection(["unicable", "jess"], default=getformat(productparameters.get("format", "unicable"), configEntry.index))
-					section.scrfrequency = ConfigInteger(default=int(srcfrequencylist[configEntry.index]))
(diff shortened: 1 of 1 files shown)


lib/python/Components/NimManager.py
```diff
@@ -1394,7 +1394,11 @@ def positionsChanged(configEntry):
 
 				def scrListChanged(productparameters, srcfrequencylist, configEntry):
 					section.format = ConfigSelection(["unicable", "jess"], default=getformat(productparameters.get("format", "unicable"), configEntry.index))
-					section.scrfrequency = ConfigInteger(default=int(srcfrequencylist[configEntry.index]))
+					default_value = int(srcfrequencylist[configEntry.index])
+					section.scrfrequency = ConfigInteger(default=default_value)
+					section.scrfrequency.value = default_value
+					section.scrfrequency.save()
+					section.scrfrequency.save_forced = True
```
Huevos, line 1401: Why are you setting "save_force" after calling "save"?
https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Components/config.py#L90

Dima73: https://github.com/OpenPLi/enigma2/pull/3612

PR #3615, opened 2023-02-16
Skin engine extensions and Visual refinements

DimitarCC:
When use iconOffset for TimerList make the offset from both sides or entries will be not centered
Also skin engine extension to support connected widgets.

### DIFF
--- a/data/keymap.xml
+++ b/data/keymap.xml
@@ -221,6 +221,8 @@
 		<key id="KEY_ESC" mapto="cancel" flags="m" />
 		<key id="KEY_LEFT" mapto="left" flags="mr" />
 		<key id="KEY_RIGHT" mapto="right" flags="mr" />
+		<key id="KEY_UP" mapto="up" flags="mr" />
+		<key id="KEY_DOWN" mapto="down" flags="mr" />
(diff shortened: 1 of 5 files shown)


DimitarCC:
ConfigList changes are removed, because this can be handled in the component that override bindings.
But have to keep keymap.xml additions since we need all actions to be available in the map

DimitarCC:
Here i introduce an extension for skinning so to support Renderer to be possible to relate to another widget.
The idea is extending the skinning engine to support attaching renderer to any screen gui component (not only to Sources).

I propose this because currently when you use renderer you have to connect it to specific source and there is no chance to have linked/connected visual components without it.

Implementation of Source however have some requirements that in my opinion limits users when they want completely custom functionality.

 
configwithcustomscrollbar


All this is in connection to my skin and my wish to attach a custom pager to listbox instead of scroller, but it can be used for any other functionality that somebody can wish.

In the renderer you can do whatever the skinner wants. There are few key methods that can be overwritten if needed:
```
def onContainerShown(self):
	pass
	
def bindKeys(self, container):
	pass
```

PR #3616, opened 2023-02-16
[Navigation] save to variable config.misc.prev_wakeup_time.value

Dima73:
-needed for some plugins that use wakeup timer
-just saves the first value of the until new start E2
-then we can compare this value with the plugin's wakeup time
-determining who wakeup the receiver

### DIFF
--- a/lib/python/Navigation.py
+++ b/lib/python/Navigation.py
@@ -41,13 +41,14 @@ class Navigation:
 		self.RecordTimer = RecordTimer.RecordTimer()
 		self.__wasTimerWakeup = getFPWasTimerWakeup()
 		self.__isRestartUI = config.misc.RestartUI.value
+		self.__prevWakeupTime = config.misc.prev_wakeup_time.value
 		startup_to_standby = config.usage.startup_to_standby.value
 		wakeup_time_type = config.misc.prev_wakeup_time_type.value
 		self.wakeup_timer_enabled = False
 		if config.usage.remote_fallback_import_restart.value:
 			ImportChannels()
 		if self.__wasTimerWakeup:
-			self.wakeup_timer_enabled = wakeup_time_type == 3 and config.misc.prev_wakeup_time.value
+			self.wakeup_timer_enabled = wakeup_time_type == 3 and self.__prevWakeupTime
 			if not self.wakeup_timer_enabled:
 				RecordTimer.RecordTimerEntry.setWasInDeepStandby()
 		if config.misc.RestartUI.value:
@@ -57,12 +58,12 @@ class Navigation:
 		else:
 			if config.usage.remote_fallback_import.value and not config.usage.remote_fallback_import_restart.value:
 				ImportChannels()
-			if startup_to_standby == "yes" or self.__wasTimerWakeup and config.misc.prev_wakeup_time.value and (wakeup_time_type == 0 or wakeup_time_type == 1 or (wakeup_time_type == 3 and startup_to_standby == "except")):
+			if startup_to_standby == "yes" or self.__wasTimerWakeup and self.__prevWakeupTime and (wakeup_time_type == 0 or wakeup_time_type == 1 or (wakeup_time_type == 3 and startup_to_standby == "except")):
 				if not Screens.Standby.inTryQuitMainloop:
 					self.standbytimer = eTimer()
 					self.standbytimer.callback.append(self.gotostandby)
 					self.standbytimer.start(15000, True) # Time increse 15 second for standby.
-		if config.misc.prev_wakeup_time.value:
+		if self.__prevWakeupTime:
 			config.misc.prev_wakeup_time.value = 0
 			config.misc.prev_wakeup_time.save()
 			configfile.save()
@@ -76,6 +77,9 @@ class Navigation:
 
 	def isRestartUI(self):
 		return self.__isRestartUI
+
+	def prevWakeupTime(self):
+		return self.__prevWakeupTime
 
 	def dispatchEvent(self, i):
 		for x in self.event:


littlesat:
Why is this needed. Can we also consider to adapt the plugins?

Dima73:
Look, after reading the current value, we reset it to zero.
https://github.com/OpenPLi/enigma2/blob/9c293526e7035980822388f6510df8b6d45c566f/lib/python/Navigation.py#L67
My patch just saves the first value of the until new start E2.
Without breaking anything.

Dima73:
Then we can compare this value with the plugin's wakeup time.
Determining who wakeup the receiver.

Huevos:
> 



> Then we can compare this value with the plugin's wakeup time. Determining who wakeup the receiver.

You should be doing this in the plugin.

In the plugin on session start check session.nav.wasTimerWakeup().
Then in the same function the plugin should check its own wake up time.

For example see here:
https://github.com/OpenViX/enigma2/blob/73072f51afb9132cbacc02336015325030979125/lib/python/Plugins/SystemPlugins/OpentvZapper/opentv_zapper.py#L877

Then once the plugin finishes you recheck that variable and if necessary put the box back in deep:
https://github.com/OpenViX/enigma2/blob/73072f51afb9132cbacc02336015325030979125/lib/python/Plugins/SystemPlugins/OpentvZapper/opentv_zapper.py#L1014

Dima73:
Thank you, according to the idea, wasTimerWakeup should be config.misc.prev_wakeup_time.value.
And if not?
That's why I want to store config.misc.prev_wakeup_time.value value in a variable so that I can check for sure.

Huevos:
wasTimerWakeup is a Bool. It does not need the wake up time because it is set at session start.

It should be done in the plugin. If you introduce a new variable in OpenPLi and then modify the plugin for OpenPLi new variable that will break the plugin for everyone that is not OpenPLi.

Dima73:
don't understand what is the dispute about?

Many years ago, I entered this config myself and decided to save it by default at the end of checking all conditions.
https://github.com/OpenPLi/enigma2/commit/ef91a7ff5306e47edf1a4d07b90afd7b3af10c20
https://github.com/OpenPLi/enigma2/commit/ef91a7ff5306e47edf1a4d07b90afd7b3af10c20#diff-2b8bbb4b70a55dacba61893410d205b0decd408994d0d75bb9566d7e30722412R49

Now I need a working variable, not a default value.

I'm not going to change anything in the plugin that will break its compatibility with other images.

What is the problem?
Why are we wasting time on this argument?

Huevos:
Are you going to change anything in the plugin?

Will the changes in the plugin make the plugin work diferent in OpenPLi than other images?

If yes then the changes to the plugin are wrong.

Dima73:
Rave.
How does this PR harm other images?

Huevos:
In PLi you can do what you like, but if you are intending to make changes in a plugin to use this variable which is not present in other distros that is wrong.

ims21:
> In PLi you can do what you like, but if you are intending to make changes in a plugin to use this variable which is not present in other distros that is wrong.

Nonsense ... If he does it in Pli and the plugin for Pli too, it's right ...

Huevos:
> > In PLi you can do what you like, but if you are intending to make changes in a plugin to use this variable which is not present in other distros that is wrong.
> 
> Nonsense ... If he does it in Pli and the plugin for Pli too, it's right ...

This is the same arrogant condescending attitude PLi had 10 years ago before starting cooperating with other teams.

The plugins are shared resources of the Enigma community irrespective of where they are hosted and who has what permissions. The codebase is contributed to by everyone, not just the repo admins.

This is just the same with our plugins where we go out of our way to ensure compatibility with all images.

Dima73:
I'm not going to change anything in the plugin that will break its compatibility with other images.

ims21:
> > > In PLi you can do what you like, but if you are intending to make changes in a plugin to use this variable which is not present in other distros that is wrong.
> > 
> > 
> > Nonsense ... If he does it in Pli and the plugin for Pli too, it's right ...
> 
> This is the same arrogant condescending attitude PLi had 10 years ago before starting cooperating with other teams.
> 
> The plugins are shared resources of the Enigma community irrespective of where they are hosted and who has what permissions. The codebase is contributed to by everyone, not just the repo admins.
> 
> This is just the same with our plugins where we go out of our way to ensure compatibility with all images.

As far as I know, each good repo has its own version of plugins, so don't write anything about arrogance here...

littlesat:
As far as I know, each good repo has its own version of plugins, so don't write anything about arrogance here... -> +1...

This is the same arrogant condescending attitude PLi had 10 years ago before starting cooperating with other teams. -> This is not arrogance.. this is just trying to keep our code as clean as possible.

Changing or add 'dummy' style things for just getting a plugin working is something we should not do. Only when a plugin needs something with real added value behind it should be considered to adapt. As far I can see adding this config does not have a real added value... so e.g. the plugin can check if a required config does not exist already and then create it.

Dima73:
All this is very strange.
Argument for the sake of argument.

If this PR were applied, then I would show how you can really (absolutely accurately) find out which plug-in automatically wake up the receiver.
By the way, this config config.misc.prev_wakeup_time , just for this, was made by me many years ago.

Well, okay.

littlesat:
Sorry I still do not understand the real added value...

Dima73:
Look at this topic, request ims
https://forums.openpli.org/topic/93843-epgrefresh-and-deep-standby/?view=findpost&p=1519167

Here is a solution
https://github.com/opendreambox/enigma2-plugins/blob/master/epgrefresh/src/EPGRefresh.py#L529

We need this config config.misc.prev_wakeup_time.value

Understand.
I myself once added this config to openPli.
But for some reason I decided to reset it to zero after the start of E2.
Please don't argue, just take a close look at the code and you'll see that I'm just storing the first value before resetting to zero.

You can just delete these  lines
`		if config.misc.prev_wakeup_time.value:
			config.misc.prev_wakeup_time.value = 0
			config.misc.prev_wakeup_time.save()
			configfile.save()`

littlesat:
I still do not understand why it is needed. What it is doing and why a plugin should need it.... aka a plugin can also add configs that are note there when they are needed.

Dima73:
This need for [[epgrefresh] expand option "Shutdown after EPG refresh"](https://github.com/OpenPLi/enigma2-plugins/commit/a8536961c653c5218fbf9157b8f5f4f01c1591d9)
And in the future for other plugins, for example EPGimport.

littlesat:
I still recommend to go for huevos’s suggestion!!!!
Could be I missed it in the epg refresh patch because it was overloaded with translation changes.

Dima73:
It's some kind of circus.
Have you even looked carefully at the code that I wanted to change?
One added fucking variable!!
After all, this does not affect anything in enigma2, but for plugins it can be of great importance.
For those who want to really develop E2 in the future, this would be a good example.
How to find out exactly who wake up the receiver.

Huevos:
> It's some kind of circus. Have you even looked carefully at the code that I wanted to change? One added fucking variable!! After all, this does not affect anything in enigma2, but for plugins it can be of great importance. For those who want to really develop E2 in the future, this would be a good example. How to find out exactly who wake up the receiver.

So why don't you save that then, instead of this?

Dima73:
What's the difference?
Why is this option bad?

littlesat:
The differente is for what you want to do is a more structuren methode here nothing nerds to be added at the enigma2 code….the extra config is not mandatory at all… and that is why we should not do it.

n the plugin on session start check session.nav.wasTimerWakeup().
Then in the same function the plugin should check its own wake up time.

For example see here:
https://github.com/OpenViX/enigma2/blob/73072f51afb9132cbacc02336015325030979125/lib/python/Plugins/SystemPlugins/OpentvZapper/opentv_zapper.py#L877

Then once the plugin finishes you recheck that variable and if necessary put the box back in deep:
https://github.com/OpenViX/enigma2/blob/73072f51afb9132cbacc02336015325030979125/lib/python/Plugins/SystemPlugins/OpentvZapper/opentv_zapper.py#L1014

That is why I think adding the config does not have added value as the plug-in can do the need on it’s own.

PR #3624, opened 2023-02-28
[ActionMap] sanity check when contexts not list

Dima73:


Taapat:
If this is really necessary, then maybe it would be better:
`if contexts and not isinstance(contexts, list):`
Then anything that is not a list, not just a string, would be converted to a list.

Dima73:
#>>> contexts=None
>>> isinstance(contexts, list)
False
>>> contexts = [contexts]
>>> contexts
[None]#
 this not good

littlesat:
Why is this sanity check needed?

Dima73:
E.g. 
[ActionMap] Keymap(s) 'I, n, f, o, b, a, r, V, m, o, d, e, B, u, t, t, o, n, A, c, t, i, o, n, s' -> Undefined action(s) 'vmodeSelection'.


Same as HelpableActionMap
https://github.com/OpenPLi/enigma2/blob/2bc264b922cb3c73a87e94ab1385bb465491888b/lib/python/Components/ActionMap.py#L103

ims21:
If there will be  'I, n, f, o, b, a, r, V, m, o, d, e, B, u, t, t, o, n, A, c, t, i, o, n, s', then it is wrong conversion in code to py3 and then will be much better, when in this case will be enigma crashing, imho

Dima73:
> If there will be 'I, n, f, o, b, a, r, V, m, o, d, e, B, u, t, t, o, n, A, c, t, i, o, n, s', then it is wrong conversion in code to py3 and then will be much better, when in this case will be enigma crashing, imho

ok, I agree

PR #3628, opened 2023-03-04
[Network] completely rewrite the DNS name server code

Dima73:
-now you can use two options from openresolv or the old way


WanWizard:
Thanks !

PR #3629, opened 2023-03-04
[Network] update DNS setup and save

Taapat:
Remove openresolv removing and code to work without it.
Therefore don't write anything in the file /etc/resolv.conf.
Use this file only to get DHCP nameservs.

Also code style fixes.


neo-pli-bsps:
I think the resolv..conf  part isn't complete.

you can only add IP's you find to `self.nameservers` if that IP is not defined in `ifaces`, since fixed DNS entries will also be written to it.

Taapat:
Actually, with DNs there is a conflict.
As I understand it, the changes made by Dima73 at the request of WanWizard are to show and set DNS individually for each device.
Therefore, if DNS is specified for an individual device, only that is displayed and nameservers from resolv.conf is not used.
In fact, the receiver uses DNS globally, regardless of the device, and it would be correct to show only nameservers from resolv.conf, but then you will not be able to set dns individually for each device.

neo-pli-bsps:
Yes, and that is not correct.

What should be displayed at the bottom of the interface screen as "primary" and "secondary" are the active DNS IP's, and should come from resolv.conf, not from whatever is configured.

Which is why I understand the original code used DNS info in ifaces for configuration, and self.nameservers for display. Which I think Dima73 broke in his changes. The two sets of DNS info are and should be completely different in the code.

Taapat:
Check this: https://github.com/OpenPLi/enigma2/pull/3632

PR #3632, opened 2023-03-05
Add RunTimeNameservers screen in Network

Taapat:
In NameserverSetup shows and allows to configure only adapter static nameservers. Run-time nameservers are shown in a separate screen RunTimeNameservers, and if necessary offer to add them in adapter static nameservers.

Shows real nameservers from run-time nameservers for information.

### DIFF
--- a/lib/python/Components/Network.py
+++ b/lib/python/Components/Network.py
@@ -326,6 +326,15 @@ class Network:
 		print("[Network] get %s nameserver list: %s" % (iface, servers))
 		return servers
 
+	def getIfaceNameservers(self, iface):
+		dns_servers = self.getAdapterAttribute(iface, "dns-nameservers") or []
+		servers = dns_servers[:]
+		return servers
+
+	def getNameservers(self):
+		servers = self.nameservers[:]
+		return servers
+
 	def clearNameservers(self):
 		self.nameservers = []
 
--- a/lib/python/Screens/NetworkSetup.py
+++ b/lib/python/Screens/NetworkSetup.py
@@ -200,29 +200,28 @@ class NameserverSetup(ConfigListScreen, HelpableScreen, Screen):
 	def __init__(self, session, iface=None):
 		Screen.__init__(self, session)
 		HelpableScreen.__init__(self)
-		self.setTitle(_("Configure nameservers"))
+		self.title = _("Configure nameservers")
+		self.session = session
 		self.iface = iface
-		self.backupNameserverList = iNetwork.getNameserverList(iface=self.iface)
-		print("[NameserverSetup] backup-list:", self.backupNameserverList)
 
 		self["key_red"] = StaticText(_("Cancel"))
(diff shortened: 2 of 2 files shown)


neo-pli-bsps:
I'm not sure what you've done here, I'll have to build and see. I don't see why a separate screen is needed, the info at the bottom of the interface screen is (or should be) already runtime information?

p.s. "Runtime" is one word, not "run-time". And I would like to suggest to use "Active" instead of "Runtime", as a better description of what the values are.

Taapat:
In NameserverSetup screen there is no widget that displays runtime information.
In skin, of course, you can add your own widgets that will display additional information.

This additional screen is called in NameserverSetup screen with the blue button and you can add nameservers in adapter static nameservers from it if you want. Although I can't imagine who needs it.

I agree that run-time is not a good term, "Active" looks better.
I took run-time from the resolvconf man page: https://manpages.ubuntu.com/manpages/trusty/man8/resolvconf.8.html

Taapat:
These changes are for the NameserverSetup screen.
This is the screen for setting nameservers for the adapter.
If you use DHCP, then the adapter does not have static nameservers to set, and it would be misleading to show active nameservers here.
If no nameserver is shown, it can also cause a misunderstanding.
Therefore, in this case, I chose not to show any nameservers in the NameserverSetup screen, but to add a warning and the option to view active nameservers in a separate screen.

neo-pli-bsps:
As far as I understood, the original suggestion ( = WanWizards plan) was to completely remove that nameserver screen, and add the 2 fields for static nameservers to the interface screen, where they belong. At the bottom of the interface screen there is a block with runtime data, including IP, gateway and active primary and secondary DNS.

Which imho is a much better solution than the current hacks.

Dima73:
for info
dhcp nameservers  https://github.com/openatv/enigma2/blob/master/lib/python/Screens/NetworkSetup.py#L373

Taapat:
After the Dima73 changes, I think he has made everything as Wanwizard wanted. I don't like the result and I offer to change it.

I found three discussions in forum. And after reading it all again, I now found the Wanwizards plan you mentioned.
And I see that it was another, without NameserverSetup Screen. It sounds better for me too.

The question is whether we are realizing the original plan, or there is something that I will not know, and need a separate nameserversetup screen?

Dima73:
By the way, just for information, my install/uninstall openresolv option worked great:
see code https://github.com/OpenPLi/enigma2/commit/92380300750da897a44d78df1fd8910c4e6ec451#diff-1c30cb205b11cb23be8fe654de7e7ddb0305fed569759ab8cae8d4ffb6d34a06R800
so if not openresolv use writeNamesver as before.

Taapat:
Quote from the forum:
The hardcoded resolv.conf (which is per machine) is a 20+ year old solution from the time one interface was a luxury.

Taapat:
I hope this is WanWizards plan: https://github.com/OpenPLi/enigma2/pull/3634

PR #3639, opened 2023-03-08
Clean up Pager.py

DimitarCC:


DimitarCC:
A complete new solution is pending so i close that.

PR #3645, opened 2023-03-11
[Pager] Various fixes

Huevos:
Various commits. Please don't squash.

Fixes for Pager.

Remove redundant code.

Fix Makefiles and configure.ac.


Taapat:
Maybe you could also add applySkinFactor to the default values.
Then they could not to be specify in different resolutions skin.

Huevos:
Yes, good idea. But can this pull be accepted first?

PR #3654, opened 2023-03-14
[AdapterSetup] delete static nameservers if dhcp is enabled

Taapat:
If you switch to dhcp and save the setting, but before that static nameservers have been set, delete these static nameservers to obtain nameservers only from dhcp.
Otherwise, the previous static nameservers are used, but you cannot set them and may not understand why they are used.


neo-pli-bsps:
What if you want to use DHCP for IP, and static DNS to override DHCP DNS? 

Does that have an impact on that use case?

PR #3679, opened 2023-03-31
[InfoBarInstantRecord] improvize code

Dima73:
- add more information to warning messages output
- change the logic of displaying messages / time duration of records -code cosmetics
-add Hotkey option "Start recording current event" -set always 24 h when use Hotkey option "Start instant recording"

### DIFF
--- a/lib/python/Screens/InfoBarGenerics.py
+++ b/lib/python/Screens/InfoBarGenerics.py
@@ -2495,7 +2495,7 @@ class InfoBarInstantRecord:
 				self.recording = InfoBarInstance.recording
 
 	def moveToTrash(self, entry):
-		print("instantRecord stop and delete recording: ", entry.name)
+		print("[instantRecord] stop and delete recording: %s", entry.name)
 		from Tools.Trashcan import createTrashFolder
 		trash = createTrashFolder(entry.Filename)
 		from Screens.MovieSelection import moveServiceFiles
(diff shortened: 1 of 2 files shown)


lib/python/Screens/InfoBarGenerics.py
```diff
@@ -2495,7 +2495,7 @@ def __init__(self):
 				self.recording = InfoBarInstance.recording
 
 	def moveToTrash(self, entry):
-		print("instantRecord stop and delete recording: ", entry.name)
+		print("[instantRecord] stop and delete recording: %s", entry.name)
```
Taapat, line 2498: Is it really intended here print symbol %s?

lib/python/Screens/InfoBarGenerics.py
```diff
@@ -2562,66 +2562,87 @@ def getProgramInfoAndEvent(self, info, name):
 			info["eventid"] = curEvent[4]
 			info["end"] = curEvent[1]
 
-	def startInstantRecording(self, limitEvent=False):
+	def startInstantRecording(self, limitEvent=""):
 		begin = int(time())
-		end = begin + 3600      # dummy
+		end = begin + 3600 # 1h
```
Taapat: If you are changing the comment format, why not change it according to PEP with two spaces from the statement?
https://peps.python.org/pep-0008/#inline-comments

lib/python/Screens/InfoBarGenerics.py
```diff
@@ -2562,66 +2562,87 @@ def getProgramInfoAndEvent(self, info, name):
...
+				message += duration_message
+			self.session.open(MessageBox, text=message, type=MessageBox.TYPE_INFO, timeout=timeout)
+		return added_timer
+
+	def startRecordingCurrentEvent(self):
+		self.startInstantRecording(limitEvent="event")
 
 	def isInstantRecordRunning(self):
-		print("self.recording:", self.recording)
+		#print("[InfoBarInstantRecord] self.recording:", self.recording)
```
Taapat: If you are commenting a line, why not do it according to PEP with a single space after the #?
https://peps.python.org/pep-0008/#inline-comments

Taapat:
I commented only on the first place where I think correction is needed to show what is wrong.
It would be good if you check all the proposed changes.
Print without keyword is also introduced in lines 2739 and 2748.
Two spaces from the statement are also missing in lines 2601, 2607 and 2612.

PR #3681, opened 2023-04-03
add Kexec support

Dima73:


littlesat:
Wasn't this already in?

Dima73:
this  8.3 branch

littlesat:
The OpenPLi team needs to decide it we should add this feature. I also saw a new OFGWrite was needed....
It is added value, but also a new feature, usually we do not add new features to a stable image. For a stable image only 'bug' fixes are wishful...

Dima73:
This is not a new feature, but simple compatibility.
I remind you that 8.3 is the last image with python 2.7 and many users will put it in the new multiboot as stable.
But they won't be able to move to another slot.
Why don't we help users?
I checked everything myself, there are no problems.
And a new version ofgwfite, what's the problem with updating it?

littlesat:
I understand but I cannot decide this alone. Basically it is a new feature 'to be compatible' with something new. Please open a discussion on the forum and when the team agrees I'll merge this.

PR #3692, opened 2023-04-11
[TimerList] add tuner number before tuner type name

Dima73:


littlesat:
Can you please describe why this is needed?

Dima73:
> Can you please describe why this is needed?

Very convenient for those who use a complex system of tuners and a lot of recordings.
Now you will know exactly which tuner is used for which recording.

RobvanderDoes:
Although that may change because of live-tv.

Dima73:
Timerlist dynamic list.
Changes every time any timer state changes.

littlesat:
For what is it needed....? for showing stuff with the recording info bar plugin or so?
I think it is a good idea to know which tuner is used for which recording. Can also be used to get OWIF better showing where tuners are used for etc....  Would also be nice to have the tuner number when something is streamed from a box....

In the patch there are also some cosmetic things observable for example " replaced by ' and visa versa. Can they be removed to keep the merge request clean for the change it does.

littlesat:
Just a small observation to think about

And maybe also those can be 'streamlined'

return tuner_name + ("%s" % "DVB-T") + alternative

->

return ("%s %s %s" % (tuner_name, "DVB-T", alternative)).strip()

(then also for tuner_name and alternative (A) you do not need the extra space)

Dima73:
> Just a small observation to think about
> 
> And maybe also those can be 'streamlined'
> 
> return tuner_name + ("%s" % "DVB-T") + alternative
> 
> ->
> 
> return ("%s %s %s" % (tuner_name, "DVB-T", alternative)).strip()
> 
> (then also for tuner_name and alternative (A) you do not need the extra space)

In your version, if there is tuner_name = "", then we will get an extra space.

littlesat:
The strip will remove it. But it is ok like you did it. I have to sleep on that part. But preferably the ' to " changes could be our.

PR #3697, opened 2023-04-12
[NimManager] remove "internally_connectable" from the first tuner of …

Dima73:
…the second slot duo2

it's a dummy value
1 slot

A  --> Vuplus DVB-S NIM(AVL6222)  --> I2C_Device: 2 -->

B  --> Vuplus DVB-S NIM(AVL6222)  --> I2C_Device: 2 --> /proc/stb/frontend/1/rf_switch:external

 

2 slot

C --> Vuplus DVB-S NIM(AVL6222)  --> I2C_Device: 4 --> /proc/stb/frontend/2/rf_switch:external   have internal loopthrough??

D --> Vuplus DVB-S NIM(AVL6222)  --> I2C_Device: 4 --> /proc/stb/frontend/3/rf_switch:external

### DIFF
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -949,6 +949,8 @@ class NimManager:
 							entry["internally_connectable"] = 1
 					elif id:
 						entry["internally_connectable"] = entry["frontend_device"] - 1
+						if HardwareInfo().get_device_model() == "vuduo2" and entry["i2c"] != entries[id - 1]["i2c"]:
+							entry["internally_connectable"] = None
 			else:
 				entry["frontend_device"] = None
 			if "multi_type" not in entry:


littlesat:
We have an if box x then do y on an incorrect place so when you want to do this then via systeminfo.py. If box x then do y is not allowed here

Dima73:
> We have an if box x then do y on an incorrect place so when you want to do this then via systeminfo.py. If box x then do y is not allowed here

I don't understand how to write the code.
Please show an example.

littlesat:
Something like this..
(I know it is a pain.... but then it is a 'feature'... and you centralize the if box x then y to one file.... )

In SystemInfo.py for example:

SystemInfo["IgnoreInternallyConnectableTuner"] = HardwareInfo().get_device_model() == "vuduo2"

Then in NimManager.py:

if SystemInfo["IgnoreInternallyConnectableTuner"] and entry["i2c"] != entries[id - 1]["i2c"]:

For the SystemInfo key of course we might find a better name.

Dima73:
`SystemInfo["IgnoreInternallyConnectableTuner"] = HardwareInfo().get_device_model() == "vuduo2"`
What is the meaning of this?
After all these SystemInfo["IgnoreInternallyConnectableTuner"],SystemInfo is impossible to read, dozens of similar lines that are difficult to understand.
This line(HardwareInfo().get_device_model() == "vuduo2") is read one when enigma is loaded, the receiver is discontinued.
The company does not exist.
On the contrary, I think it's right, everything is clear and why.

littlesat:
In openpli if box x then do y is only allowed in SystemInfo.py. That is something mandatory we agreed long time ago to avoid this becomes anywhere in the code so when a new box comes up you have to make exceptions all over the code. Actually that we make the exception for SystemInfo.py was already an exception/compromis. We should not deviate from this rule! So indeed the only work a round is call it a feature and do it via systeminfo.py. This sounds indeed complicated but at the end effect it still forces that are still somehow structured and do not have exceptions all over the place.

The biggest pain here is that other images do except it on multiple places and they never told tenders to adapt and fix their drivers.... :(

The keyname I made might be made shorter of course... and maybe better... it is also a kind of pain you need to check on i2c stuff there. Maybe there is another way to work-a-round this?

WanWizard:
There is still the plan on the table to migrate to OE-A's new form of "box branding". That migration is virtually impossible if there are checks on hardware dependencies all over the place.

PR #3712, opened 2023-05-17
[MovieInfo] not use full path of the directory

Dima73:


littlesat:
Good idea as this makes it more readable..

But

Why 
+from os.path import basename, normpath

I do not see it used in you merge request

Why
+raise ElementError("'%s' is not <ShortDescription|MetaDescription|RecordServiceName|FileSize|FullDescription|Name> for MovieInfo converter" % type)

Nothing to do with your merge request?

And just my 2 cents... make this a function:

isDirectory = (service.flags & eServiceReference.flagDirectory) == eServiceReference.flagDirectory
->
[code]
def isDirectory():
     return (service.flags & eServiceReference.flagDirectory) == eServiceReference.flagDirectory
[/code]
then it can maybe also be used elsewhere

Dima73:
`from os.path import basename, normpath`
so as not to import the entire os module
`Why
+raise ElementError("'%s' is not <ShortDescription|MetaDescription|RecordServiceName|FileSize|FullDescription|Name> for MovieInfo converter" % type)`
because FullDescription|Namе types were missing

jbleyel:
This is the original commit.

https://github.com/openatv/enigma2/commit/dc93cfcb6c4e0c551a4ca5b1f1adeb787ced1289

littlesat:
The isDirectory can also be made simpler.... remove the == and what is behind it and put bool(..) around the flag &

PR #3724, opened 2023-06-12
[FlashImage] add flashing recovery  image

Dima73:


littlesat:
I think we should not do this…. First get it alligned. I prefer having a kind of recovery image or make the slot 1 a normal slot and arrange the other slots are somehow backed up on HDd/usb in a stealthy way via e.g. ofgwrite

littlesat:
See my comments. As it is now it is better not to get this exception slot flashed.

PR #3725, opened 2023-06-16
[Satconfig] Reword "Connected"

AbuBaniaz:
Include wording that is in hint text


AbuBaniaz:
Some people deduced if thy were being asked if a signal wire was was connected to the tuner.  If you have a wire `connected` to the tuner, you say no to the `Connected` question. By expanding the field, it is clearer what the question is asking.

Connected through another tuner

PR #3733, opened 2023-06-23
[About] not visible str(fp_version).startswith("0")

Dima73:
gbquad4k login: root
root@gbquad4k:~# python
>>> ret = open("/proc/stb/fp/version", "r").read()
>>> ret
'0\n'


littlesat:
Maybe better use strip() what comes back in the line above and then ensure it is a string also over there and check for "0". Maybe better then do it with startswith().

Dima73:
My version works with both string and number.

littlesat:
It works but with strip() instead of startswith() it works always...

littlesat:
Please use strip to get rid of the extra \n the gigablue returns that is more universal solution

PR #3736, opened 2023-07-01
Fix ntpdate-sync crontab entry never written

scriptmelvin:


littlesat:
Can you please describe why this exactly is needed?

scriptmelvin:
Because `/var/spool/cron/crontabs/root` exists but is empty after a fresh flash without backup restore, so the next line that writes the crontab entry never gets executed.

littlesat:
Good point!!! And good finding... Thanks!!!
Next time maybe also describe it in the merge request 😀

scriptmelvin:
I described it in the forum, I thought that was enough. Will describe in next PR, thanks.

PR #3739, opened 2023-07-04
Update menu.xml

badsystem:
This is required to display the icon in main-setup, as in the main-menu. https://www.mediafire.com/view/4w2u8yxhjet4f5r/%25D0%2591%25D0%25B5%25D0%25B7_%25D0%25BD%25D0%25B0%25D0%25B7%25D0%25B2%25D0%25B8-2.png/file https://www.mediafire.com/file/hr3tpgvksia2hko/809.jpg/file

### DIFF
--- a/data/menu.xml
+++ b/data/menu.xml
@@ -32,6 +32,8 @@
 				<item text="Automatic scan" entryID="auto_scan" conditional="nimmanager.somethingConnected()"><screen module="ScanSetup" screen="ScanSimple"/></item>
 				<item text="Manual scan" entryID="manual_scan" conditional="nimmanager.somethingConnected()"><screen module="ScanSetup"/></item>
 				<item text="Fallback remote receiver setup" entryID="fallbacktuner_settings"><screen module="SetupFallbacktuner"/></item>
+				<item text="Signal finder" entryID="satfinder"><screen module="Satfinder"/></item>
+				<item text="FastScan" entryID="fastscan"><screen module="FastScanScreen"/></item>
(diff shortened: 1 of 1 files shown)


badsystem:
I'm sorry. Instead of deleting the message, I deleted the pull request

WanWizard:
@littlesat Is this ok?

Afaik these two are plugins. If this is hardcoded, then the menu entries remain after you removed the plugin, which might lead to crashes?

badsystem:
Crashes really happen when plugins are removed, they remain in the menu.

Dima73:
return [(_("FastScan"), FastScanMain, "fastscan", None)]
"fastscan"(this icon) isn't that the same?

WanWizard:
> Crashes really happen when plugins are removed, they remain in the menu.

That is my point. On a box without DVB-S, the fastscan plugin isn't installed, and both are not required for Enigma to function so can be removed by the user.

Dima73:
> > Crashes really happen when plugins are removed, they remain in the menu.
> 
> That is my point. On a box without DVB-S, the fastscan plugin isn't installed, and both are not required for Enigma to function so can be removed by the user.

I don't need this plugin at all.
I delete it right away.

littlesat:
It are plugins so it is not a good idea to put them in the menu.xml. You can remove them. So when a icon is missing it should be resolved a different way inside the plugin or we need to put it in standard code somehow. I have not any idea how it should be resolved in the plugin...

PR #3745, opened 2023-08-06
[Ci] add boot delay/extra_nigh bitrate

Dima73:
thanks @jbleyel


littlesat:
What does it resolve/what does it do/why is this needed?
I gamble it does work-a-round initiation issues and had also something else with high rate to it is a solution and an improvement in one commit. But as I gamble it will likely resolve init issues for ci I merged it.

PR #3746, opened 2023-08-06
[filepush] revert back to 2020-11-23

Dima73:


jbleyel:
Hi @Dima73 , what's the reason for this change?

Dima73:
> Hi @Dima73 , what's the reason for this change?

[Enigma 2 freezes "[eFilePushThreadRecorder] thread could not be stopped!" VU+ DUO 4K SE](https://forums.openpli.org/topic/95724-enigma-2-freezes-efilepushthreadrecorder-thread-could-not-be-stopped-vu-duo-4k-se/)

littlesat:
Please give more information? I do not understand why this needs to be reverted

WanWizard:
I disagree with a blanket revert of a three year old commit without knowning what the problem is, and with a fat chance of breaking other things.

littlesat:
That were also my thoughts... and I can confirm the revert breaks streamconvert for vu+.

AbuBaniaz:
We reverted these on ViX as they caused issues with recording. They were added to ViX on Jan 23, 2021 and reverted on Feb 7, 2021. IIRC, there was discussion on PLI forum about this at time. As no indication on fixing the underlying issue, the reliability of recordings and system stability led to reverting the changes.

We have not had any issues on Vix with omission of the broadcom encoding changes

@SimonCapewell had added these, if they are of use
https://github.com/OpenViX/enigma2/commit/3172dd4b978dcd692cb8291b6c3e95c718d2b661
https://github.com/OpenViX/enigma2/commit/d93ddcd81f25972e6b4933278d6ca5f52806286c

WanWizard:
I don't debate reverting it could or would fix it. I only think finding the bug and fixing it is a better approach, especially since the number of people using streaming by far outweighs the few reporting a recording issue that can't be reproduced by other,

In dutch we call this "throwing the baby out with the bath water"...

Dima73:
` breaks streamconvert for vu+`
This patch does not affect this in any way.

littlesat:
But nobody tels here how we can verify/confirm recording issues. What triggers it and I never experience it. So it is simply not clear why we should revert it. And in a statistical way for years we did not get any reports. So before a revert/fix we we need to know what triggers it. 'Just revert' is finding that a change 'triggered' an issue but does not mean it resolved the issue...

Dima73:
[eriksl](https://github.com/OpenPLi/enigma2/commits?author=eriksl) knows all the details and is aware of this issue.
For three years no solution has been found.
Over the years, users have complained many times about the freezing of the receiver and it has always been a recording.
So conclusions can be drawn.

littlesat:
I never heard about this issue and for now I have no idea how to trigger/verify it. So no cause is known. Reverting back is attack the symptom and not the problem.
Please get a better problem description. At least we need some instruction to be able to verify/trigger it ourselves.

Dima73:
> I never heard about this issue and for now I have no idea how to trigger/verify it. So no cause is known. Reverting back is attack the symptom and not the problem.
> Please get a better problem description. At least we need some instruction to be able to verify/trigger it ourselves.

I don't have such problems.
But I pointed out where they come from.

AbuBaniaz:
You must have missed the posts/discussion on forum then.

I can reproduce the problem on my duo4k. I will try and connect it and repeat the test again within few days. For me, it happens after you record several channels.

ims21:
> You must have missed the posts/discussion on forum then.
> 
> I can reproduce the problem on my duo4k. I will try and connect it and repeat the test again within few days. For me, it happens after you record several channels.

I am using my Duo4K for recording many channels in same time, in same time is streaming from Duo4k and no any problems...

PR #3747, opened 2023-08-08
[eDVBServiceRecord] fix some recording issues

Dima73:
- do not create cuts file if there is no recording
- do not create access points if there is no recording file

thanks @jbleyel


littlesat:
I hate to do the return when the file does not exits. I prefer to do the check if file exists and intend the rest what should be done. But it makes sense

PR #3750, opened 2023-08-12
[Navigation] recordService print ref.toString()

Dima73:


littlesat:
As far I can see it Conley prints some comments in the debug log so does not have a real functionality. And why did you put the if lower in the code?

WanWizard:
Wild guess: because "ref" was used before it was assigned? :grin:

Dima73:
before
[Navigation] recording service: <enigma.eServiceReference; proxy of <Swig Object of type 'eServiceReference *' at 0xb10946c8> >
[Navigation] recording service: <enigma.eServiceReference; proxy of <Swig Object of type 'eServiceReference *' at 0xb10946f8> >

PR #3753, opened 2023-08-17
[UsageConfig] Add qaf to original language

Dima73:


Huevos:
I think you have made a "gaf"... opps, I mean "gaffe". LOL.

WanWizard:
Indeed. 

Should it be "qaf" or "gaf"? In the first case the commit is wrong, in the second case the commit text is wrong... :grin:

PR #3761, opened 2023-08-27
About.py: Fix wrong driver date dm8000

Hains:
Due to an appended 'a' in driver date, driver date is displayed as '014-60-4a'. Increase sliding window to fix this.

### DIFF
--- a/lib/python/Components/About.py
+++ b/lib/python/Components/About.py
@@ -150,7 +150,7 @@ def getDriverInstalledDate():
 	try:
 		from glob import glob
 		try:
-			driver = [x.split("-")[-2:-1][0][-8:] for x in open(glob("/var/lib/opkg/info/*-dvb-modules-*.control")[0], "r") if x.startswith("Version:")][0]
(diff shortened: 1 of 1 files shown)


littlesat:
A bit tricky solution. What is the unexpected input string? Maybe we can find a better solution?

Hains:
it can do without the last slider [-8:]. Then date is still 2014-06-04a.
Problem is, I don't know what it does to other boxes?

The complete input string is 

`Version: 3.2-20140604a-r7`

BTW.

Hains:
Tested with Gigablue4KTrio and can also without [-8:]

So this:

`driver = [x.split("-")[-2:-1][0] for x in open(glob("/var/lib/opkg/info/*-dvb-modules-*.control")[0], "r") if x.startswith("Version:")][0]`

littlesat:
Maybe just take the date part and then limit it to 8 first characters...? So the a is always not taken

Hains:
New PR.

littlesat:
Why not just [:8] to take the first 8 characters from the date part? Instead of keeping the last 8 characters?

PR #3768, opened 2023-09-06
[TunerInfo] fix division

Dima73:


ims21:
int(log(val) // log(2)) is equal int(log(val) / log(2)) ... int(log(val) // log(2)) it's like wearing a belt and suspenders at the same time, imho

When val is not an integer, it must be return **int**(val * 100 / 65535). No matter if then there is //

Dima73:
> int(log(val) // log(2)) is equal int(log(val) / log(2)) ... int(log(val) // log(2)) it's like wearing a belt and suspenders at the same time, imho
> 
> When val is not an integer, it must be return **int**(val * 100 / 65535). No matter if then there is //

OK
Closed

littlesat:
With the log doing // makes no sense

littlesat:
When you use // I think the conversion to int is not mandatory.

PR #3773, opened 2023-09-20
Audio codec description replace one standard

Dima73:
-stream services return completely different audio codec, it is necessary to centralize this audio codec

thanks @original-birdman

### DIFF
--- a/lib/python/Components/Converter/ServiceInfo.py
+++ b/lib/python/Components/Converter/ServiceInfo.py
@@ -5,6 +5,41 @@ from Components.Element import cached
 
 WIDESCREEN = [3, 4, 7, 8, 0xB, 0xC, 0xF, 0x10]
 
+_acedits = (
+	("A_", ""),
+	("AC-3", "AC3"),
+	("(ATSC A/52)", ""),
+	("(ATSC A/52B)", ""),
+	(" Layer 2 (MP2)", ""),
+	(" Layer 3 (MP3)", "MP3"),
+	("-1", ""),
+	("-2", ""),
+	("2-", ""),
+	("-4 AAC", "AAC"),
+	("4-AAC", "HE-AAC"),
+	("audio", ""),
+	("/L3", ""),
+	("/mpeg", "AAC"),
+	("/x-", ""),
+	("raw", "Dolby TrueHD"),
+	("E-AC3", "AC3+"),
+	("EAC3", "AC3+"),
+	("IPCM", "AC3"),
+	("LPCM", "AC3+"),
+	("AAC_PLUS", "AAC+"),
+	("AAC_LATM", "AAC"),
+	("WMA/PRO", "WMA Pro"),
+	("MPEG", "MPEG1 Layer II"),
+	("MPEG1 Layer II AAC", "AAC"),
+	("MPEG1 Layer IIAAC", "AAC"),
+	("MPEG1 Layer IIMP3", "MP3"),
+)
+def StdAudioDesc(description):
+	if not description:
+		return ""
+	for orig, repl in _acedits:
+		description = description.replace(orig, repl)
+	return description
 
 class ServiceInfo(Converter):
 	HAS_TELETEXT = 0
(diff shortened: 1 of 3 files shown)


Taapat:
1. Why create _acedits as a global variable if it is only used in the StdAudioDesc function? I think it would be correct to place it in the in function where we use it, and not to create unnecessary global variables.
2. Maybe I'm missing something, but description should be the name of the codec.
Why is a loop through a tuple needed to replace the description at all? Why not create a dictionary and set the name directly without a loop?
3. After description.split()[0] you check if it is not "WMA Pro".
How can it consist of two words if split will divide it?

original-birdman:
> 1. Why create _acedits as a global variable if it is only used in the StdAudioDesc function? 

So that it is created once, not every time the function is called.

>     2. Maybe I'm missing something, but description should be the name of the codec.

What you are missing is that this was taken from an OpenVix PR, where there were 4 long strings of **identical** replace()s all in a single-statement. All this did was centralize/rationalize code which _already_ existed there.

>        Why is a loop through a tuple needed to replace the description at all? Why not create a dictionary and set the name directly without a loop?

Multiple replace()s can occur in one description.

Taapat:
> So that it is created once, not every time the function is called.

And it stays in memory forever. Why don't we define all variables globally so that they are only created once?


> What you are missing is that this was taken from an OpenVix

But this is not a PR in OpenVix. 4 long strings of identical replaces do not exist here.
In OpenVix it has come from openatv. In my opinion, on openatv this replace is in the correct place - in servicemp3.
Why get the description from one place in servicemp3 and change it ater in python code in several places?

original-birdman:
> And it stays in memory forever. Why don't we define all variables globally so that they are only created once?

It would need to be in memory to create it locally. It doesn't get magicked out of thin air.
Yes, it is now a global name (hence the starting "_"), but unless you import it you won't have a clash.

original-birdman:
> But this is not a PR in OpenVix. 

It is as far as my involvement goes. and where it originated. And hence why I'm getting message about it.

Ev0-BH:
> But this is not a PR in OpenVix. 4 long strings of identical replaces do not exist here. In OpenVix it has come from openatv. In my opinion, on openatv this replace is in the correct place - in servicemp3. Why get the description from one place in servicemp3 and change it ater in python code in several places?

The original code was in the 4 python files, which I updated in OpenBh and was pushed to OpenVix, then OpenAtv merged it and then ported it to servicemp3, however that does not work properly as every media player gives a different name for a codec, so you need to patch serviceapp too, but again it still did not work properly (in OpenBh atleast), so it was reverted and then updated in the 4 python files again.

original-birdman has just cleaned up 4 files and put all the codec replaces into 1 file to make it nice and simple and then created a PR for OpenPLi, implement it however you want, either in python or in c++.

Dima73:
My test...
This replace work fine in serviceapp/servicemp3 for OpenPLi.


I’m also tired of these incomprehensible names of codecs from streaming services, now everything is as it should be.

Dima73:
`So that it is created once, not every time the function is called.`

Converter ServiceInfo сalled after each zap.
And there it is constantly checked "IsMultichannel"(standard OpenPLi skins).

Taapat:
As I understand this only affects streaming.
So ,I don't see a problem to add it also in serviceapp if needed.
Ok, in addition to that, it will also be needed in servicehisilicon and then we can discuss whether it should be placed in python or not.

I assume that the name of the codec does not depend on the media player, as you write, but on the streaming service.
Why you want to convert something when the streaming service has specified a codec based on their own considerations or standards?
Are all the necessary replace variants here? Maybe you need more, but maybe you need less?
As far as I can observe in the commit history, all changes have been made without specific examples what needs to be replaced and why.

Taapat:
@Ev0-BH and how can you comment on this: https://github.com/openatv/enigma2/commit/1891f68cf79c19f55c3f79471f3c5ecd50cc02b6
I understand that all the replaces offered here are not needed?

Ev0-BH:
> @Ev0-BH and how can you comment on this: [openatv/enigma2@1891f68](https://github.com/openatv/enigma2/commit/1891f68cf79c19f55c3f79471f3c5ecd50cc02b6) I understand that all the replaces offered here are not needed?

Not sure what you are calling me out on ?

As I said above, after OpenAtv merged the python changes, they then converted it to C++ in servicemp3 and serviceapp, but it did not work well in OpenBh, I made some changes that improved things (which is what the commit above is that OpenAtv also merged), but ultimately it was reverted in OpenBh and OpenVix and we went back to the replaces in python.

PR #3785, opened 2023-10-04
Python changes for compatibility with PLi-DarkOS skin

DimitarCC:
This is python changes required for PLi-DarkOS skin to work fine and the design to be modern.
As well it contains changes for supporting remembering of audio and subtitle tracks for ITPV services. (this works best if IPTV service is DVB compliant)
There are also some additional GUI elements for some plugins

### DIFF
--- a/lib/python/API.py
+++ b/lib/python/API.py
@@ -1,0 +1,9 @@
+session = None
+
+class API(object):
+	def __init__(self):
+		self.__session = None
+		
+	def setSession(self, current_session):
+		global session
+		session = current_session
--- a/lib/python/Components/ChoiceList.py
+++ b/lib/python/Components/ChoiceList.py
@@ -1,6 +1,7 @@
 from Components.MenuList import MenuList
+from Components.MultiContent import MultiContentEntryText
 from Tools.Directories import SCOPE_CURRENT_SKIN, resolveFilename
-from enigma import RT_HALIGN_LEFT, eListboxPythonMultiContent, gFont
+from enigma import RT_HALIGN_LEFT, RT_VALIGN_CENTER, eListboxPythonMultiContent, gFont
 from Tools.LoadPixmap import LoadPixmap
 from skin import applySkinFactor, fonts, parameters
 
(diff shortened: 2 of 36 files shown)


littlesat:
Better to fix the skin for these exceptions. Add more then 1500 lines for just satisfying a skin is not what we should do

DimitarCC:
Skin can not be fixed when skinning engine doesn't support the required visual elements.
It is up to you to decide should you merge that or not. My opinion is that OpenPLi stays a bit in the past in the matter of design and not progressing. Thats why i made the skin in the way it is. 
If you not like merging this....OK...Then users will install that manually ;)
Just there was request that to be provided for merging so i made that PR.

P.S. actually the changes are not only for satisfying a single skin, but satisfying all next skins that wants to look modern.
Also provide extensions for IPTV that no one made so far (just because devs here not use it)

jbleyel:
You are mixing complete different things in one big commit. And to be honest. Your code is not modern and full of bloated functions which are slow down the box.

DimitarCC:
I agree the code is not optimized. I try to do something that requires c++ changes to be only in python. (mainly because no one of you wants to do any kind of c++ changes in here)
Actually no one of you wants to understand what i am trying to achieve.
Anyway and which function exactly slow down the box? In my opinion when somebody claim a function slow down the box have to say which is it (which means he have to test the code and see i it really slow down the box) ;)

I think that discussion is pointless anyway. As i said i dont want to force anyone to merge that or to consider it. I made that PR by request. If you dont want to consider it i am fine with it. On my 4 Vu+ boxes there are no slowdowns and everything works fine.

So simply accept or reject this PR and finish with it. Thanks!

littlesat:
The skin can be fixed when the required needed stuff is implemented correctly and weighted where it belongs and a lot of time to do thinks where they belong (eg in c++). In addition all adds are not really add a real value.... it slows down stuff just for a look and it add stuff that is already done in plugins (eg permanent clock).

DimitarCC:
@littlesat As i wrote in the forum let me break the functions one by one and will see

PR #3786, opened 2023-10-04
Remember the last selected audio and/or subtitles for IPTV streams

DimitarCC:
This is for remembering the last selected audio or subtitle track for IPTV services.
I know there is a plugin for AutoAudioSelection but it relies on rules only and can not remember the actual user selection.

Where is needed that? Well for IPTV services that are DVB compliant and have multiple audio or subtitle tracks. Now it can not behave like the real DVB service regarding audio and subtitle user selections.

This change fix/add that functionality.
Forum thread [here](https://forums.openpli.org/topic/96254-storing-last-selected-audio-and-subtitle-track-for-iptv/)

### DIFF
--- a/lib/python/Screens/AudioSelection.py
+++ b/lib/python/Screens/AudioSelection.py
@@ -6,12 +6,17 @@ from Screens.ChoiceBox import ChoiceBox
 from Components.ServiceEventTracker import ServiceEventTracker
 from Components.ActionMap import NumberActionMap
 from Components.ConfigList import ConfigListScreen
-from Components.config import config, ConfigSubsection, ConfigNothing, ConfigSelection, ConfigOnOff, ConfigYesNo
+from Components.config import config, ConfigSubsection, getConfigListEntry, ConfigNothing, ConfigSelection, ConfigOnOff, ConfigYesNo
 from Components.Label import Label
 from Components.Sources.List import List
 from Components.Sources.Boolean import Boolean
 from Components.SystemInfo import SystemInfo
 from Components.VolumeControl import VolumeControl
+from os import path as os_path, open as os_open, close as os_close, O_RDWR as os_O_RDWR, O_CREAT  as os_O_CREAT 
+from Tools.General import isIPTV
+from Tools.Directories import resolveFilename, SCOPE_CURRENT_SKIN
+from Tools.LoadPixmap import LoadPixmap
+from pickle import load as pickle_load, dump as pickle_dump, dumps as pickle_dumps
 from Components.UsageConfig import originalAudioTracks, visuallyImpairedCommentary
 from Components.Converter.ServiceInfo import StdAudioDesc
 from Tools.ISO639 import LanguageCodes
@@ -20,6 +25,19 @@ from enigma import iPlayableService, eTimer, eSize, eDVBDB, eServiceReference, e
 
 FOCUS_CONFIG, FOCUS_STREAMS = range(2)
 [PAGE_AUDIO, PAGE_SUBTITLES] = ["audio", "subtitles"]
+
+CONFIG_FILE_AV = '/etc/enigma2/config_av'
+
+selectionpng = LoadPixmap(cached=True, path=resolveFilename(SCOPE_CURRENT_SKIN, "selections/selectioncross.png"))
+
+def getAVDict():
+	if os_path.exists(CONFIG_FILE_AV):
+		pkl_file = open(CONFIG_FILE_AV, 'rb')
+		if pkl_file:
+			avdict = pickle_load(pkl_file)
+			pkl_file.close()
+			return avdict
+	return {}
 
 
 class AudioSelection(ConfigListScreen, Screen):
@@ -78,6 +96,27 @@ class AudioSelection(ConfigListScreen, Screen):
 		self["config"].instance.setSelectionEnable(False)
 		self.focus = FOCUS_STREAMS
 		self.settings.menupage.addNotifier(self.fillList)
+		
+	def saveAVDict(self, dict):
+		pkl_file = open(CONFIG_FILE_AV, 'wb')
+		if pkl_file:
+			pickle_dump(dict, pkl_file)
+			pkl_file.close()
+			
+	def setAVInfo(self, service):
+		playinga_idx = service and service.audioTracks().getCurrentTrack() or -1
+		ref = self.session.nav.getCurrentlyPlayingServiceReference()
+		x = ref.toString().split(":")
+		ref_str = ":".join(x[:10])
+
+		playing_idx = self.infobar.selected_subtitle
+
+		if playing_idx:
+			self.infobar.av_config[ref_str] = pickle_dumps(playinga_idx, 0).decode() + "|" + pickle_dumps(playing_idx, 0).decode()
+		else:
+			self.infobar.av_config[ref_str] = pickle_dumps(playinga_idx, 0).decode()
+
+		self.saveAVDict(self.infobar.av_config)
 
 	def fillList(self, arg=None):
 		streams = []
@@ -274,8 +313,12 @@ class AudioSelection(ConfigListScreen, Screen):
 	def changeAudio(self, audio):
 		track = int(audio)
 		if isinstance(track, int):
-			if self.session.nav.getCurrentService().audioTracks().getNumberOfTracks() > track:
+			service = self.session.nav.getCurrentService()
+			ref = self.session.nav.getCurrentlyPlayingServiceReference()
+			if service.audioTracks().getNumberOfTracks() > track:
 				self.audioTracks.selectTrack(track)
+				if isIPTV(ref):
+					self.setAVInfo(service)
 
 	def keyLeft(self):
 		if self.focus == FOCUS_CONFIG:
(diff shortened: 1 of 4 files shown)


Taapat:
Why are so many unused imports added here?
Each file contains at least one or more.

Why an unused variable selectionpng added to AudioSelection that points to a non-existent folder in enigma?

Why almost all the code in the avChange method placed in try?
If you are convinced that all this can cause GSOD in any place, maybe you can think about how to make it better?

lib/python/Tools/General.py
```diff
@@ -0,0 +1,5 @@
+from enigma import iServiceInformation, eServiceReference, eServiceCenter
+	
+def isIPTV(service):
+	path = service and service.getPath()
+	return path and not path.startswith("/") and service.type in [0x1, 0x1001, 0x138A, 0x1389]
```
jbleyel, line 4: I think, this iptv detection is wrong and should be moved to c++ by the way.
getPath() have not always the correct value depending on where you get the eServiceReference.

lib/python/Screens/InfoBarGenerics.py
```diff
@@ -467,6 +480,41 @@ def ToggleHideVBI(self, service=None):
 				whitelist.vbi.append(service)
 			open('/etc/enigma2/whitelist_vbi', 'w').write('\n'.join(whitelist.vbi))
 			self.showHideVBI()
+	
+	def queueChange(self):
+		self._waitForEventInfoTimer.stop()
+		self._waitForEventInfoTimer.start(50, True)
```
jbleyel, line 486: Why 50?

lib/python/Screens/AudioSelection.py
```diff
@@ -78,8 +93,33 @@ def __layoutFinished(self):
...
+		ref_str = ":".join(x[:10])
+
+		playing_idx = self.infobar.selected_subtitle
+
+		if playing_idx:
+			self.infobar.av_config[ref_str] = pickle_dumps(playinga_idx, 0).decode() + "|" + pickle_dumps(playing_idx, 0).decode()
+		else:
+			self.infobar.av_config[ref_str] = pickle_dumps(playinga_idx, 0).decode()
+
+		self.saveAVDict(self.infobar.av_config)
```
jbleyel, line 116: This is very bad if you call this every x seconds.

jbleyel: And how big will this file be if you start 1000 of different iptv streams?

lib/python/Tools/General.py
DimitarCC: Its true. But in the place i use it is correct. But yes if it is in the enigma itself will be better. If you willing to add it why not add the required c++ code  ;)

lib/python/Screens/InfoBarGenerics.py
DimitarCC: Why not? There should be a delay otherwise it will not enable the subtitles correctly.

lib/python/Screens/AudioSelection.py
DimitarCC: Why you think that is called every x seconds? And why you think it will ever be called often? You will open and select language or subtitles every second or what?!

setAVInfo is called only on change of the audio track or on pressing OK button. I dont see a reason that to be called often on purpose.


> And how big will this file be if you start 1000 of different iptv streams?

Not more than the enigma database that stores track information for normal DVB services. But if you want to test it out make 1000 services and post here how big it is ;)

I use it with 50 services that have subtitles or audio tracks and the size is  3Kb

DimitarCC:
> Why are so many unused imports added here? Each file contains at least one or more.
> 
> Why an unused variable selectionpng added to AudioSelection that points to a non-existent folder in enigma?
> 

> Why almost all the code in the avChange method placed in try? If you are convinced that all this can cause GSOD in any place, maybe you can think about how to make it better?

I think you become too picky and search any reasons that to not be merged...
But i will answer...

Yes there are unused imports since that left from the other functionality i stripped so to separate the code as @littlesat wanted.
In general that imports will not have such a great impact dont you think?

Regarding the try in avChange it is there because the pickle load can fail if someone corrupts the storage file and some of the values inside are not what is expected.
But if you have idea how to overcome this why no share it? It is nice that you point it out but i also can point a lot of issues in OpenPLi code just for the sport ;)

littlesat:
Not a good idea to do it this way. This is a work-a-round. Normally this is stored in lamedb. Better to do a real redesign and store it somehow in lamedb. Ensure or made something that it can be done via service ref.

DimitarCC:
@littlesat That means it will never be done because for last 10 year no one wants to mess with lamedb :(
And ofcource it's a workaround because unwilling from nobody to do it in c++.

Sometimes workarounds have to be accepted when there is no light in the tunnel to be done in the right way. ;)

At least with that workaround it will work vs not work at all without it.

lib/python/Screens/AudioSelection.py
jbleyel: Sorry.. I have missread this part.

lib/python/Screens/InfoBarGenerics.py
```diff
@@ -271,12 +274,19 @@ def __init__(self):
 
 		self.__event_tracker = ServiceEventTracker(screen=self, eventmap={
 				iPlayableService.evStart: self.serviceStarted,
+				iPlayableService.evEnd: self.serviceEnded,
+				iPlayableService.evUpdatedInfo: self.queueChange,
```
jbleyel, line 278: This can be a problem. How often do you get the evUpdatedInfo event?

jbleyel: So why not using evStart?

DimitarCC: Well evStart can not be used because on that event the info for available audio and subtitle tracks is not available yet.
That happens on evUpdatedInfo for audio tracks and evSubtitleListChanged for subtitles. But since we have to call this twice in this case i use only evUpdatedInfo. Indeed it can be called multiple times but i dont think is more than 2 times.

Huevos:
> Not a good idea to do it this way. This is a work-a-round. Normally this is stored in lamedb. Better to do a real redesign and store it somehow in lamedb. Ensure or made something that it can be done via service ref.

What do you mean. URLs are not stored in lamedb. Lamedb is for tuning information. Streaming is nothing to do with tuning. Please don't think of messing that up. 

We have already had the abandoned lamedb5 project and everyone had to write new code for that. Now we have this strange suggestion.

littlesat:
And now again I should make it really right like all the stuff in the past to prove it.... sorry I pass! My notification is gone due to responses like those from you..

littlesat:
I know urls aren't stored there. And lamedb5 is also not really accepted yet. But I miss design and agreement here. Again just a lot of hacking code is created so the whole thing got huger hackier and bigger and worser maintable and introducing bugs and more unexpected behavior.... by lake of first discus need... brainstorm, design and create.... the enigma2 code is dead due to this....

DimitarCC:
In my opinion the enigma2 code becomes dead because there are no developers left to maintain it. If every developer faces so much denial, he will not want to work on this ofcource. Or simply will move to another team. Maybe that is the reason OpenPLi to stuck and not evolve anymore... :(

littlesat:
It are about 90 extra lines for skin features and the code is not really lean and clean as far I can see in a quick view. This is huge for just satisfy some features used in skins without any thoughts about the real need and/or can something be done also with the code already available and why/how you need something. Actually the whole stuff around skins is crazy already for years. In the past skins did add their own python code renderers when needed. Also a crazy method. Just to get something that looks a bit nicer for someone... making often things slower.... more difficulty maintainable. Without any design thoughts and clear discussions how something is done as it is just done and hacked by some single person.... and then suddenly it is threwes over the wall, other images are spoiled by it and we should follow.... in the past I changed a lot to prove things could made better but I do not have the time and patience anymore.... so this is not my call

littlesat (commented):
Someone here exactly described the problem for something in events what is missing so we made code that work a round it instead of resolving it by finding a trigger or something else what is missing or we do not have. Having a delay or so is not done that is work a round something... exactly a mechanisme we should strive not to make prefer.... so first think about how we can create a kind of trigger/event when subs or audio channels becomes known  with streams

DimitarCC:
But there is not such a trigger created yet from years. And in this way it will not be created next years as well.
We are going in circles.....
You do not have a developer willing to do any kind of c++ extensions, then you dont want to accepts workarounds (even if they work) and demand the c++ code to be fixed and again no developers that willing to do so... Where you are going? In that way i doubt it ever be fixed.

That is discouraging for every developer as i stated.

littlesat:
So better fix it with work a rounds and tricks instead of discuss it and try to get something nice lean clear and clean?Verstuurd vanaf mijn iPhoneOp 4 okt. 2023 om 21:56 heeft Dimitar Tsenev ***@***.***> het volgende geschreven:﻿
But there is not such a trigger created yet from years. And in this way it will not be created next years as well.
We are going in circles.....
You do not have a developer willing to do any kind of c++ extensions, then you dont want to accepts workarounds (even if they work) and demand the c++ code to be fixed and again no developers that willing to do so... Where you are going? In that way i doubt it ever be fixed.
That is discouraging for every developer as i stated.

DimitarCC:
Not. But have you got other options? Since you have no developers left who will fix it even if you discuss it for months or years.
At the end it will stay just one discussion, and nothing will be implemented because there will be nobody left to do so.

Isnt it better to have something working (ofcource if it not breaks anything) instead to have nothing working and just a long long discussion.

littlesat:
I think we should find a better method

PR #3787, opened 2023-10-04
- added several convertors and renderers

DimitarCC:


lib/python/Components/Converter/ServiceNameEx.py
jbleyel, line 1: why not extending ServiceName Converter?

lib/python/Components/Renderer/BoxLogo.py
jbleyel, line 1: What's this?

lib/python/Components/Converter/ServiceNameEx.py
DimitarCC: Because the logic here is different...
I want to be possible to produce configurable single label with separators.
It can be integrated in ServiceName but the chance to be accepted as a separate converter is bigger in my opinion because it doesnt mess with system one. ;)

lib/python/Components/Renderer/BoxLogo.py
DimitarCC: This is renderer for display brand and model logos based on the hardware info model name.
I use this in my skin for the logo in the menu to display image for vuuno4kse or vuduo4kse and so on. So the skin to be able to display correct logo not matter at what box it works.

DimitarCC:
This is not relevant anymore since some changes are offered so the part of the code to be included in standard files.

PR #3788, opened 2023-10-04
UI extensions

DimitarCC:
- added menu icons
- added some UI refinenments and adjustments
- fix overlapping frame when used drawframe in servicelist

### DIFF
--- a/lib/python/Screens/AudioSelection.py
+++ b/lib/python/Screens/AudioSelection.py
@@ -15,11 +15,16 @@ from Components.VolumeControl import VolumeControl
 from Components.UsageConfig import originalAudioTracks, visuallyImpairedCommentary
 from Components.Converter.ServiceInfo import StdAudioDesc
 from Tools.ISO639 import LanguageCodes
+from Tools.Directories import resolveFilename, SCOPE_CURRENT_SKIN
+from Tools.LoadPixmap import LoadPixmap
 
 from enigma import iPlayableService, eTimer, eSize, eDVBDB, eServiceReference, eServiceCenter, iServiceInformation
 
 FOCUS_CONFIG, FOCUS_STREAMS = range(2)
 [PAGE_AUDIO, PAGE_SUBTITLES] = ["audio", "subtitles"]
+
+
+selectionpng = LoadPixmap(cached=True, path=resolveFilename(SCOPE_CURRENT_SKIN, "selections/selectioncross.png"))
 
 
 class AudioSelection(ConfigListScreen, Screen):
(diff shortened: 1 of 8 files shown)


lib/service/listboxservice.cpp
jbleyel, line 1: What's the purpose of this change?

DimitarCC: The comment say it all. Move that to be first so to be rendered under all other content. In this way for selections can be used images and that will not overlap the actual texts or icons in selected item.

littlesat:
I'm afraid just for a skin we should not merge this at all... it are huge changes that might break some other stuff.... maybe better to adapt the skin and discuss why/what/how you need something. It makes the code such bigger....

DimitarCC:
My apologies, but that sounds hilarious. What is the biggest part of it? Only seven files changed. What is huge and which part exactly can cause issues or break something? Most of the code already works in other distros and i didn't hear issues. 
People that use my skin didn't report issues as well.
And i don't understand why it is always noted "just for one skin"?! Don't you understand?! It's not for one skin, it's to make it possible diffrent new skins to be created!

lib/python/Screens/AudioSelection.py
```diff
@@ -15,13 +15,18 @@
...
+from Tools.Directories import resolveFilename, SCOPE_CURRENT_SKIN
+from Tools.LoadPixmap import LoadPixmap
 
 from enigma import iPlayableService, eTimer, eSize, eDVBDB, eServiceReference, eServiceCenter, iServiceInformation
 
 FOCUS_CONFIG, FOCUS_STREAMS = range(2)
 [PAGE_AUDIO, PAGE_SUBTITLES] = ["audio", "subtitles"]
 
 
+selectionpng = LoadPixmap(cached=True, path=resolveFilename(SCOPE_CURRENT_SKIN, "selections/selectioncross.png"))
```
Taapat: Folder "selections" does not exist in OpenPLi enigma2, selectioncross.png located in the folder "icons".

DimitarCC: well it was intended to be different image so to not overwrite the one in icons. It may want to have different image for that, because selectioncross in icons is used for other places as well. Maybe the naming is not right anyway.
And if you not define in the list template that you want to use image it will be no problem even if it is not there.

Taapat: You are using a folder that is not used in enigma and an image that does not exist in enigma.
It looks like you want to make an exception for your skin.
And by the way, you are using SCOPE_CURRENT_SKIN which is marked as deprecated scope.

DimitarCC: Not exception only for my skin. But yes if you want to use it other skin you have to add that folder in the skin directory.
Yes you are right for SCOPE_CURRENT_SKIN it have to be SCOPE_GUISKIN i think.

PR #3789, opened 2023-10-06
[skin/menu] patches

Huevos:
Fix some crash in previous commit.

Fix skin.

Add menuitems to skin.py.

Use latest version of menuitems in Menu.py for compatibility with other images that already use this system.

Menu.py, add missing handling for menu graphic. This has been implemented in skin.py for 5 years but is missing from menu.

Add a doc/MENU to explain how everything works when adding these feature to a skin.

### DIFF
--- a/doc/MENU
+++ b/doc/MENU
@@ -1,0 +1,93 @@
+Adding a menu image that identifies the current menu screen based on menuID
+===========================================================================
+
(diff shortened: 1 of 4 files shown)


Taapat:
My personal opinion is if the previous UI extensions are merged, then this one also needs to be merged, because this significantly improves the previous changes.

PR #3793, opened 2023-10-08
UI patches

DimitarCC:
- fixed some UI redundancies
- remove blanck labels
- extended several converters with single formatted string possibility
- endend skinning engine to be possible to add string value parameters

### DIFF
--- a/lib/python/Components/Converter/Converter.py
+++ b/lib/python/Components/Converter/Converter.py
@@ -5,9 +5,17 @@ class Converter(Element):
 	def __init__(self, arguments):
 		Element.__init__(self)
 		self.converter_arguments = arguments
+		self.separator = ""
 
 	def __repr__(self):
 		return str(type(self)) + "(" + self.converter_arguments + ")"
 
 	def handleCommand(self, cmd):
 		self.source.handleCommand(cmd)
+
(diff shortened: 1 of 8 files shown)


littlesat:
First I think we should not have stringparameters... please find a different solution...
Your skin needs such extra features. It can be intended but it should not add too much extra's

littlesat:
I'm not sure this is really needed anymore

DimitarCC:
Stringparameters has been reverted. So its not in the commit.
All the rest is the extensions for several convertors to have possibility to accept  this type of definition
`<convert type="ServiceName">•,NUMBER,NAME,ORBPOS,PROVIDER</convert>`

So i think should not be closed.

For example without extensions to ServiceName converter above can not be achieved.

Taapat:
Maybe you can create a separate pull request for ServiceName.
I think the option you offer is very good.
The only thing that is not clear to me is why types is intended in uppercase and not in camelcase as it is when specifying only one type.
I think these types should not be different.

DimitarCC:
I have already made new PR https://github.com/OpenPLi/enigma2/pull/3799
Well yes it can be in camel case...its just a matter of wording...
There is one parameter that is not in ServiceName but in Service info and that is Orbital position

PR #3796, opened 2023-10-11
[elistbox] Add method to calculate max list-item width for use in Mes…

Huevos:
…sageBox

Removes terrible "autoresize" hack in MessageBox.

Author: @DimitarCC

### DIFF
--- a/lib/gui/elistbox.h
+++ b/lib/gui/elistbox.h
@@ -53,6 +53,7 @@ protected:
 	virtual int getItemHeight()=0;
 	virtual int getItemWidth() { return -1; }
 	virtual int getOrientation() { return 1; }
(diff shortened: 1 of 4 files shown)


littlesat:
Please do not close... I now postpone it in regards to the release of 9.0.

Taapat:
This also could be used in ChoiceBox: https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Screens/ChoiceBox.py#L102-L106

Huevos:
ChoiceList is eListboxPythonMultiContent.

MenuBox is eListboxPythonStringContent.

We would need to check.

DimitarCC:
it should work in my opinion. But have to be tested on a skin that uses it.

PR #3800, opened 2023-10-18
- add getMaxTextWidth to eListboxPythonMultiContent

Huevos:
This is the same as the recent commit to fix autoresize in MessageBox but for eListboxPythonMultiContent.

As suggested by: @Taapat 

This can be used for ChoiceBox so the current workaround can be removed.

### DIFF
--- a/lib/gui/elistboxcontent.cpp
+++ b/lib/gui/elistboxcontent.cpp
@@ -929,7 +929,7 @@ int eListboxPythonMultiContent::getMaxItemTextWidth()
 
 					if (pfnt) {
 						int fnt_i = PyLong_AsLong(pfnt);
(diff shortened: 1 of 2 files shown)


Taapat:
As I wrote here: https://github.com/OpenPLi/enigma2/pull/3796
ChoiceBox does not require such a complex implementation. Just a few lines.
But it won't be correct in all cases when eListboxPythonMultiContent is used, so it might be worth implementing this.

DimitarCC:
There is a small bug in calculation that is fixed already. The PR will be updated soon from @Huevos.

littlesat:
Then I wait.

Huevos:
1) Calculation fixed.
2) Added sanity test requested by Taapat.

PR #3804, opened 2023-10-18
Change logic for formatted string in convertors

DimitarCC:
This is because of remarks of @Taapat ....

Now you can define in the skin with whatever  case you want and that will work.
Also spaces can be configured in the skin as well now.


littlesat:
Why not force the complete thing to upper instead do it label by label

DimitarCC:
I cant force it to upper because of the separator string and normal types that was used before. If i force type to be uppercase the normal
`<convert type="ServiceName">Reference</convert>`
not going to work anymore.

DimitarCC:
P.S. i was thinking for even more flexible way to define that string format....
I will prepare another pull for discussion....

PR #3815, opened 2023-10-26
ntp time -change logic

Dima73:
-when config.ntp.timesync.value is 'ntp' always use only ntp time
-config.misc.useTransponderTime. need call when initial( initial_call=False) because
timesyncChanged(when config.ntp.timesync) always set eDVBLocalTimeHandler.getInstance().setUseDVBTime(value)

### DIFF
--- a/lib/python/Components/UsageConfig.py
+++ b/lib/python/Components/UsageConfig.py
@@ -772,7 +772,7 @@ def InitUsageConfig():
 	config.ntp = ConfigSubsection()
 
 	def timesyncChanged(configElement):
-		if configElement.value == "dvb" or not GetIPsFromNetworkInterfaces():
+		if configElement.value == "dvb" or (configElement.value == "auto" and not GetIPsFromNetworkInterfaces()):
 			eDVBLocalTimeHandler.getInstance().setUseDVBTime(True)
 			eEPGCache.getInstance().timeUpdated()
 			if os.path.isfile('/var/spool/cron/crontabs/root'):
--- a/lib/python/StartEnigma.py
+++ b/lib/python/StartEnigma.py
@@ -92,7 +92,7 @@ def useTransponderTimeChanged(configElement):
 	enigma.eDVBLocalTimeHandler.getInstance().setUseDVBTime(configElement.value)
 
 
-config.misc.useTransponderTime.addNotifier(useTransponderTimeChanged, initial_call=False)
+config.misc.useTransponderTime.addNotifier(useTransponderTimeChanged)
 
 profile("Twisted")
 try:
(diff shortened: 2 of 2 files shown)


littlesat:
Sorry I read it triple but I still do not understand what you actually tried to resolve here. From the forum the issue here is when the network came up slow after enigma2 was started so there was no internet causing it falls back to transponder time. How does this resolve this mechanism?
As far I can see in your situation it starts with no time at all or somewhere in 1970s when the network is not running while enigma2 is started. I'm afraid the only real solution here is wait to start enigma2 until the network is up.... now you eg can still get recordings not started etc... and for this add an third option 'force ntp' or so instead of fallback to transponder time when the network is not up yet.

Dima73:
`and for this add an third option 'force ntp' or so instead of fallback to transponder time when the network is not up yet.`
config.ntp.timesync = ConfigSelection(default="auto", choices=[("auto", _("auto")), ("dvb", _("Transponder Time")), ("ntp", _("Internet (ntp)"))])
What's the difference between value 'auto'/'ntp'?
I thought that 'ntp' was 'force ntp'.

littlesat:
Auto does when there is no network at the moment enigma2 is started that it automatically falls back to transponder time. Ntp should give you no time at all (or 1-1-1970) when there is no network.
But you get a point. No need to add an option...  The only thing that is needed when you select ntp that you can only start enigma2 when there is a network and the date is above the 1970s. The biggest issue here is why is you network such slow.... :(

littlesat:
I also do not understand why this is needed...
config.misc.useTransponderTime.addNotifier(useTransponderTimeChanged, initial_call=False)
As initial should be true here as it configs enigma2 in the proper way.....
I'm afraid with this it always goes back to the default which is transponder time. Then it never can get to ntp, unless you change it during using the same enigma2 session.
I'm afraid that there is no real easy fix here. The only real solution is get you network running quicker... -or- wait a long time so enigma2 is started once there is a network and then add a big timeout here and leave the setting on 'auto'. So e.g. wait until the network is running with a timeout of e.g. 1 minute. Maybe arrange this in enigma2... when there is no network go to transponder time and then check with a timeout of 1 minute when the setting is ntp to check if the network is running then and then switch back to ntp.

WanWizard:
> I also do not understand why this is needed... config.misc.useTransponderTime.addNotifier(useTransponderTimeChanged, initial_call=False) As initial should be true here as it configs enigma2 in the proper way..... 

That is correct, the problem isn't the notifier, the problem is that `config.misc.useTransponderTime` is old code, this value doesn't exist anymore, after your change to implement NTP. You forgot to update this bit...

littlesat:
It is still used somehow for the rtc....

PR #3816, opened 2023-10-27
[StartEnigma] use correct value DVB time when start E2

Dima73:
-According to settings "Time synchronization method"


WanWizard:
What is the point of this as we have already established it is not needed as it is set a bit later when usageconfig is called?

Dima73:
> What is the point of this as we have already established it is not needed as it is set a bit later when usageconfig is called?

But this is correct.

WanWizard:
Besides the point. It is no longer needed, a few steps down in the code it is set again.

PR #3823, opened 2023-11-07
[VariableText] sanity check text 'str' type

Dima73:
File "/usr/lib/enigma2/python/Components/VariableText.py", line 12, in setText
TypeError: in method 'eLabel_setText', argument 2 of type 'std::string const &'

### DIFF
--- a/lib/python/Components/VariableText.py
+++ b/lib/python/Components/VariableText.py
@@ -7,9 +7,12 @@ class VariableText:
 		self.instance = None
 
 	def setText(self, text):
-		self.message = text
+		try:
+			self.message = str(text)
+		except:
+			self.message = ""
 		if self.instance:
-			self.instance.setText(self.message or "")
+			self.instance.setText(self.message)
 
 	def setMarkedPos(self, pos):
 		if self.instance:
@@ -21,4 +24,8 @@ class VariableText:
 	text = property(getText, setText)
 
 	def postWidgetCreate(self, instance):
-		instance.setText(str(self.message) or "")
+		try:
+			self.message = str(self.message)
+		except:
+			self.message = ""
+		instance.setText(self.message)


WanWizard:
Does this also need a bytes or bytesarray check?

as `str( b'abc' )` still returns bytes...

Dima73:
`Python 3.9.9 (main, Jan 26 2023, 17:53:23)
[GCC 10.3.0] on linux
>>> a = b'abc'
>>> str(a)
"b'abc'"
`

`>>> a = b"abc"
>>> str(a)
"b'abc'"`

`
>>> import os
>>> a = os.system
>>> str(a)
'<built-in function system>'
`

jbleyel:
I think that's not good because you hide the real error.

Dima73:
> I think that's not good because you hide the real error.

Yes, this is true, but what can you offer instead of an exception?
For example, if the text for some reason returned the int type.

I have been facing this error for many years in openPli.
Other images simply applied an exception.

WanWizard:
> > > > a = b'abc'
> > > > str(a)
> > > > "b'abc'"

My point, yes, as with Python 3, `"abc" != b'abc'`. 

Are both accepted by setText? 

And @jbleyel has a point, it hiddens problems and poor coding elsewhere. If an int is passed, ideally it should be converted/cast to string there where it is called.

jbleyel:
> And @jbleyel has a point, it hiddens problems and poor coding elsewhere. If an int is passed, ideally it should be converted/cast to string there where it is called.

Thank you !! 
That's exactly what I mean.
Fixing bugs from other sources by adding extra code and extra instructions is bad coding and less performant.

Dima73:
Saying one thing and doing another is simply double standards.
Example:
https://github.com/openatv/enigma2/blob/master/lib/python/Components/VariableText.py#L10
I can't change someone else's closed source plugin, but I can change the image.

Taapat:
If you introduce an exception, you must specify what type of exception you expect: https://www.flake8rules.com/rules/E722.html
In this case TypeError.

IanSav:
> Saying one thing and doing another is simply double standards. Example: https://github.com/openatv/enigma2/blob/master/lib/python/Components/VariableText.py#L10 I can't change someone else's closed source plugin, but I can change the image.

Before you quote and blame someone for coding code that conflicts with their comment you should first check who wrote the code and when.  The "try" code was added 11 years ago by "babsy98".  I believe that this predates @jbleyel's involvement with Enigma2.  This is also code which we have yet to refactor.  This code *will* be fixed in due course.

PR #3831, opened 2023-11-12
VideoHardware: Set resolution to 720p in language select screen

Hains:
Some TVs can't handle 1440x576i resolution which result in a black screen during language select screen after a fresh image install.
After language selected, the resolution jumps to 720p without any user interaction.

So use 720p from the start, which seems better than 576i.


Hains:
Side-effect.

littlesat:
Crazy work-a-round for I think an non issue.... with side effects as the default is changed. I have also no idea how we get to that obvious resolution while changing the language...

WanWizard:
Why is this merged if Hains closed it because of side effects?

littlesat:
I understood it was resolved a side effect as extra comment...

Taapat:
And why did you not pay attention to the syntax error shown by the workflow before merging?
https://github.com/OpenPLi/enigma2/actions/runs/6840472985/job/18599770324#step:7:3518

Taapat:
Since there was no response, I revert this commit because it causes a syntax error: https://github.com/OpenPLi/enigma2/commit/17cdccde9cb49f1a354852a19f0a8923418ed3e2

PR #3849, opened 2023-11-29
Stream Relay fixed to present information like normal DVB service

DimitarCC:
This makes all stream relay channels to report all the info like it was on the normal DVB channel.
So Provider, Transponder info, DVB namespace, SNR, AGC, BER is all there.
Also it fixes grayed out channels in Channel list.

This is test so keep in mind that have to be tested how it behaves on PLi more deeply.


littlesat:
Thanks!

PR #3853, opened 2023-12-01
[Added] Option to disable blinking entierly

DimitarCC:


WanWizard:
Wouldn't it be more logical to make the default False, so the default is the current behavior?

littlesat:
Years ago we had the discussion not to go for blinking in openPli. I also see a blinking clock when a recording is ongoing that currently is not in a setting, but the config is there

littlesat:
The name of the config is not correct... is says disable_blanking... while it actually does enable_blanking

DimitarCC:
Ahh yes.... OK i will change the name and will propose a PR

PR #3886, opened 2023-12-29
[FlashImage] increase search format checkModels()

Dima73:


littlesat:
Better resolve you homebuild sorry...

WanWizard:
> Better resolve you homebuild sorry...

How? To what? 

Changing the name of the created zip file requires changes to the BSP, not something we should ask from users.

ims21:
I am aggree with Harro. But It would probably be sufficient to use the variant for '-model-' and then add for generated by BSP: '-model_'

PR #3887, opened 2023-12-29
[UsageConfig] rename 'None' as 'No'

Dima73:
-remove twice entry config.usage.record_indicator_mode

### DIFF
--- a/lib/python/Components/UsageConfig.py
+++ b/lib/python/Components/UsageConfig.py
@@ -33,18 +33,17 @@ def InitUsageConfig():
 		refreshServiceList()
 	config.usage.alternative_number_mode.addNotifier(alternativeNumberModeChange)
 
-	config.usage.servicelist_twolines = ConfigSelection(default="0", choices=[("0", _("None")), ("1", _("Two lines")), ("2", _("Two lines and next event"))])
+	config.usage.servicelist_twolines = ConfigSelection(default="0", choices=[("0", _("No")), ("1", _("Two lines")), ("2", _("Two lines and next event"))])
 	config.usage.servicelist_twolines.addNotifier(redrawServiceList, initial_call=False)
 
 	config.usage.hide_number_markers = ConfigYesNo(default=True)
 	config.usage.hide_number_markers.addNotifier(redrawServiceList, initial_call=False)
 
-	config.usage.servicetype_icon_mode = ConfigSelection(default="0", choices=[("0", _("None")), ("1", _("Left from servicename")), ("2", _("Right from servicename"))])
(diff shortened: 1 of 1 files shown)


littlesat:
Why the replace None by No? And why default 3? For those who want tot get recordings highlighted can set it?

Dima73:
> Why the replace None by No? And why default 3? For those who want tot get recordings highlighted can set it?

What is None?
What is the meaning of programming language (Python, C+)?

default 3...
It was like this before, it's not my decision:
https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Components/UsageConfig.py#L47

littlesat:
I do not see any reason to change None to No here. None is in English also the same for show None... Nothing... Then No is also doubtful. Maybe better discuss this before on the forum then doing a direct merge request. Also for the default of the red letters for recordings in the channel selection screen. When was that changed to from '3' to '0'?

Ev0-BH:
> I do not see any reason to change None to No here. None is in English also the same for show None... Nothing... Then No is also doubtful. Maybe better discuss this before on the forum then doing a direct merge request. Also for the default of the red letters for recordings in the channel selection screen. When was that changed to from '3' to '0'?

Totally up to you want you want the options to say (it's your image).

But....

The question on the screen is asking "Do you want to show" ....

So the reply to that in correct English would be No not None

WanWizard:
Agreed.

Same with the others, they are all questions where "No" is the answer, not "None" (which would be correct for countable items, for 0 or the absence of a value).

WanWizard:
And while we're at it, can we also have a discussion about the use of ucfirst/title?

Nothing annoys me more than seeing "Yes" and "yes" or "No" and "no" in the same list. Can we standardize on either all UC first, or none of them, but not some?

WanWizard:
> When was that changed to from '3' to '0'?

@DimitarCC changed it, two weeks ago: https://github.com/OpenPLi/enigma2/commit/c9be99cf32ae9ad05bd6f09a73cccf6d47cb033d

littlesat:
I think that was by accident...
Should we accept this merge request for now?

DimitarCC:
This seems to be merging issue since that was duplicated. The default is different in oe-a for where the originally code was made for. But now i am confused which have to be the default value.

littlesat:
Sounds we had red letters 😉

WanWizard:
It most have changed multiple times, as it was 0 is OpenPLi 7.x and 8.x.. So that seems to be correct.

But the other part of this PR, the "None to No", must be merged, so that requires a new PR, as merging is all or nothing.

WanWizard:
The problem now is that the line is there twice: https://github.com/OpenPLi/enigma2/blob/develop/lib/python/Components/UsageConfig.py#L46-L47

So the one with default=3 just needs removing.

Dima73 (commented):
this not need

PR #3914, opened 2024-01-28
AVSwitch: allow importing iAVSwitch

athoik:
This commit allows to import name 'iAVSwitch' from 'Components.AVSwitch'

Plugin IPTVPlayer was failing with the following error: Plugin  Extensions/IPTVPlayer failed to load: cannot import name 'iAVSwitch' from 'Components.AVSwitch' (/usr/lib/enigma2/python/Components/AVSwitch.py) ImportError: cannot import name 'iAVSwitch' from 'Components.AVSwitch' (/usr/lib/enigma2/python/Components/AVSwitch.py)

Declaring the iAVSwitch prior calling InitAVSwitch fixes the error.

### DIFF
--- a/lib/python/Components/AVSwitch.py
+++ b/lib/python/Components/AVSwitch.py
@@ -3,6 +3,7 @@ from enigma import eAVSwitch, eDVBVolumecontrol, getDesktop
 from Components.SystemInfo import BoxInfo
 import os
 
+iAVSwitch = None # will be initialized later, allows to import name 'iAVSwitch' from 'Components.AVSwitch'
 
 class AVSwitch:
 	def setInput(self, input):


WanWizard:
Don't you love them globals... :(

athoik:
https://github.com/OpenPLi/enigma2/blob/0873597353c7d8a874e579d82f3a511268aac472/lib/python/Components/AVSwitch.py#L69

https://github.com/OpenPLi/enigma2/blob/0873597353c7d8a874e579d82f3a511268aac472/lib/python/Components/AVSwitch.py#L150 

It's initialized lated when called from `InitAVSwitch` but it's not possible to import `iAVSwitch`

WanWizard:
Because iAVSwitch wasn't defined as a global...

littlesat:
Why not resolve it in the plugin? Why does the plugin needs it to initialize here?
Personally I prefer we should not merge this. It can be easily resolved in the plugin.

WanWizard:
You can't resolve it in the plugin, problem is the instance is created in a method, so it is a local variable, which can't be accessed from elsewhere without an interface, be it a global or a "getInstance" type method.

littlesat:
Why does the plugin need it in the first place?Verstuurd vanaf mijn iPhoneOp 28 jan. 2024 om 19:55 heeft Harro Verton ***@***.***> het volgende geschreven:﻿
You can't resolve it in the plugin, problem is the instance is created in a method, so it is a local variable, which can't be accessed from elsewhere without an interface, be it a global or a "getInstance" type method.

littlesat:
Just relocate the import in the plugin when it exists should resolve it instead of import it on top of the plugin..

athoik:
How it should change?

https://github.com/oe-mirrors/e2iplayer/blob/e653b08e9b8cb880fd43a2f04717b18717165c35/IPTVPlayer/components/iptvplayer.py#L14

Plugin is doing an `from Components.AVSwitch import iAVSwitch` long after the init has run.

iAVSwitch is now a local variable.

littlesat:
By moving the import just before the line where it is needed instead on top of the pyVerstuurd vanaf mijn iPhoneOp 28 jan. 2024 om 22:59 heeft Athanasios Oikonomou ***@***.***> het volgende geschreven﻿
How it should change?
https://github.com/oe-mirrors/e2iplayer/blob/e653b08e9b8cb880fd43a2f04717b18717165c35/IPTVPlayer/components/iptvplayer.py#L14
Plugin is doing an from Components.AVSwitch import iAVSwitch long after the init has run.
iAVSwitch is now a local variable.

WanWizard:
afaik that still requires iAVSwitch to be global, you can't import a local variable?

PR #3916, opened 2024-02-01
[InfoBarStreamRelay] sanity check play service

Dima73:
File "/usr/lib/enigma2/python/Screens/InfoBarGenerics.py", line 182, in toggle
if nav.getCurrentlyPlayingServiceReference() == service: 
ValueError: invalid null reference in method 'eServiceReference___eq__', argument 2 of type 'eServiceReference const &'

### DIFF
--- a/lib/python/Screens/InfoBarGenerics.py
+++ b/lib/python/Screens/InfoBarGenerics.py
@@ -179,7 +179,7 @@ class InfoBarStreamRelay:
 				self.__srefs.remove(servicestring)
 			else:
 				self.__srefs.append(servicestring)
-			if nav.getCurrentlyPlayingServiceReference() == service:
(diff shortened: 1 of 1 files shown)


littlesat:
The crashlog was not related to this issue... Also the crash log tells the 'service' so the variable after the '=' has the issue.
I'm afraid further investigation is needed.
WanWizard who experience the issue was also not able to reproduce it.

Dima73:
You are wrong, I have fixed similar problems many times and this is due to getCurrentlyPlayingServiceReference() is None.

WanWizard:
If that is the case, isn't the question not "why"? Because I am absolutely sure a service was playing when I had the crash.

Dima73:
I'll try to check it again.
Stop play service and add/remove streamrelay sevice.

PR #3917, opened 2024-02-02
[Fixed] Exception when comparing eServiceReference with null

DimitarCC:


jbleyel:
How about the != operator?

DimitarCC:
I don't think there will be a problem since if c == null anyway the comparison will return false imho.

PR #3920, opened 2024-02-10
[MultibootSelection] move to index (preselect the current slot)

Dima73:


littlesat:
Why should this be an advantage?

Dima73:
Place the cursor on the current image.
Why not.

Dima73:
Place the cursor on the current image.
Why not.

littlesat:
Because I think there is an easier way to do it. And it is not completed as it does not cover the modes.

littlesat:
E.G. modes sorting it does not work in that.....

littlesat:
I will commit an alternative

PR #3936, opened 2024-03-31
[Converter.py] fix NameError in last commit

Huevos:


WanWizard:
What exactly is being fixed here? It looks more like conversion to a oneliner?

Huevos:
Previous:
```
	def appendToStringWithSeparator(self, str, part):
		if str == "":
			str = part
		else:
			text = f"{text}{self.separatorChar or ' '}{part}"
		return text
```

With this commit:
```
	def appendToStringWithSeparator(self, text, part):
		return f"{text}{self.separator or ' '}{part}" if text else part
```

There are 2 NameErrors.

str -> text
self.separatorChar -> self.separator

PR #3941, opened 2024-04-02
[PositionerSetup] update as last openViX code

Dima73:


ims21:
We have MENU and INFO in skins as buttons ?

PR #3947, opened 2024-04-10
[setup.xml/menu.xml] update "channelselection" entry

Dima73:
-setup.xml add more entry/remove old description
-menu.xml add "channelselection" entry


littlesat:
I do not think this is a good idea. Alternate numbering for example is something that does not only have effect on the channel list.

PR #3957, opened 2024-04-19
[Dish/PictureInPicture] support stream relay when using rotor

Dima73:


WanWizard:
Just wondering: what happens now if you request a stream from the box, or record a service that goes through streamrelay?

Dima73:
> Just wondering: what happens now if you request a stream from the box, or record a service that goes through streamrelay?

When is the motor running?

PR #3961, opened 2024-04-24
[Navigation] getClientsStreaming - filtering internal stream

Dima73:


littlesat:
It is still streamed somehow... so why mask it out of the streams?

Dima73:
> It is still streamed somehow... so why mask it out of the streams?

For example, the power off timer will not work or recording will not start if there is eStreamServer.getInstance().getConnectedClients().
This patch works in conjunction with this https://github.com/OpenPLi/enigma2/commit/72009475e66346bbc771b2b9280fef9dc53c9446

littlesat:
Sorry I still do not get it

Dima73:
Start recording and not free tuner
first step -->PiP enabled? if yes -->stop PiP
second step -->if eStreamServer.getInstance().getConnectedClients()? if yes -->stop all stream
again second step--> if eStreamServer.getInstance().getConnectedClients()? if yes -->stop all stream
but for 'stream relay' not need stop stream
simple need stop current service
So why do we need to add  'stream relay' to the getClientsStreaming list in this case?

Dima73:
> Not fully true. You can stop it... the overview of streams also indicate it is localhost. So somehow it is still up to the user and that is why I did decide not to stealth/hide this
Yes for overview of streams this indicate it is localhost.
But try to stop it, it will appear in the list of clients again in a few seconds.
That's why I only exclude it from getClientsStreaming because it causes an endless loop in start recording/power off timer.
Only getClientsStreaming for start recording/power off timer not eStreamServer.getInstance().getConnectedClients()!!

littlesat:
I think it is smarter to list it and when you stop it then when it is a stream for stream relay give a warning when you stop it or simply block it.... I suggest this is better than hiding it.

Hiding stuff always rise a flag to me. That triggers always comments and a second thought.

I hope you understand.

Dima73:
I don't need to stop the 'stream relay' and show a warning message in this case.
There is a special window for this StreamingClientsInfo.
But for these actions([ServiceStopScreen.py/PowerOffTimer.py/RecordTimer.py), which are performed automatically and use getClientsStreaming, you must ignore 'stream relay'.

https://github.com/search?q=repo%3AOpenPLi%2Fenigma2%20getClientsStreaming&type=code

littlesat:
Ok... now I get it... but do we have two options now to get the list of streams...?

And maybe next time try to give a better description about the problem you tried to resolve and why... 

Just filtering in the description hooked me to hide it anytime.

Then still in the streaming list it should still be blocked or warned.

Dima73:
`Then still in the streaming list it should still be blocked or warned.`
For  StreamingClientsInfo yes.

littlesat:
Indeed that is my suggestion.

PR #3963, opened 2024-04-26
[fastscan] skip transponders if satelite position is '0'

Dima73:
more info:

https://forums.openpli.org/topic/97184-unable-to-delete-the-00e-services-created-by-fastscan/

### DIFF
--- a/lib/dvb/fastscan.cpp
+++ b/lib/dvb/fastscan.cpp
@@ -496,8 +496,8 @@ void eFastScan::parseResult()
 			if (!westeastflag)
 				orbitalpos = 3600 - orbitalpos;
 
-			if (drop && transponderParameters.orbital_position != orbitalpos &&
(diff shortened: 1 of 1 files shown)


littlesat:
When does this happen? And why?

Dima73:
If the provider gives fictitious data, the orbital position is zero, a zero satellite appears that cannot be deleted.
This patch solves the problem.

littlesat:
What provider is doing this.... this should not be an openpli issue in general? Then the provider sends an incorrect channel list (and this means this is a work a round for someone else's failure) or there might be another reason behind it.

Dima73:
provider TéléSAT
read this https://forums.openpli.org/topic/97184-unable-to-delete-the-00e-services-created-by-fastscan/?view=findpost&p=1592824

littlesat:
Why didn't you add it to the description then I did not need to ask it a reference to the tread from months ago was also helpful

PR #3964, opened 2024-04-26
[PictureInPicture] stream relay  limit one service

Dima73:


littlesat:
Why is this needed? I can do multiple here and when there are limits to stream relay than much more is needed.
Streamrelay has no connection at all with transcode. So I do not see any need.

littlesat:
I understand you want to block PIP for stream relay when the box can only do one transcode and the box cannot do pip and transcode together. But a stream relay stream does not need transcode... so why are you trying to block it here?

littlesat:
Oh sorry I misread...

But still I can do multiple stream relays here and two (from the same channel) in PIP and main screen with stream relay with also a third stream on a client box + recording.... I do not see why this should be limited....
The merge request described what it does, but I miss a how/why and what you try to target.

Dima73:
OK.
I'll check the settings stream relay again on another receiver.

PR #3965, opened 2024-04-26
Skip 'stream relay' service if need stop stream

Dima73:


littlesat:
Sorry again I do not understand why this is needed.... please discuss this first on the forum before doing such merge requests here and/or better describe why it is needed, which issue it does resolve etc....

Dima73:
> Sorry again I do not understand why this is needed.... please discuss this first on the forum before doing such merge requests here and/or better describe why it is needed, which issue it does resolve etc....

[[Navigation] getClientsStreaming - filtering internal stream](https://github.com/OpenPLi/enigma2/pull/3961/commits/999cbb1fe0ab8644cbf386f21cf544be4a7209dd)

Dont stop stream if this stream relay(for stop stream relay need stop service)

littlesat:
On that other commit I had some questions... Why hide it as it is still somehow streamed....?
Where did you get this idea from?
It would be helpful when you add more comments so I understand the need.... I need to guess why you're suggesting this (with good attentions of course). What is the problem and how/why/what did you try to target.

I prefer to show the stream as it is a stream (unless it is streamed locally), but indeed when you try to stop it, you can block it at the screen with the streams with a popup that gives a warning that it is a local stream which may break a 'streamrelay'....

Dima73:
'stream relay' not really stream this local host.
For stop 'stream relay'  need only stop service.
He should not be counted as one of the eStreamServer.getInstance().getConnectedClients().
Same as [When box is shutting down do not show stream is busy for stream relay](https://github.com/OpenPLi/enigma2/commit/b38e77f4c48616d8593de36103fdb6e49b02a0e6)

littlesat:
Partly agree... but it is still a stream like any other...

Dima73:
This stream can only be disabled by disabling the service.
We shouldn't count him on the eStreamServer.getInstance().getConnectedClients() list at all.

littlesat:
Not fully true. You can stop it... the overview of streams also indicate it is localhost. So somehow it is still up to the user and that is why I did decide not to stealth/hide this

PR #3983, opened 2024-05-17
[CI] Possibility to disable operator profiles for CI+ module

Dima73:
thanks @DimitarCC


WanWizard:
What does this do?

DimitarCC:
Exactly what description says...For CI+ 1.3 modules disable operator profiles since implementation is incomplete (in enigma) and for some modules it give "error 10 " on every enigma restart.

WanWizard:
That would require me to know what an Operator profile is... :grin:

DimitarCC:
Its meant for TVs so module to have information about services and their operator's order and numbering. And it makes a channel list with that services in that order/numbering.
Completely  meaningless for enigma imho.

WanWizard:
Clear.

PR #3985, opened 2024-05-18
[Fastscan] inherit

Huevos:


littlesat:
Thanks but do I see a session = ession here?

littlesat:
Indeed a typo
ssession=ession,

Huevos:
LOL. Sorry. Can you copy and correct and test on your machine... Please?

littlesat:
I do not have ‘boxtime’ now…

Huevos:
Ok... Is your save/cancel commit tested?

littlesat:
Yes that is tested.
But I need a bit more investigation because eg in en av setup I needed the hasattr… and I want to avoid this. And in configuration it crashed on fan missing. That is why I added the try/except around the eval while adding it to the list. But this is finetuning. Now everything is canceled as it should be and saved as it should be. Another solution here is of course put them all in the list and do not show them (by flag)…. But I think that does not make the code better.

PR #3994, opened 2024-05-24
[FlashImage.py] /dev/sda and /dev/sdb

Huevos:
e.g. my vu its /dev/sdb


TwolDE2:
Yes, didn't notice Huevos had posted ......so close it ...logs however show reason

PR #4009, opened 2024-06-03
[SoftcamSetup] remove embedded skin

Huevos:
Skin will never be used. Automatically falls back to "Setup" skin.


littlesat:
Very nice.... and I guess there are more from these embedded skins....

PR #4010, opened 2024-06-08
[ScanSetup] always run cable cmd

Dima73:


littlesat:
What is this doing and why is this needed?

Dima73:
Currently, if the cable scan vendor binary does not exist, further execution stops.
But this is not true, you just need to warn the user about this via a message, but continue scanning according to the specified scanning parameters.

I'm correcting my mistake [[ScanSetup] show error message when not cable scan utility ](https://github.com/OpenPLi/enigma2/commit/7eea5e7892a23e6911368c18ab1d5e94b0363300)

PR #4024, opened 2024-06-16
SleepTimer - Blue for Start/Restart Sleeptimer only, Yellow for Stop

ims21:
- Yellow Stop Sleeptimer and it retains the selected option in the menu
- Blue is for Start or Restart Sleeptimer


littlesat:
Why should you add a yellow button while blue is sufficiant?

ims21:
No. See (my) comment on [https://forums.openpli.org/topic/99155-sleeptimer-broken/#entry1626923]

littlesat:
I understand that this gives a direct access… and you also demonstrated the idea of having yellow, blue (and menu) buttons added easily in setup screens.

PR #4027, opened 2024-06-21
[Harddisk] more readable device name

Dima73:


littlesat:
The idea is good, but the coding looks bad…. Could be better streamlined.

WanWizard:
That's @littlesat's idea of streamlining yes, but unreadable and difficult to debug. 
I hate complex one-liners that aren't needed... it has absolutely zero benefit, only downsides. :disappointed:

littlesat:
It is not the one liner idea that I do not like but more the try/except stuff when one of the files to parse do not exist. Maybe just check if it exists instead. And indeed do if both exists then give a combined value if one exists give back a string and if none exists then return unknown.

PR #4041, opened 2024-07-29
[ConfigList] sanity check use dummyConfigActions

Dima73:
File
"/usr/lib/enigma2/python/Plugins/SystemPlugins/SatipClient/plugin.py", line 387, in DiscoveryStart
  File
"/usr/lib/enigma2/python/Components/ConfigList.py", line 457, in dummyConfigActions
KeyError: 'menuConfigActions'

### DIFF
--- a/lib/python/Components/ConfigList.py
+++ b/lib/python/Components/ConfigList.py
@@ -452,11 +452,16 @@ class ConfigListScreen:
 		self.keySave()
 
 	def dummyConfigActions(self, value):  # Temporary support for legacy code and plugins that hasn't yet been updated.
-		self["configActions"].setEnabled(value)
-		self["navigationActions"].setEnabled(value)
-		self["menuConfigActions"].setEnabled(value)
-		self["charConfigActions"].setEnabled(value)
-		self["editConfigActions"].setEnabled(value)
+		if "configActions" in self:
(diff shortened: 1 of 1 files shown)


littlesat:
How do you trigger this?

littlesat:
OK… I’m afraid this could be better resolved in that plugin.

Dima73:
The problem is not in the plugin, but in the fact that self["menuConfigActions"] and etc is not always initialized at startup, but only under certain conditions. That's why I added an additional check.
`			if "key_menu" not in self and menu_button:
				self["key_menu"] = StaticText(menu_button.get('text', ''))
				self["menuConfigActions"] = HelpableActionMap(self, "ConfigListActions", {`
So,need fullUI=True and "key_menu" not in self and menu_button...

Dima73:
Same as self["charConfigActions"].
Where does this even get initialized?

littlesat:
Put it on the forum needs some investigation
It should be declared/initialized.
The check is as far I see not a solution.

Huevos:
https://forums.openpli.org/topic/98908-setup-code-proposal-and-more/page-14#entry1635297

The plugin cannot be calling `ConfigListScreen.__init__()`.

PR #4048, opened 2024-08-04
[VideoEnhancement] fix reset defaults settings

Dima73:


littlesat:
Also here a lot of lines can be removed regards to the new setup screen stuff.

PR #4050, opened 2024-08-09
Update unicable.xml Inverto Item 5458

cfob:
Adds matrix Inverto IDLU-UST110-CUO8O-32P (Item 5458)


littlesat:
Why item 5458 as comment? And not somehow at the model?

cfob:
Hi, other entries did not include Inverto item numbers. I followed the existing practice but felt it might be useful as a comment. 
Regards, Conor

⁣Get BlueMail for Android ​

littlesat:
It just triggered me that actually it is wrong

PR #4057, opened 2024-08-22
[Added] More options for load unlinked bouquets

DimitarCC:


littlesat:
For this one I have a better thing pending… so actually this is not needed

PR #4066, opened 2024-09-08
[RecordTimer] remove margin before/after min. if start lookup event

Dima73:


WanWizard:
You're not removing the margin, you're adding it? 

Or maybe better something like "[RecordTimer] take before/after margin into account when looking up an event" ?

Dima73:
I'm deleting because it was added earlier

`def parseEvent(ev, description=True):
	if description:
		name = ev.getEventName()
		description = ev.getShortDescription()
		if description == "":
			description = ev.getExtendedDescription()
	else:
		name = ""
		description = ""
	begin = ev.getBeginTime()
	end = begin + ev.getDuration()
	eit = ev.getEventId()
	begin -= config.recording.margin_before.value * 60
	end += config.recording.margin_after.value * 60
	return (begin, end, name, description, eit)

`

WanWizard:
What exactly is the reason for this change? 

I vaguely recall these margins were there for a reason, @littlesat ?

littlesat:
What are you trying to do here? Do you have an example why you think it is needed?

Dima73:
begin( before - 5 min) 12:00
end(after + 25 min) =12:20

queryTime = begin + (end - begin) // 5
we may get the wrong event

littlesat:
Sorry... I still do not understand why this is needed....

PR #4077, opened 2024-10-11
[TextBox] missing feature

Huevos:
Add possibility to specify widget name.
Feature is in other distros for a long time, e.g. OpenBlackHole, OpenViX. Prevent BSoD in PLi when using plugins that rely on this feature, e.g. https://github.com/DimitarCC/iptv-m3u-reader/blob/8560e189eb630bb39eecdb8a7620903d29740fb3/src/plugin.py#L1219

### DIFF
--- a/lib/python/Screens/TextBox.py
+++ b/lib/python/Screens/TextBox.py
@@ -4,19 +4,20 @@ from Components.ScrollLabel import ScrollLabel
 
 
 class TextBox(Screen):
-	def __init__(self, session, text="", title=None, pigless=False):
+	def __init__(self, session, text="", title=None, pigless=False, label=None):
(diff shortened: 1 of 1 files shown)


littlesat:
Actually I do not understand why you call this should be seen as a feature!!!!
As far I can see this is a work-a-round to get a wrong plugin running instead of adapting the plugin!!!!
Maybe I missed somehow to see this as feature.... 

But to avoid discussions I merged it....

But it would still be helpful why I should see this as feature?

Huevos:
Call it what you like (feature, workaround, hack) it is about code and skin reuse.

Without this you would have to create a whole new class and skin from scratch instead of just inheriting this class and a current skin.

PR #4088, opened 2024-10-26
[AutoDiseqc] code reorganization

Dima73:
-divide the satellites into several groups (west, east, central, circular)
-update data of many frequencies (old frequencies do not exist for a long time)
-add a 4 second pause at the last satellite found (before closing the window)
-minor code optimization (for red button/check status/circular lnb) 
-add two sat 4.8E/56E circular lnb
-[Satconfig] create a list(all,west, east, central, circular) to select the required satellites (yellow button)
green button (select all satellites - as before)

### DIFF
--- a/lib/python/Screens/AutoDiseqc.py
+++ b/lib/python/Screens/AutoDiseqc.py
@@ -10,7 +10,7 @@ from enigma import eDVBFrontendParametersSatellite, eDVBResourceManager, eTimer
 
 class AutoDiseqc(ConfigListScreen, Screen):
 	skin = """
-		<screen position="c-250,c-100" size="500,250" title=" ">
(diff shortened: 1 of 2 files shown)


littlesat:
Merged... but note it undermined the full auto idea, with keeping in mind I have some kind of hybrid idea.
There also more improvement suggestions in this merge commit that are separate features and maybe should preferably merged separately .

As far I understand when you start AutoDiseqc now you get an extra selection screen. I prefer to avoid that.... maybe add to nothing connected, auto all, auto central... with the sat positions... or even better 'fine tune' the order when you find the first connected lnb... 

Might be in (near future) I'll remove that screen...

So please expect if I have some time I'll modify it.... remove the extra screen and make it kind of automatically (switch order when first lnb pos is found)

PR #4090, opened 2024-10-30
[Satconfig] ignore "toneburst_a_b" mode for autodiseqc

Dima73:
-also re-add forgot DVBS value


littlesat:
Why? As far I known an auto scan is possible with this... (currently rarely used thing)
Dvbs check is not needed... you can only go to simple when it is dvbs....

Dima73:
> Dvbs check is not needed... you can only go to simple when it is dvbs....

Wnen use self.nimConfig.configModeDVBS.value  autodiseqc works

PR #4094, opened 2024-11-02
[ServiceInfo] fix empty audio name for iptv service

Dima73:
-also [UsageConfig] update audio codec


littlesat:
Why is that hacky style cnt stuff needed?
There is also stuff added that is not related to the merge request. It does also do something with guided audio. For not having description for iptv I think an easier solution could be planned.

PR #4103, opened 2024-11-23
[Picon] fix issues found by @Orlandox

Huevos:
Skip lookups when the channel name contains what is obviously junk.

Change priority to avoid selection of the incorrect picon.


WanWizard:
@Huevos Code crashes

See https://forums.openpli.org/topic/100639-crash-with-commit514071b11fa181687b12173dcd5b31e02ec24a62

PR #4178, opened 2025-01-26
Fixed movieSelected - part2

ims21:
- should be cherry pick to 9.1


littlesat:
But together with part 1!!!

littlesat:
But together with part 1!!!

PR #4180, opened 2025-01-30
Pull1

Dima73:


littlesat:
Good try and thanks.... but I'm afraid this is not the good way to resolve it... this is adding exceptions to undermine the boxinfo idea.

WanWizard:
Then why did you merge it in the first place?

I explained what the problem was, and I will do that again:
* you should use "model" in Enigma, not "machine"
* enigma.info should contain the correct model, any logic needed make sure of that should be in enigma-info.bb
  the code to do runtime detection is already there

PR #4184, opened 2025-02-04
Update sk.po

Pr0metheus2:
Slovak translation update


Pr0metheus2:
New update in progress...

PR #4199, opened 2025-03-11
[Navigation] more check set setWasInDeepStandby()

Dima73:
-attempt to fix a bug in duo4k, which sometimes turns off after finishing recording with the auto type.

### DIFF
--- a/lib/python/Navigation.py
+++ b/lib/python/Navigation.py
@@ -54,7 +54,7 @@ class Navigation:
 			ImportChannels()
 		if self.__wasTimerWakeup:
 			self.wakeup_timer_enabled = wakeup_time_type == 3 and self.__prevWakeupTime
-			if not self.wakeup_timer_enabled:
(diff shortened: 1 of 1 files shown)


littlesat:
Why is this needed. Is this only for the box you described or is this in general needed

WanWizard:
It is related to this discussion: https://forums.openpli.org/topic/101964-switching-the-box-to-deep-standby-after-timer-recording

littlesat:
So these additional checks were always missing. I asked because it is difficult to understand why/what/how these additional checks are needed and how they work. Therefore it would be helpful to have some additional comments. It is fine to just have them here

littlesat:
So these additional checks were always missing. I asked because it is difficult to understand why/what/how these additional checks are needed and how they work. Therefore it would be helpful to have some additional comments. It is fine to just have them here

littlesat:
So no additional changes in the coding itself needed.

WanWizard:
I can't judge whether or not these are the correct fixes, I haven't been able to test it.

PR #4216, opened 2025-05-01
[InfoBarNumberZap] don't  use  '0' key when service is seekable

Dima73:


littlesat:
We have discussions on the forum to do this completely different. Actually this is completely wrong.

PR #4219, opened 2025-05-03
SingleEPG filter - moved under standallone helpable actions

ims21:
- added helptexts
- simplified code, fixed code
- added saving changed values with key 0


littlesat:
Whey remove original time interval stuff?

ims21:
> Whey remove original time interval stuff?

No need more. Original time interval is restored always on exit of EPG. 
When user wants save changes, he can use "0" for it.

It is mentioned in the help texts in the settings for start/end time, as well as in the help section.

littlesat:
Ok

PR #4230, opened 2025-05-25
[Videomode] force set 10 bit for gbquad4kpro

Dima73:
-when use mode 2160p config.av.hdmicolordepth.value -->auto no image


littlesat:
Can be only accepted when you name it as a feature and put it in systeminfo.py...

No if box is a specific type then do something else outside systeminfo.py please.

littlesat:
Can we expect a proper merge request? I cannot easily test this due to I do not have that hardware

PR #4248, opened 2025-07-28
[WirelessLan] adapt for new pytyon3-wifi

Dima73:
-thanks oe-alliance team (from wifi.scan import Cell)

-old pytyon3-wifi(from pythonwifi.iwlibs import Wireless, getWNICnames) not work for gbquad4kpro

Note:
necessary new python3-wifi_0.8.0.bb python3-pbkdf2_1.3.bb

### DIFF
--- a/lib/python/Plugins/SystemPlugins/WirelessLan/Wlan.py
+++ b/lib/python/Plugins/SystemPlugins/WirelessLan/Wlan.py
@@ -1,65 +1,67 @@
+from re import compile
+from os.path import exists
+from wifi.scan import Cell
+
 from Components.config import config, ConfigYesNo, NoSave, ConfigSubsection, ConfigText, ConfigSelection, ConfigPassword
 from Components.Console import Console
 from Components.Network import iNetwork
 
-import enigma
+encryptList = ["WPA/WPA2", "WPA2", "WPA", "WEP", "Unencrypted"]
 
-import os
-from pythonwifi.iwlibs import Wireless, getWNICnames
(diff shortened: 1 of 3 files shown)


littlesat:
Is this really tested... 
actually we consider to go for a complete different way instead of hack around in existing code... instead of selecting interfaces give a list of ssids found nearby + add cabled option... and also when it was not cabled select 'known' wifi ssids etc...

WanWizard:
In itself it is not a bad idea to replace the existing 25 year old code, but I hope as part as a tested redesign and rework, not as a copy/paste from somewhere else...

littlesat:
Yep... but this merge request also looks like copy and paste... imports have different orders and ' replaced by " and visa versa... the real feature change is not really visible in this merge request.

WanWizard:
Which is why I haven't merged it yet. Haven't had the time to look at it in more detail, and test it on a box.

littlesat:
I’m afraid it needs to be cleaned.. see my previous comments.

WanWizard:
There is no need for all this, in this form.

There is a small bug in pythonwifi that is causing the problem, which is a repo we maintain ourselfs (@athoik), and the fix is very easy. But apperently the cause of the problem was never debugged, it only took me 5 minutes to find and fix the problem.

We do need to switch at some point, but I suggest to do that with a rewrite of the entire network code, which is long overdue, but not as a fix for a problem that is easily fixed. This is shooting with a canon at a mosquito...

PR #4256, opened 2025-09-16
[Picon.py] Fallback to service type 1 for radio services too.

AbuBaniaz:
Revert: https://github.com/OpenPLi/enigma2/commit/c0adcd06f2e66d6db78a52848d557a1a2c7788cf
Re-apply:  https://github.com/OpenPLi/enigma2/commit/feb612712c2c002c2dfa5065413fc929f45228d6

### DIFF
--- a/lib/python/Components/Renderer/Picon.py
+++ b/lib/python/Components/Renderer/Picon.py
@@ -89,8 +89,8 @@ def getPiconName(serviceRef):
 		#fallback to 1 for IPTV streams
 		fields[0] = '1'
 		pngname = findPicon('_'.join(fields))
-	if not pngname and fields[2] != '2':
-		#fallback to 1 for TV services with non-standard service types
+	if not pngname and fields[2] != '1':
+		#fallback to 1 for TV services with non-standard service types. And radio services.
 		fields[2] = '1'
 		pngname = findPicon('_'.join(fields))
 	if not pngname: # picon by channel name


littlesat:
I suggest not to do this... but checking for != 1 is still a good addition. Better to fix the picon set or at least as the picon set is common use always check on 1_0_1 first. But in general the picon set is wrong!

littlesat:
The check on != 1 is also not a good idea as this was 'checked' before

WanWizard:
Please look at the code again. The first one is for field 0, the second one for field 2.

AbuBaniaz:
I have tested this. I have just reverted what I was blamed for. Please make any changes you want.  

image

image

WanWizard:
Hey, no (intentional) blame was given here, just observed your commit changed it from "2" to "1". I didn't do a complete trace of all commit history :grin: . Sorry if you feld this was an attack on you.

For me, removing that single fallback step (of DVB type to 1) doesn't gain us a lot of processing or lookup time, especially if we move a specific 1_0_1 check further up the lookup chain, as that is the most likely hit. 

Compared to changing the entire generation process to use the full service reference for the symlink, and then still needing the fallback steps for those with channel lists using other service refs.

littlesat:
The != 1 is not needed.... as for 1 it should already be found... just remove that check.
And indeed when the piconset has the full service reference this step is indeed not needed.

AbuBaniaz:
@WanWizard 
So you want me to change the fallback filed checks in this order: 7, 3, 1 (python 6, 2, 0)?

WanWizard:
No, I think we need to review the entire picon fallback chain. Currently it checks
* exact service reference
* exact service reference with removed sub-namespace
* service reference with stream type changed to 1
* service reference with DVB type changed to 1, only for refs not type 2
* by normalized channel name

I don't think this is the optimal order, given the fact most people use SNP type picons, and for those using SRP picons, most use picons named 1_0_1_....

AbuBaniaz:
With regards to SNP, the only way to distinguish between two different channels with the same name, or those with different logos, is to use the service reference for some of them.

So please the last order as
Utf8  then 
SNP

littlesat:
Why not search on \ *_unique_part*...?

AbuBaniaz:
The whole reference is the only unique part. Tha is what is  searched first.

Anyway, changing order should be a different commit.

PR #4257, opened 2025-09-20
[frontend] VU+ Zero4K use more sat_max

Dima73:
-for more accurate values ​​at high signal levels


littlesat:
Please explain.... what makes this more accurate? It just tops the max nothing else....

Dima73:
The maximum signal level is clearly too high.
This change will slightly reduce the signal strength limit.

littlesat:
Please next time better describe what is changed and why... openATV is not good with this 😉

PR #4266, opened 2025-10-04
Parental rating

DimitarCC:


littlesat:
De ratings to epg import gives a red cross

DimitarCC:
> De ratings to epg import gives a red cross

yes but next commit fixes it.

PR #4267, opened 2025-10-05
[Picon.py] 1) make PiconLocator class ...

Huevos:
... This removes the need of global variables and is compatible with M3UIPTV plugin.

2) Update SRP algorithm to match https://github.com/picons/picons/blob/master/resources/tools/create-symlinks.sh#L74

3) Various changes to remove spaghetti code and stop unnecessary file system lookups.


littlesat (approved):
Thanks

PR #4282, opened 2026-02-04
Update de.po

harryhavermelk:
various fixes and improvements of the translations


WanWizard:
Sorry, can't accept this, we only accept PR's against the default branch, which currently is "scarthgap".

harryhavermelk:
@WanWizard Fine. But then this article should be updated with the correct link accordingly: https://wiki.openpli.org/Information_for_translators

WanWizard:
Thanks, I wasn't aware a specific branch was named there. I'll have that updated.

PR #4285, opened 2026-02-06
Add Softcsa support 

AbuBaniaz:


AbuBaniaz:
May be odd niggles that will need ironing out. So please beware before merging.

technl:
I tested this on my local build environment and didn't encounter any issues...

PR #4296, opened 2026-02-10
Reword whitelist file option

AbuBaniaz:


littlesat:
Please note the file itself is still there... you can still remove/add service from them. Should I also 'hide' these options in the ChannelSelection screen?

PR #4298, opened 2026-02-12
Update sk.po

Pr0metheus2:


Hains:
Please use scarthgap branch, not develop.

PR #4301, opened 2026-02-12
Fix audio track display after selection when SoftDecoder is active

AbuBaniaz:
When SoftDecoder handles audio track selection, m_current_audio_pid was not being updated, causing getCurrentTrack() to return the wrong index. The audio selection menu would show the old track as active instead of the newly selected one.

Fix by updating m_current_audio_pid before delegating to SoftDecoder.


AbuBaniaz:
To resolve this issue: https://forums.openpli.org/topic/104432-softcsa/page-4#entry1732800

PR #4304, opened 2026-02-14
[InfoBarSubserviceSelection] hide "Quick zap" if not current service …

Dima73:
…in subservices list

- Example: Sky Sport 1 19.2E --> event name "Sendepause"
File "/usr/lib/enigma2/python/Components/ActionMap.py", line 76, in action
File "/usr/lib/enigma2/python/Components/ActionMap.py", line 56, in action
File "/usr/lib/enigma2/python/Screens/SubservicesQuickzap.py", line 61, in nextSubservice
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'


littlesat:
I do not think this is the solution... the quickzap services should still be shown... this looks like a work-a-round in solution and not as a real solution.

PR #4321, opened 2026-03-12
[SoftCSA] Free resources after ECM analysis, fix ePMTClient leak

WXbet:
- ECM Monitor: Always call stopECMMonitor() after CSA-ALT analysis completes, freeing Demux FD, Section Reader and ~40 callbacks/s. Previously the monitor kept running for all CSA-ALT channels (Live-TV, Recording, PiP, FCC) because stopECMMonitor() was only reachable in the !m_active branch.

- ECM Monitor cache path: Return early from startECMMonitor() when the cache already provides the CSA-ALT result, avoiding creation of an unnecessary Section Reader.

- ePMTClient: Add destructor to free receivedData buffer on mid-packet disconnect (softcam crash/restart).

- eDVBCSASession: Remove unused m_cw_connection member. CW signal uses sigc::trackable for automatic disconnect.

- Improve logging:
  - Rename caller log "Starting ECM monitor" to "Requesting ECM monitor" to clarify that startECMMonitor() may return early (cache hit or invalid PID) without actually starting a reader.
  - ECM Info: Replace sparse eDebug + eTrace with single eDebug containing all parsed fields grouped by pipe separators (crypto | timing | routing). Fix format string where cardsystem was mislabeled as reader.

### DIFF
--- a/lib/dvb/cahandler.cpp
+++ b/lib/dvb/cahandler.cpp
@@ -329,14 +329,9 @@ bool ePMTClient::processEcmInfoPacket()
 					memcpy(&ecmTime, receivedData + 11, sizeof(uint32_t));
 					ecmTime = ntohl(ecmTime);
 
-					eServiceReferenceDVB service;
-					if (parent->getServiceReference(service, serviceId) == 0)
-					{
-						eDebug("[ePMTClient] ECM Info %s", service.toString().c_str());
-					}
 					// Store CAID for this service (will be sent with next CW)
 					parent->m_service_caid[serviceId] = caid;
-					eTrace("[ePMTClient] ECM Info serviceId %u, caid %04X, ecmTime %u, cardsystem %s, reader %s, from %s, protocol %s, hops %d", serviceId, caid, ecmTime, cardsystem, reader, from, protocol, hops);
+					eDebug("[ePMTClient] ECM Info msgid %u | caid %04X pid %04X prov %06X | %ums | %s via %s from %s hops %d", serviceId, caid, pid, providerId, ecmTime, reader, protocol, from, hops);
 
 					return true;
 				}
--- a/lib/dvb/cahandler.h
+++ b/lib/dvb/cahandler.h
@@ -104,6 +104,7 @@ protected:
 	bool processEcmInfoPacket();
 public:
 	ePMTClient(eDVBCAHandler *handler, int socket);
+	~ePMTClient() { delete[] receivedData; }
 	void sendClientInfo();
 	int writeCAPMTObject(const char* capmt, int len);
 	bool isProtocol3() const { return m_serverInfoReceived; }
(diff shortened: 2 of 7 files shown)


littlesat:
I suggest when cashed still check and when it is not for softcsa then remove it from the cache.... I hope the cache is stored within lamedb.

WXbet:
The `s_csa_cache` is a plain `static std::map` in memory, internal to `csasession.cpp`. It has no relation to lamedb and is not persisted to disk. It starts empty on every enigma2 restart.

Its purpose is to avoid repeating the ECM analysis when switching back to a channel that was already analyzed during the current session. Here is how it works:

**First tune to a channel:**
1. `startECMMonitor()` is called — cache is empty for this service
2. A Section Reader is opened, an ECM filter is started on the ECM PID
3. The first ECM arrives, `ecmDataReceived()` analyzes it and determines whether the channel uses CSA-ALT
4. The result (`is_csa_alt`, `ecm_mode`) is stored in the cache, keyed by SID/TSID/ONID
5. The Section Reader is closed (this is what the PR fixes — previously it kept running)

**Subsequent tune to the same channel:**
1. `startECMMonitor()` is called — cache hit
2. The cached result is used directly (`is_csa_alt`, `ecm_mode`)
3. If CSA-ALT: activate immediately, return without opening a Section Reader
4. If not CSA-ALT: `ecm_mode` is loaded from cache, no further analysis needed — no Section Reader opened either

There is no need to "re-check and remove from cache" because CSA-ALT is a property of the encryption scheme used by the provider for that channel. It does not change spontaneously during a session. And since the cache is purely in-memory, it is automatically cleared on every enigma2 restart — so any theoretical change would be picked up after the next restart at the latest.

However, if you have a specific scenario in mind where the current behavior is causing a problem, I'd be happy to look into it. Do you have a concrete suggestion for what should be reviewed or changed?

littlesat:
Maybe it is just an idea to store it with an addition in lamedb.. so it 'stays' cached after an enigma2 restart. With this keeping in mind clear the cash might be needed.
Thanks also as now I understand why I could not find a file where the cash could be stored...
But if storing it does not have a real added value then we can consider to forget my comments

WXbet:
Thanks for the suggestion. Persisting the cache in lamedb would indeed eliminate the ECM analysis on the first tune to each channel after a restart. However, in practice the added value is very small:

The ECM analysis completes on the very first ECM packet per channel, which typically arrives within 100ms after tuning. So the "cold start" cost is barely noticeable. Once a channel has been analyzed, the in-memory cache handles all subsequent switches back to that channel instantly for the rest of the session.

Storing it in lamedb would also introduce some complexity: lamedb is the service database, and mixing in runtime detection data means we would need invalidation logic (what if a provider changes their encryption scheme). With a memory-only cache that starts fresh on each restart, this is simply not a concern.

So I think your conclusion is right — it does not have enough added value to justify the extra complexity. The current approach keeps it simple and reliable.

PR #4332, opened 2026-04-12
[ServiceName] add sanity check 'sref'

Dima73:


littlesat:
This merge request contains a sanity check and a few lines above another change... that does not do the same as it did before. And then likely the sanity check is not needed.

Huevos:
This commit deletes `sref = sref or ref`

PR #4337, opened 2026-05-15
SingleEPG - added 'Rating sort' under instant rec button with filteri…

ims21:
…ng and existing sort preservation


ims21:
will be rewrited

PR #4338, opened 2026-05-18
[Picon.py] Cosmetic readability changes

AbuBaniaz:
Better shows the order of selection

### DIFF
--- a/lib/python/Components/Renderer/Picon.py
+++ b/lib/python/Components/Renderer/Picon.py
@@ -84,14 +84,33 @@ class PiconLocator:
 		fields = GetWithAlternative(serviceRef).split(":", 10)[:10]
 		if not fields or len(fields) < 10:
 			return ""
-		basenames = ["_".join(fields), (p := "1_0_1_%s_0_0_0") % (x := ("_".join(fields[3:7]))), p % (x[:-4] + "0000")]
-		for basename in dict.fromkeys(basenames).keys():  # skip duplicates, maintain order
+		full_ref = "_".join(fields)  # 1. Full service reference
+		partial_ref = "1_0_1_%s_0_0_0" % "_".join(fields[3:7])  # 2. SID, TSID, ONID and NAMESPACE with padding
+		partial_ignore_subnet = "1_0_1_%s_0_0_0" % ("_".join(fields[3:7])[:-4] + "0000")  # 3. SID, TSID, ONID and NAMESPACE with padding, ignoring namespace subnet differences
+		basenames = [full_ref, partial_ref, partial_ignore_subnet]
+
+		pngname = ""
+		for basename in dict.fromkeys(basenames):  # skip duplicates, maintain order
 			if pngname := self.findPicon(basename):
 				break
+
 		if not pngname:  # picon by channel name
-			if (sname := eServiceReference(serviceRef).getServiceName()) and "SID 0x" not in sname and (utf8_name := sanitizeFilename(sname).lower()) and utf8_name != "__":  # avoid lookups on zero length service names
-				legacy_name = sub("[^a-z0-9]", "", utf8_name.replace("&", "and").replace("+", "plus").replace("*", "star"))  # legacy ascii service name picons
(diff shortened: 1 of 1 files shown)


Huevos:
More AI code.

WanWizard:
May be so, but I don't think improving readability is AI slop.

Or is there more in this PR than just improving readability?

Huevos:
Does it improve readability.

As I see it it adds a lot of bulk for no good reason.

This is the exact opposite type of code that Littlesat likes.

The current code uses simple oneliners where this adds intermediate variables to the namespace.

To me that makes this commit do the opposite of what the title suggests. It actually makes it harder to read. Adds 20 lines of unnecessary code. Increase the file length by 20%.

But you do what you think best, but certainly will not be added to our repo.

Huevos:
And by the way it changes the logic, so is going to break stuff.

WanWizard:
It was a question, not a statement, as you know I'm not very proficient in Python.

but if it changes the logic, than that's a big NO in my book.

Huevos:
It removes this commit: https://github.com/OpenViX/enigma2/commit/ebe467d8c93a107aa73e6bea0d0be52296d63ef7

WanWizard:
@littlesat close?

Huevos:
The commit above was important. @DimitarCC found several cases where the original algorithm failed without it.

AbuBaniaz:
The order of the search logic is still:
```
Full service reference
SID, TSID, ONID and NAMESPACE with padding
SID, TSID, ONID and NAMESPACE with padding, ignoring namespace subnet differences
UTF8SNP name
UTF8SNP name with resolution suffix stripped
SNP name
SNP name with resolution suffix stripped
```

The change was submitted to make it more readable as to the order in which the picons are selected.

Somoene has consistently criticised Littlesat's one liners that remove readability. It is a bit rich to start using Littlesat's approach as a basis to obstruct changes.

AbuBaniaz:
The history of the changes prior to this commit that allegedly broken is here: 
https://github.com/OpenViX/enigma2/commits/ebe467d8c93a107aa73e6bea0d0be52296d63ef7/lib/python/Components/Renderer/Picon.py

If someone wants to make the changes in a better way, please do. The file currently is not human readable.

Huevos:
> The order of the search logic is still:
> 
> ```
> Full service reference
> SID, TSID, ONID and NAMESPACE with padding
> SID, TSID, ONID and NAMESPACE with padding, ignoring namespace subnet differences
> UTF8SNP name
> UTF8SNP name with resolution suffix stripped
> SNP name
> SNP name with resolution suffix stripped
> ```
> 
> The change was submitted to make it more readable as to the order in which the picons are selected.
> 
> Somoene has consistently criticised Littlesat's one liners that remove readability. It is a bit rich to start using Littlesat's approach as a basis to obstruct changes.

And that order changes the logic of the original. Which is what happens when AI plays with code.

AbuBaniaz:
I deliberately had the utf8snp names to be together in the file and snp names to be together. So there is a slight change in order. They are now in the the order in which I understood them to be. This minor order issue highlights the fact that the search order was never readable nor was it documented correctly.

From:
```
utf8_name
legacy_name
utf8_name with suffix stripped
legacy_name with suffix stripped
```


To:
```
utf8_name
utf8_stripped
legacy_name
legacy_stripped
```

Huevos:
> I deliberately had the utf8snp names to be together in the file and snp names to be together. So there is a slight change in order. They are now in the the order in which I understood them to be. This minor order issue highlights the fact that the search order was never readable nor was it documented correctly.
> 
> From:
> 
> ```
> utf8_name
> legacy_name
> utf8_name with suffix stripped
> legacy_name with suffix stripped
> ```
> 
> To:
> 
> ```
> utf8_name
> utf8_stripped
> legacy_name
> legacy_stripped
> ```

And that order is wrong which @DimitarCC discovered 2 years ago during extensive testing.

AbuBaniaz:
That order is correct to me. If I am using utf8 names, why would I want it to be interjected by SNP.

I have closed the pull request, you can stop making things up.

littlesat:
Doing the picons by service name was always a bad idea. Therefor I also refuse it for a longer time. The only method that is correct and unique for now is by reference. To do it really right we need a good re-design or go for reference only.

PR #4344, opened 2026-06-11
[EPGSelection] add support catchupPlayerFunc

Dima73:
-this need for plugins when using archive for IPTV service

### DIFF
--- a/data/keymap.xml
+++ b/data/keymap.xml
@@ -663,6 +663,11 @@
 		<key id="KEY_6" mapto="endDown" flags="m" />
 	</map>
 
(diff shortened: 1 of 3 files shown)


Huevos:
How does this add the icon to the list?

Dima73:
This will be the text.

Huevos:
It is supposed to hace an icon in the epg, similar to record icon.

Dima73:
See.
self.catchupicon = LoadPixmap(cached=True, path=resolveFilename(SCOPE_GUISKIN, 'icons/catchup.png'))

Huevos:
ok

PR #4345, opened 2026-06-24
[EPGSelection] single EPG - add support catchupPlayerFunc

Dima73:
-this need for plugins when using archive for IPTV service

### DIFF
--- a/data/keymap.xml
+++ b/data/keymap.xml
@@ -663,6 +663,11 @@
 		<key id="KEY_6" mapto="endDown" flags="m" />
 	</map>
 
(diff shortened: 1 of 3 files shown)


ims21:
If this is intended for specific plugins, wouldn't it be better for the plugin to handle it using "hijacked" functions, similar to how the PartnerBox plugin does it?

Dima73:
In the future, this can be used universally not only for one plugin, but for several that are designed for IPTV.

littlesat:
I hate hijacked functions...

PR #4346, opened 2026-06-27
UsageConfig - apply saved NTP/DVB time sync setting on startup

ims21:
timesync notifier was registered with initial_call=False, so saved configuration was not applied after Enigma2 startup until the user changed the setting manually.


littlesat:
I suggest this should be reverted as this is not a solution. See forum

When it is set to 'NTP' and there is no 'NTP' time known (yet) when enigma2 starts-up the time will be in 1970... so recordings/timers do not start. You need to 'fallback' to DVB time in these cases. When enigma2 does start-up and the 'system' already as an appropriate time I'm afraid adding the non initial check was a good idea.

PR #4348, opened 2026-07-01
Clock source: reduce timer jitter by synchronizing to second boundaries

ims21:
- this not only avoids starting the timer off the exact second boundary, but also compensates for timer delays caused by box load.
- the trade-off is that seconds are updated 20 ms later, which is completely imperceptible.

### DIFF
--- a/lib/python/Components/Sources/Clock.py
+++ b/lib/python/Components/Sources/Clock.py
@@ -10,7 +10,7 @@ class Clock(Source):
 		Source.__init__(self)
 		self.clock_timer = eTimer()
 		self.clock_timer.callback.append(self.poll)
-		self.clock_timer.start(1000)
+		self.startTimer()
 
 	@cached
 	def getClock(self):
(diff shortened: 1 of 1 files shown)


littlesat:
I still think this is not a good idea... it may improve it 'cosmetically' but not always. This is what happens as enigma is busy... then timers will jitter.
When done then better do it eg in cpp where the repeated timer is arranged. Then also all other timers are compensated.

ims21:
The point is that after the system load drops, the second hand becomes synchronized again. Nothing more, nothing less. And it's reliable. I doubt anyone will ever rework the C++ timer implementation.

Moreover, even if the timers were fixed, by design a timer will never fire exactly on a whole-second boundary.

littlesat:
I do not think this is a real issue. For your our previous standard skin had a second counter on the infobar without any issue as far I remember. I saw your logs on the forum... big question what holds the ui interface at your box?

ims21:
I'm therefore very glad that you'll rework the timer so that it fires exactly on the whole second.

WanWizard:
There is plenty that interrupts. I remember from "the old days" that it didn't matter what the box was doing, audio and video was never impacted. Now I see freezes and hickups quite regularly, when the box is doing something CPU intensive.

PR #4354, opened 2026-07-11
Font - use \c to restore the original color

ims21:
- ff \c is not followed by 8 valid hexadecimal digits, restore the original foreground color.
- variant without change signature in appendGlyph()

### DIFF
--- a/lib/gdi/font.cpp
+++ b/lib/gdi/font.cpp
@@ -769,6 +769,7 @@ int eTextPara::renderString(const char *string, int rflags, int border)
 
 	unsigned long newcolor = 0;
 	bool activate_newcolor = false;
+	bool activate_colorreset = false;
 	int nextflags = 0;
 
 	for (std::vector<unsigned long>::const_iterator i(uc_visual.begin());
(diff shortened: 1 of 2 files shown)


littlesat:
How can you use /caarrggbb for set a color and /c for clear a color? I recommend at least to use something different.... (maybe /o for orriginal color?)

I know it is rare when the text after /c for clear might be isdigit, but in 'theory' this could happen.... That was why the idea for /C was initially a good one....

WanWizard:
Why? What is wrong with using /c without a code?

jbleyel:
> Why? What is wrong with using /c without a code?

\c20260101

littlesat:
It sounds maybe strange... but what if the text after the /c is a hexidecimal number? Rare but it could in theory happen. And therefor a red flag for me to do it like this. So please consider a different 'syntax'.
The date example will eg also change a color as it is hexidecimal...

ims21:
And heaven forbid someone writes \new or \the :-)

littlesat:
I hope you now understand the why....
Not \n and \t is also a standard. \c is as far I know enigma2 fantasy... initialy intended for subtitles.

jbleyel:
I will go with `\c` as start and `\C` as the reset.

ims21:
OK, guys, I'll rework it for you to \C ... but this is the last time :-)

ims21:
But I feel a bit like I'm just shunting things around :-)

ims21:
Guys, how do you want this to behave?

If text contains \C but no color was previously activated in string using \caarrggbb, should \C be displayed as part of the text, or should it simply be treated as an escape sequence and disappear?

Should it always disappear?

I have both variants ready.

jbleyel:
\C is a escape sequence and should be not displayed

ims21:
but \c without parameters is displayed too (ok, it is false using => should be displayed whole)...  For \C ... always drop it ? better, imho

ims21:
I will close it and add PR for \C as reset. (And \C ... always drop. )

PR #4358, opened 2026-07-16
servicedvbrecord - retry EIT file creation after initial failure

ims21:
- prepare() is called before the receiver tunes to the channel being recorded, so event may not yet exist in the EPG cache. 
- if the initial .eit creation fails for this reason, store the required parameters and retry once after 30 seconds.

### DIFF
--- a/lib/service/servicedvbrecord.cpp
+++ b/lib/service/servicedvbrecord.cpp
@@ -17,8 +17,10 @@ DEFINE_REF(eDVBServiceRecord);
 
 eDVBServiceRecord::eDVBServiceRecord(const eServiceReferenceDVB &ref, bool isstreamclient): m_ref(ref)
 {
+	m_eitRetryTimer = eTimer::create(eApp);
 	CONNECT(m_service_handler.serviceEvent, eDVBServiceRecord::serviceEvent);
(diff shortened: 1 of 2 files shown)


jbleyel:
Can you please explain how to reproduce the issue?
I have never seen an empty or missing eit file.

ims21:
> Can you please explain how to reproduce the issue?
> I have never seen an empty or missing eit file.

Just set a timer well in advance on a channel you rarely tune to, then make sure the EPG cache gets cleared.

Since prepare() (which looks up the event in the EPG cache and creates the .eit file) is executed about 20 seconds before the recording actually starts, the event is no longer present in the EPG cache at that moment, so the .eit file is not created.

A common way to reproduce this is to enable clear EPG before scan in EPGRefresh and use a short dwell time per transponder.

ims21:
Example , when I had there debug: 

saveEventToFile failed
eitret=-1
eit_event_id=20684
begTime=1783287300
endTime=1783293900
ref=1:0:1:3736:C8E:3:EB0000:0:0:0:
now=1783287280
queryTime=1783290600
lookupEventId=-1
lookupEventTime(begin=1783287300)=-1
lookupEventTime(middle=1783290600)=-1
lookupEventTime(end-1=1783293899)=-1

PR #4380, opened 2026-09-11
Add 'After any restart' option to 'Startup to Standby'

ims21:


WanWizard:
Would it be useful to change the description of "yes" to "Only after power-on" or something like that?

WanWizard:
And I think it should be "After every restart" instead of "any". Or, if we rename "yes", this could be simply "Yes", or "Always" ?

ims21:
What about:

No
After box startup only 
No, except Wakeup timer
Always

WanWizard:
I'm struggling a bit with the word "Startup", but I understand  that some people might see "power on by physically switching on the power" and "power on by pressing the On button" as different.

I also think, the double negative "no, except" isn't very clear.

What about
````
No, never
Yes, only wake-up timers
Yes, only after startup
Yes, always
````
to make it logical, and in a logical order?

ims21:
and change helptext to: "Configure when the set top box should enter standby after startup or restart." or "Configure when the receiver should enter standby after startup or restart."

PR #4383, opened 2026-09-13
ePositionGauge - add hidePointerOnZeroLength skin parameter

ims21:
- add hidePointerOnZeroLength skin parameter to hide pointer when playback length is zero.
- default behavior remains unchanged for backward compatibility.


ims21:
A completely optional parameter, which only has an effect if it is actually used, is not acceptable?

WanWizard:
That is a comment for?

ims21:
It was just a comment on the fact that this still hasn't been added.

WanWizard:
Ok. 

Haven't done it because I have no clue what this is about, and I think @littlesat commented on it, so I leave it to him to decide.

PR #4385, opened 2026-09-17
SoftCSA Updates

kueken:
Not finished yet, but I can't get any further, sorry

I would wait to accept the PR


WanWizard:
What is not finished yet? What is missing?

kueken:
This is certainly not an easy question to answer.
But in relation to the commit itself, the revert must be considered.
https://github.com/OpenPLi/enigma2/pull/4385/changes/1031665d4e9f4efbd2033ce5e477a7ef7e50da87
The actual commit seems to me to be important, but it cannot be adopted so easily.

WanWizard:
I've looked at 1031665d4e9f4efbd2033ce5e477a7ef7e50da87 but that is way over my head, so I can't judge what it does and what the impact is of not implementing it.

PR #4386, opened 2026-09-17
HDMI-CEC - removed unnecessary UTF-8 conversion.

ims21:
- it did not cause problems in practice because CEC physical addresses are normally low, but it was still incorrect: some 2-byte values can form a valid UTF-8 sequence and be collapsed into a single character.


WanWizard:
So CEC data is always ISO 8859-01, never UTF-8?

ims21:
CEC data itself is neither ISO-8859-1 nor UTF-8; it is raw byte data. ISO-8859-1 is only used here as a 1:1 bytes-to-str mapping for the Python binding.

WanWizard:
Thought so. But wouldn't it be more logical to use `decode('ascii')`, given that it doesn't have an encoding ?

ims21:
decode('ascii') = only 128 values, may fail
decode('utf-8') = may combine multiple bytes into a single character
decode('latin-1') = every byte 00–FF is converted to exactly one character

It probably wouldn't happen here that something like b'\xC3\xA3' would be received, but UTF-8 was not correct here.

WanWizard:
Ah, ok, I thought ascii was 8-bit, not 7-bit.

littlesat:
'ascii' was always 7 bits ;)

WanWizard:
Dumb language, Python, I'll stick to PHP... :rofl:

